PR #13, opened 2016-07-18
Add config commands for managing SDK config files

### LINKED ISSUE #12: Add `config` command for setting up and modifying the `~/.globus.cfg`
This was a feature request I got from Steve -- he seemed to feel pretty strongly about it being a necessary intro point.

My instinct is that semantics should be somewhere between what `knife configure` and `git config` offer.

So, `globus config init` (knife-like) should be prompt-oriented and allow users to enter info line by line.
For example,
```shell
$ globus config init
Please enter your auth token: XXXXX
Please enter your transfer token: XXXXX
...
$ globus config init --auth-token XXXXX
Please enter your transfer token: XXXXX
```

However, it might also be useful to have git-like setter/getter functionality. Maybe something more like this:
```shell
$ globus config show general.auth_token # echoes the value
...
$ globus config add general.auth_token XXXXX
```
of course, anything without a dot should be treated as part of the general config section, so these are equivalent to
```shell
$ globus config show auth_token
...
$ globus config add auth_token XXXXX
```
and of course we'd want `globus config remove`

Questions I still have:

1. Should we support a `--system` flag, like `git-config`, to operate on `/etc/globus.cfg` ?
2. Are the verbs "get", "set", and "unset" (like git) or "show", "add", and "remove" (like the other parts of the CLI for endpoints &c.)?
3. Support for `globus config edit` which does `$EDITOR ~/.globus.cfg`?
4. Should the `config` command know that certain values (like tokens) are secret, and show redacted versions of them?
5. What does show/get do when the key is absent? Empty string seems ambiguous -- it's a valid value.
6. What does add/set do when the key is already present? Should there be a `--overwrite` flag?

@bd4 I'm concerned that this needs direct access to the SDK's `GlobusConfigParser`. It's a very unusual internal use-case, and I don't think we want to support it for external groups.
I'm happy to `from globus_sdk.config import GlobusConfigParser`, but I'll need a new function attached to flush its state to a file (the internal `SafeConfigParser` should support this for us), and I'll need to update the `GlobusConfigParser` to support reading only one file. Otherwise, it would load system and default config, and flush all of that back out when it's done, which would be bad.
There are some alternatives:
- Should we do a `0.3.0` of the SDK in which the config parser insists that there is a `[general]` section heading, like a typical `ConfigParser`? Then this would not need the specialized `GlobusConfigParser` at all. That breaks compatibility with existing config, which I don't like. We'd need clear messaging for anyone who has started using it.
- I could make a CLI config parser with implements its own specialized [_read()](https://github.com/globus/globus-sdk-python/blob/master/globus_sdk/config.py#L48)

sirosen:
This resolves #12 

We add the following commands:

  - `globus config edit`
  - `globus config init`
  - `globus config show`
  - `globus config set`
  - `globus config remove`

`show`, `set`, and `remove` all use the same syntax for specifying option names.
Options are always given to them as the first positional argument, and may contain both the section as well as the specific option name.

An option without a section is just a string, like `output_format`.
To include a section, a dot is used to delimit the section from the option name, as in `cli.output_format`.
When no section is specified, the section defaults to `general`.

The following, therefore, are all valid commands, with intuitive semantics:

```shell
# set the auth token to "abc"
$ globus config set general.auth_token abc
# set the auth token (same as above) to "xyz"
$ globus config set auth_token xyz
# unset the CLI output format
$ globus config remove cli.output_format
# show the current output format
$ globus config show cli.output_format
```

The `init` and `edit` commands are a bit fancier.

`globus config edit` will open the `$EDITOR` or `nano` on the `~/.globus.cfg`

`globus config init` is designed to be an interactive "guide" through setting the various config values that you probably want/need.
It also accepts optional arguments to pre-fill any of the values which, left to its own devices, it will prompt about.

Many of these commands could be expanded in the future to take a `--system` flag to operate only on the system-wide config in `/etc/`.
However, this functionality is somewhat complex, and it's not clear that it's necessary yet.


globus_cli/config_command/edit.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+@click.command('edit', help='Edit your Globus Config file')
+@common_options(no_format_option=True)
+def edit_command():
+    """
+    Executor for `globus config edit`
+    """
+    editor = os.environ.get('EDITOR', 'nano')
+    subprocess.call([editor, os.path.expanduser('~/.globus.cfg')])
```
bd4: Should this use `notepad.exe` on windows?

bd4:
The common option decorator stuff is hairy, at least needs some comments. No standard way of doing this with click?

globus_cli/config_command/edit.py
sirosen: I don't know if the SDK will even work on Windows yet.
I'm not sure that it's a target platform for the CLI?

Until we have a more clearly established stance on Windows (and I have a testing setup for it), I'm not sure I want to add windows-specific features.

sirosen:
@bd4 I agree that it's hairy (really hairy), but this is about the best that we've got to date.
Defining a custom decorator which attaches multiple options is basically the standard way of doing this in `click`.

You can handle options as non-decorator objects, but when it comes time to attach them you'll probably end up stuffing them back inside a decorator anyway.
I'm not sure that there's a way of polishing this turd any further than I have -- if options are attached to functions, then they need to be wrapped in a decorator to be syntactically clean.

---

A lot of it is working around the decorator syntax in general though, and is the best way of doing this particular pattern that I've seen.

The issue is that I find this ugly and confusing (you may disagree):
```python
@common_options()
def funcA(...):
    ...
```

I would much rather write
```python
@common_options
def funcB(...):
    ...
```

Requiring those extra parens by not accepting a function directly to the decorator results in confusing stacktraces when someone (who doesn't know what they are and why they're required) messes up and writes the latter form. It's a stacktrace from the decorator application, not function invocation, and likely to confuse.
It also seems fussy to require empty parens to a decorator -- as a caller, I either want to be giving it special arguments or not.

Either you need empty parens whenever you use the base decorator, which looks odd and confusing, or you specially handle the case of a single positional argument.
I'd rather put the ugly on the inside than the outside, though i see the argument against all of this fanciness as well.

(Hey, at least I didn't write a decorator decorator to handle this stuff! :stuck_out_tongue_winking_eye: )

---

I agree that it needs good/thorough comments, but I thought that what I had was sufficient.
Should there be more on it? If so, could you give some direction as to where it's confusing?

globus_cli/config_command/edit.py
bjmc: You can just have [Ubuntu userland on Windows 10 kernel now.](http://www.zdnet.com/article/microsoft-and-canonical-partner-to-bring-ubuntu-to-windows-10/)

bjmc:
@sirosen @bd4 I don't have all the context here, but if you're doing fancy stuff with decorators, and you're worried about getting ugly, hard-to-interpret stack traces, you might want to look at [`functools.wraps`](https://docs.python.org/2/library/functools.html#functools.wraps) that will preserve the original function name and some other metadata to make traces a little more comprehensible.

globus_cli/config_command/edit.py
sirosen: That's a good point, actually. If the Ubuntu shell environment they provide is good enough, we might start treating that as our path to Windows support.
There's the lunch meeting today, maybe we discuss a bit then.

sirosen:
@bjmc The trouble in this case is a matter of how intelligible the decorator itself is to developers working on the project. If you write this the less complex way, there's a chance that application itself fails (i.e. at import time) when a developer misuses it, which `wraps` doesn't change. Adding more complexity here allows for better errors to be generated, since they're custom exceptions rather than syntax errors for bad arguments to a decorator.
It seems every messy problem I end up producing here or in the SDK is decorator-related -- maybe I'm overusing them and trying too hard to make "composable" python.

@bd4 Do we want more added to the comments? We discussed face-to-face a little bit, and it seems like maybe this stuff is safe to merge? I'm not clear on needed work to clean this up at this point.

bjmc:
> maybe I'm overusing them and trying too hard to make "composable" python.

Perhaps. You might be trapped in that uncanny valley where you're using advanced language features to try and make solving some problem easier, but you haven't fully solved the problem in a way that lets you hide all the complexity behind the scenes and have it "just work" like magic for other developers using your thing. That can be a bad place to be and you have to choose between:

* Backing off and trying to make it simpler and more explicit

* Pressing forward: If gasoline isn't doing the job, have you tried napalm? More magic: Make a class-based decorator? Use context managers instead of decorators? Can you abuse metaclasses?

* Just document all the ugliness really extensively so that when somebody steps on the rakes, you can be like "Didn't you read the sign? CAUTION: RAKES"

If you're trying to avoid having to use `@common_options()` why don't you just check `if callable(args[0])` inside `common_options` and special-case that situation?

sirosen:
@bjmc I'm doing something pretty close to that last suggestion, special casing `len(args) == 1`, and insisting that either I get one positional arg, OR I get a set of keyword args. I think I should be checking `callable(args[0])` though -- I'm not doing so now and it "feels wrong", so I'll add that.

Upon a re-read, I think my comments are insufficient. Going to beef that up a bit.

bd4:
I do wonder if a non-decorator implementation would be simpler, e.g. every command function has to explicitly call `parse_common_options()` at the top. I don't have that strong an opinion though - I'm more concerned about the external interface, we can always refactor internals later if they prove to be difficult to maintain.

PR #22, opened 2016-09-12
Corrected instructions for bleeding-edge version

corpulentcoffee:
Use `git clone` for cloning the repository rather than `svn`-style `checkout`, which won't work.


sirosen:
I know we fixed this for the SDK. I thought I did the CLI instructions as well at that time.
Thanks, merging.

PR #23, opened 2016-09-12
Python3-friendly re-raise exception in custom hook

corpulentcoffee:
The `raise exception_type, exception, traceback` form does not work in Python 3, but we can use the `reraise` compatibility function from Six to support either.


sirosen:
Tested on 2.x.
Seems to work, merging.

Thanks for the fix. :+1:

PR #32, opened 2016-10-09
Add --map-http-status Flag

### LINKED ISSUE #21: Add --map-http-status global option to convert HTTP statuses into nonzero exit codes
Karl suggested this a while back, and I've had it in my queue for a while.
I definitely want this before we consider the CLI done, but I'm not sure when we'll get to it.

Several commands have a set of non-200 HTTP responses that might still be considered part of normal operations. For example, a command like `globus transfer endpoint delete ...` could be considered successful on a 200 or a 404.

It may be desirable for us to hand that information -- the 404 response, vs. a 403 or other error, to the caller, but doing so with some fixed mapping is burdensome for scripting (you need to know our statuses table), and not very extensible if we omit codes that we start using later.
The generic solution is to just let the invoking script specify what it would like us to translate various HTTP statuses into.
That lets them get meaningful status codes, avoids a magical lookup table, and makes those scripts more readable because they specify what the various error codes that they're dispatching on mean / come from.

So, the basic form would be something like this:
```bash
globus transfer endpoint delete --map-http-status "404=54" ...
```

and we can even allow crazy things like this check for nonexistence:
```bash
globus transfer endpoint show --map-http-status "404=0" --map-http-status "200=1"
```

In the original version of this proposal the idea was to restrict the caller to using the range from 50 to 99 (reserving other exit codes for our own purposes into the future).
However, I think there's a case for allowing clients to say "allow 404s" by mapping to 0.
Whether or not we allow mapping to 1 is a little bit more tenuous, but I think still acceptable, since it just means "generic error".

There's also a question of whether or not it should be
```bash
globus transfer endpoint delete --map-http-status "404=54" --map-http-status "403=53"
```
or
```bash
globus transfer endpoint delete --map-http-status "404=54,403=53"
```

That second is a little bit nicer, but also a little harder to implement.

sirosen:
I don't think this is the final version of the flag, or the last word on this by any means, but it's a good first cut.
Supports fairly "robust" parsing in the form of multiple instances of the flag and multiple values given to a single instance of the flag (best of both approaches mentioned in #21).

As mentioned in the commit messages, there's a big question about how well the semantics of this will play with auxiliary calls to the services like identity mapping or autoactivation.
Ideally, this would be hooked in more directly on the error or response objects created by the "main" calls of the various CLI commands. However, that will have to wait for a future effort.

Resolves #21 for now

### DIFF
--- a/globus_cli/parsing/command_state.py
+++ b/globus_cli/parsing/command_state.py
@@ -17,6 +17,8 @@ class CommandState(object):
         self.output_format = config.get_output_format() or TEXT_FORMAT
         # default is always False
         self.debug = False
+        # by default, empty dict
+        self.http_status_map = {}
 
     def outformat_is_text(self):
         return self.output_format == TEXT_FORMAT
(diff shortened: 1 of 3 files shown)


corpulentcoffee:
I haven't bisected this yet, but I think this might not work on Python 3.

```
  File "/home/mayor/Repos/globus-cli/.venv3/lib/python3.5/site-packages/globus_cli-0.4.2.0-py3.5.egg/globus_cli/parsing/command_state.py", line 63, in map_http_status_option
TypeError: can only concatenate list (not "range") to list
```

corpulentcoffee:
So I think in Python 3, `range()` returns a generator, so `[0, 1] + range(50, 100)` won't work.

sirosen:
Ahh! Can we add `globus --help` to the tests to protect against catastrophic "parser building" failures like this?

I don't want to try to commit from mobile, but if you listify that range I'm comfortable merging it.

corpulentcoffee:
Yes, that's a good idea. I'll look at that.

Issue #26, opened 2016-09-16
Investigate moving off `six`, along with SDK


sirosen:
Per globus/globus-sdk-python#51 , we're trying to get off of the dependency on `six` because OSX makes things difficult.
This is largely matter of supporting the naive developers who don't realize that there's a strong reason to use virtualenv.

At present, we're using

- `six.string_types` (easy to sub with some version dispatch)
- `six.reraise`

The `reraise` functionality is hard to imitate, so we'll need to take another look at where and how that's happening. I don't think we can do the current exception handling logic without a `reraise`, but maybe we don't need to preserve context (I think that's just paranoia about how `click` might behave).

corpulentcoffee:
The `reraise` implementation in Six is a [little gross](https://bitbucket.org/gutworth/six/src/ca4580a5a648fc75abc568907e81abc80b05d58c/six.py?at=default&fileviewer=file-view-default#six.py-680), I guess because the Python 2.7-style `raise x, y, z` syntax won't even parse under Python 3 (i.e. it is an uncatchable `SyntaxError` there).

sirosen:
@jaswilli @corpulentcoffee I think we need to reopen this discussion.

Karl found that a global install fails on Ubuntu 14.04 because the Debian package `python-six` ships with v1.5.6
That package is depended upon by a bunch of other Ubuntu packages, and it's all a sticky ball of mud.

Can we consider what it would be like if we wrote with `six>=1.4.1,<2.0.0`?
I know we'd have to handle the ConfigParser location change ourselves, but I think that's it.

I'm also okay with just putting six in tree (in the SDK), since it has no dependencies. However, I think that's actually significantly more work than just expanding the set of versions we support.
We already cast a pretty wide net on `requests` because people are likely to have that installed and we don't need any advanced features, so this would be "consistent" with that, FWIW.

bjmc:
You could look at https://pypi.python.org/pypi/future as an alternative to six.

corpulentcoffee:
Closed by globus/globus-cli#41

PR #41, opened 2016-10-13
Update installation instructions: always use venv

sirosen:
These new install instructions _should_ let us neatly sidestep the issues with global installs without putting too much burden on the user.

Recommendation for review: actually pull down this branch and open `file:///.../globus-cli/docs/index.html` to see how it "looks and feels".


jaswilli:
I tested this on OS X and it works fine. I also tested on Windows (10) and it works, except the instructions need to be changed for updating the path, obviously.

Also, I think if we go this route we're going to need to provide "Updating" and "Uninstalling" documentation.

sirosen:
@jaswilli If the below instructions are acceptable, let's add them. If they are not, can we punt on this as a problem for another day?

```rst
Update
------
To update your version of the CLI to the latest:

.. code-block:: bash

    source $HOME/.globus-cli-virtualenv/bin/activate
    pip install -U globus-cli
    deactivate

Uninstall
---------

To remove the CLI:

.. code-block:: bash

    rm -r $HOME/.globus-cli-virtualenv

You should also edit your ``$HOME/.bashrc`` and remove the line that reads ``export PATH="$PATH:$HOME/.globus-cli-virtualenv/bin"``.
```

jaswilli:
Yes, those look reasonable (keeping in mind that they need to be amended for Windows).

sirosen:
I think we can safely say that this closes #26 and globus/globus-sdk-python#51

corpulentcoffee:
This looks very much improved to me.

Very stupid quibble: can we use either `~` or `$HOME` consistently, instead of switching between the two? (As for which one, I think people are used to seeing `~` in terminal snippets and are used to typing it in their terminal. I don't know anyone who types out `$HOME` in interactive commands.)

sirosen:
I'd like to use `$HOME` everywhere, but `~` sometimes slips in when I'm not paying attention.
I agree that `~` is more idiomatic / "comfortable", but it's not quite as safe.

The reasoning is simply that shells respect `"~"` by not expanding, so, if someone copies a dirname and quotes it, bad times ensue.
I'll amend to have no `~`s.

corpulentcoffee:
Yes, that is true.

jaswilli:
Also, FWIW, `$HOME` works in Powershell but `~` does not.

corpulentcoffee:
That's another good reason.

sirosen:
I also realized I needed to quote `$HOME` to avoid the possibility of word-splitting if someone's home has spaces in it. That actually matters, on account of the `rm -r` command offered up.

PR #45, opened 2016-10-13
Support for new role features in Transfer 5.3

corpulentcoffee:
Adds option support for these new features in the roles expansion in Transfer 5.3:

- the `administered-by-me` endpoint search scope for `globus transfer endpoint search`
- Administrator and activity manager/monitor roles for `globus transfer endpoint role create`. This is a breaking change if a caller was previously depending upon not having to pass `--role=access_manager`, as the `--role` option is now required and has no default.


sirosen:
When we actually start caring about people's scripts breaking, we'll have to be more careful with newly required options.
With the current state of the world, it's totally fine though, obviously.

PR #48, opened 2016-10-14
Fix non-batch globus transfer async-delete --path

corpulentcoffee:
On the current release, `globus transfer async-delete --path=xxx` does not work, and instead behaves as if the user did `--batch`.

Adjusts the check to be just `if batch` rather than `if batch is not None` since `batch` will be either `True` or `False`, not `None`. This is consistent with the guard check for `batch` at the top of this command's implementation, which is just a truthy/falsey check.


sirosen:
Excellent catch.

PR #51, opened 2016-10-14
Update `--version` output to include latest

### LINKED ISSUE #43: Show latest version on output from --version
If there's a reasonable way to do it, we could pull the latest published version from pypi, otherwise we could go by GitHub releases.

Example:
```
jasons-mbp:src jtw$ vagrant version
Installed Version: 1.8.6
Latest Version: 1.8.6
 
You're running an up-to-date version of Vagrant!
```

sirosen:
Shows latest version, prints a simple blurb telling out-of-date users to install the latest.

Adds direct dependency on `requests` (already needed by SDK).
Closes #43


corpulentcoffee:
I'm shocked you didn't go for `StrictVersion`.

From [PEP 386](https://www.python.org/dev/peps/pep-0386):
> The StrictVersion class [...] Version numbering for meticulous retentive and software idealists.

:smirk:

sirosen:
`StrictVersion` doesn't support 4-digit numbers. Which I think is somewhat silly -- it's not like 4 digits is going totally hog-wild with nouveau versioning.

sirosen:
Apologies for the build failures -- I had to amend to defer an import a little bit.
My original version was trying to `import requests` at setup time (which explains why it worked on my own machine, but not on Travis).

globus_cli/parsing/shared_options.py
```diff
@@ -1,10 +1,35 @@
...
+    """
+    def callback(ctx, param, value):
+        latest, current = get_versions()
+        click.echo(('Installed Version: {0}\n'
+                    'Latest Version:    {1}\n'
+                    '\n{2}').format(
+                        current, latest,
+                        'You are running the latest version of the Globus CLI'
+                        if current == latest else
+                        'You should update your version of the Globus CLI'
```
corpulentcoffee: Maybe throw a link to https://globus.github.io/globus-cli/#updating-removing-the-cli here so people know how.

globus_cli/version.py
```diff
@@ -22,3 +23,22 @@
...
+    command runs (yuck!)
+
+    Also protects import of `requests` from issues when grabbed by setuptools.
+    More on that inline
+    """
+    # import in the func (rather than top-level scope) so that at setup time,
+    # `requests` isn't required -- otherwise, setuptools will fail to run
+    # because requests isn't installed yet.
+    import requests
+    _version_data = requests.get(
```
corpulentcoffee: I'm assuming that you have this prefixed with an underscore because originally it was module-level (i.e. it didn't used to be wrapped in a function). You could get rid of that now if you wanted.

sirosen:
I used a very slightly different link because it points them to an explicit sphinx ref in that page.
If we change the text of the section heading, the link won't break as long as we persist that ref.

PR #53, opened 2016-10-15
Let `transfer ls` default to default_directory

### LINKED ISSUE #49: Allow the Transfer service to figure out what the default path should be when `ls`ing
Instead of having the `globus transfer ls` command default to `/~/`, don't send any `path` to the Transfer service and let the Transfer service figure out what the default path is supposed to be.

Fixes S3 endpoints (which don't support `/~/`), shared endpoints (which should usually default to `/`), and any kind of endpoint that has had a `default_directory` configured.

sirosen:
Simply remove the `path` parameter unless we have a value for it.

Use kwarg-selection to only sometimes pass it to the SDK.
(I think the SDK handles path=None by removing it too, but this way is safest.)

Fixes #49

### DIFF
--- a/globus_cli/services/transfer/ls.py
+++ b/globus_cli/services/transfer/ls.py
@@ -34,9 +34,13 @@ def _get_ls_res(client, path, endpoint_id, recursive, depth):
     Note about paths: because the "path" key always has a trailing slash, and
     item names never have leading slashes, we can just string concat.
     """
+    ls_kwargs = {}
+    if path is not None:
+        ls_kwargs = {'path': path}
+
     # non-recursive ls is simple -- just make the call and return the result
     if not recursive:
-        return client.operation_ls(endpoint_id, path=path)
+        return client.operation_ls(endpoint_id, **ls_kwargs)
 
     # the rest of this is the recursive case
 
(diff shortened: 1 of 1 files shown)


globus_cli/services/transfer/ls.py
```diff
@@ -46,11 +50,11 @@ def _get_ls_res(client, path, endpoint_id, recursive, depth):
 
     if depth >= 0:
         # do an `ls` call against current path, use it to seed the result_doc
-        res = client.operation_ls(endpoint_id, path=path)
+        res = client.operation_ls(endpoint_id, **ls_kwargs)
         result_doc['DATA'] = [item for item in res]
         # set path in order to get trailing slash, if necessary
         result_doc['path'] = res['path']
-        path = res['path']
+        path = ls_kwargs['path'] = res['path']
```
corpulentcoffee: Why do you need to assign back to `ls_kwargs` here? Aren't you already done with `ls_kwargs` at this point?

sirosen: It's not necessary, but I added it before realizing that.

It seemed fine to leave in, but maybe it's confusing -- would you like me to amend and remove it?

corpulentcoffee: I was confused by it, so maybe just drop it out.

PR #76, opened 2016-10-17
[HOLD] Add command aliases as a type of click.Command

### LINKED ISSUE #71: Command heirarchy/grouping
Do we want to promote certain commands to be top-level commands? Example: "endpoint search" is currently `globus transfer endpoint search`. Should it be `globus search`?

Other commands where this would make sense?

sirosen:
The aliases are constructed in terms of a rename and an existing command object, and assign nearly all of the same attributes to the alias command. The only exception is the name, which is obviously altered in the process.
Importantly, this is a type of `click.Command`, and is not generic enough to safely apply to a `Group` or `MultiCommand` -- the attributes for those types are different, and the passthrough of original command attributes to the new command's constructor will not cover the differences.

The `AliasCommand` importantly inherits the original commands params, so it does not need to blindly forward options as unprocessed values. That saves us from some of the concerns over doing that.

Using this, it was pretty trivial to add
- `globus endpoint-search` -> `globus transfer endpoint search`
- `globus endpoint-ls` -> `globus transfer ls`

Resolves #71

### DIFF
--- a/globus_cli/parsing/__init__.py
+++ b/globus_cli/parsing/__init__.py
@@ -2,6 +2,7 @@ from globus_cli.parsing.main_command_decorator import globus_main_func
 
 from globus_cli.parsing.case_insensitive_choice import CaseInsensitiveChoice
 from globus_cli.parsing.hidden_option import HiddenOption
+from globus_cli.parsing.aliases import AliasCommand
 
 from globus_cli.parsing.shared_options import (
     common_options,
@@ -15,6 +16,7 @@ __all__ = [
 
     'CaseInsensitiveChoice',
     'HiddenOption',
+    'AliasCommand',
 
     'common_options',
     # Transfer options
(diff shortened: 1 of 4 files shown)


corpulentcoffee:
It might be nice if `globus [subcommand(s)] --help` and `globus list-commands` listed these separately, perhaps indicating what the full command is that the aliases forward to.

Could something like this be used to alias `globus transfer share show` to `globus transfer endpoint show` and `globus transfer share delete` to `globus transfer endpoint delete`?

sirosen:
Marking this as "blocked: design".

After a quick chat with Dave, we agreed that this should be held until we're clear on the direction for aliasing and the command tree in general.
Providing these aliases with better display in helptext and `list-commands` will require custom helptext, which is a very significant undertaking, and it's not obvious that the aliases are desirable if they damage the display of help.

sirosen:
With the command hierarchy changes, I think it's safe to scrap this. Closing.

Issue #35, opened 2016-10-10
Enumerate and address all cases of commands where `--format TEXT` is the JSON output


sirosen:
There are a number of commands which I took shortcuts on before I had complete and working tooling for table output.
To get things working without fiddling with output too much, I used the JSON output for both output formats.

In some cases, this may ultimately be the most sensible approach (probably not).
Regardless, we should list all of these as items to address before considering the CLI "done".

corpulentcoffee:
Items I know of where the text version is just the JSON version:

- `globus transfer task show`
- `globus transfer task pause-info`

Related to this, there are some cases where the text version is too terse, missing critical fields that are only present in the JSON version:

- `globus transfer endpoint show` (e.g. host endpoint ID when showing a shared endpoint)
- `globus transfer share show` (ditto)

aaschaer:
These should now all be covered by #251

PR #77, opened 2016-10-18
Text output for `globus transfer task show`

corpulentcoffee:
Adds support for the regular text mode output of a task retrieved via `globus transfer task show`. Fields vary depending on what kind of task it is and whether it is completed or not.

Toward #35.


sirosen:
I like this pattern for printing variable fields. Looks good to me.

Issue #64, opened 2016-10-17
Transfer: does endpoint-type=s3 work for endpoint create?


jaswilli:
Comment from Karl:

> Didn’t think bryce hooked that up, thought CLI was the only way.  (Code search shows create_s3_endpoint() is only called by CLI).  Requires URL and “is_owner_activated(bool)”.

sirosen:
Switching this to a bug -- it does not work, and we shouldn't include the option until it does.

PR #80, opened 2016-10-18
Improve error message for S3 endpoint creation

corpulentcoffee:
With regards to #64...

- we already error on `--endpoint-type=s3`
- the (current) process for setting up an S3 endpoint is relatively involved and requires the legacy CLI
- this it to be replaced with the new S3 DSI at some point soon, so support is unlikely to ever land in Transfer's current REST API

... so, what if we just improve the error message until S3 endpoints can be created natively via the new CLI?

Old:

```
(.venv3) ~/Repos/globus-cli$ globus transfer endpoint create --endpoint-type=s3 --display-name=x
NotImplementedError: S3 Endpoints cannot be created through the new CLI yet.
```

Proposed:

```
(.venv3) ~/Repos/globus-cli$ globus transfer endpoint create --endpoint-type=s3 --display-name=x
Error: At this time, S3-backed endpoints can only be created via the legacy
hosted Globus CLI.

For more information, see:
- https://docs.globus.org/cli/using-the-cli
- https://docs.globus.org/how-to/amazon-aws-s3-endpoints
```

### DIFF
--- a/globus_cli/services/transfer/endpoint/create.py
+++ b/globus_cli/services/transfer/endpoint/create.py
@@ -35,8 +35,14 @@ def endpoint_create(endpoint_type, display_name, description, organization,
         myproxy_server=myproxy_server, myproxy_dn=myproxy_dn,
         oauth_server=oauth_server)
     if endpoint_type == 's3':
-        raise NotImplementedError(
-            'S3 Endpoints cannot be created through the new CLI yet.')
+        raise click.ClickException(
+            'At this time, S3-backed endpoints can only be created via the '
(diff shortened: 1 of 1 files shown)


jaswilli:
Seems reasonable to me 👍

sirosen:
My only worry is directing people to `https://docs.globus.org/cli/using-the-cli`.
Won't that be confusing? Maybe it's okay. Merging.

jaswilli:
Yeah, I guess that's a good point.

I think it's good enough though, especially considering the number of S3 endpoint users we're likely to encounter.

Issue #40, opened 2016-10-11
Generic error handlers are exiting with status 0, not status 1


sirosen:
This was something found in #32 and fixed there.
However, in case that work is cancelled, delayed, or significantly altered, we need to be sure to fix this issue.

When the custom excepthook was replaced with a high-level dispatch on exception types, I did not correctly add explicit `exit(1)` calls as I should have.

sirosen:
With #32 closed, this is also done.

PR #84, opened 2016-10-18
Introduce `globus transfer bookmark locate`

### LINKED ISSUE #74: Add a `bookmark show` option suitable for use in a subshell
This is dependent on #73 

There should be a `bookmark show` option which produces text output suitable for use as an `EndpointPlusPath`.

Open question: should this be a case of a special `--format SUBSHELL` or similar `--format` customization, or should it be a special-case option.

corpulentcoffee:
This is sort of what I'm thinking for #74.

Notes/questions:

- I arbitrarily did this as a special-case option rather than a `--format` switch (which is an "open question" in #74) only because I think the syntax for subshells is already noisy enough and I think it will be easier to look at when used in practice
- I went ahead and implemented this new command using our new proposed positional argument style instead of requiring the user to type `--bookmark-id`/`--id` or `--name`, so this is one possible pattern for handling that ambiguity, but I'm not wedded to it
- given the use case for this, I really want a non-zero status code for this if it fails, so
  - is `click.get_current_context().exit(1)` best practice from our commands when we willfully want to exit with a non-zero status code?
  - when an uncaught exception happens, e.g. an internal server error from Transfer, a zero status code seems to be returned... why is this?
- do we need to do any special quoting, e.g. for a path that includes spaces? (I *think* the answer is "no", assuming that people quote properly, like `globus transfer ls "$(globus transfer bookmark locate 'My Bookmark')"`)

Closes #74

### DIFF
--- a/globus_cli/services/transfer/bookmark/locate.py
+++ b/globus_cli/services/transfer/bookmark/locate.py
@@ -23,18 +23,19 @@ def bookmark_locate(bookmark_id_or_name):
     # service outright forbids it for bookmark names, so we can strip it off
     bookmark_id_or_name = bookmark_id_or_name.strip()
 
+    res = None
     try:
         UUID(bookmark_id_or_name)  # raises ValueError if argument not a UUID
-        res = client.get_bookmark(bookmark_id_or_name.lower())
+    except ValueError:
+        pass
+    else:
+        try:
+            res = client.get_bookmark(bookmark_id_or_name.lower())
+        except TransferAPIError as exception:
+            if exception.code != 'BookmarkNotFound':
+                raise
 
-    except (ValueError, TransferAPIError) as ex:
-        if isinstance(ex, TransferAPIError) and ex.code != 'BookmarkNotFound':
-            raise
-
-        # either the user provided a string that isn't a UUID (ValueError) or
-        # the provided UUID doesn't identity a bookmark (TransferAPIError with
-        # code='BookmarkNotFound'); fallback to matching a bookmark by name
-
+    if not res:  # non-UUID input or UUID not found; fallback to match by name
         try:
             # n.b. case matters to the Transfer service for bookmark names, so
             # two bookmarks can exist whose names vary only by their case
--- a/globus_cli/services/transfer/bookmark/locate.py
+++ b/globus_cli/services/transfer/bookmark/locate.py
@@ -1,0 +1,49 @@
+from uuid import UUID
+
+import click
+
+from globus_sdk.exc import TransferAPIError
+from globus_cli.safeio import safeprint
+from globus_cli.parsing import common_options
+from globus_cli.services.transfer.helpers import get_client
+
+
+@click.command('locate', help='Output <UUID>:<path> of a bookmark; useful in '
+                              'a subshell as input to another command')
+@common_options
+@click.argument('bookmark_id_or_name', required=True)
+def bookmark_locate(bookmark_id_or_name):
+    """
+    Executor for `globus transfer bookmark locate`
+    """
+
+    client = get_client()
+
+    # leading/trailing whitespace doesn't make sense for UUIDs and the Transfer
+    # service outright forbids it for bookmark names, so we can strip it off
+    bookmark_id_or_name = bookmark_id_or_name.strip()
+
+    try:
+        UUID(bookmark_id_or_name)  # raises ValueError if argument not a UUID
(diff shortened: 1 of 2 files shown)


sirosen:
The uncaught exception issue is #40 , but it's also handled by #21 , which is why it hasn't been fixed separately.

I'm pretty sure special quoting is not needed -- we can revisit if it's an issue.

WRT the `--format` option, I think this is probably the right call.

The `get_current_context().exit(1)` is what we are supposed to be doing, and moving forward we should do it. Some places have bare `sys.exit(...)`s, and I don't think it's important enough to go back and retrofit those.

Reviewing the actual code now.

globus_cli/services/transfer/bookmark/locate.py
```diff
@@ -0,0 +1,49 @@
...
+    # leading/trailing whitespace doesn't make sense for UUIDs and the Transfer
+    # service outright forbids it for bookmark names, so we can strip it off
+    bookmark_id_or_name = bookmark_id_or_name.strip()
+
+    try:
+        UUID(bookmark_id_or_name)  # raises ValueError if argument not a UUID
+        res = client.get_bookmark(bookmark_id_or_name.lower())
+
+    except (ValueError, TransferAPIError) as ex:
+        if isinstance(ex, TransferAPIError) and ex.code != 'BookmarkNotFound':
```
sirosen: Can this be refactored into multiple `except` clauses? `isinstance(e, ExceptionType)` is something I like to avoid when possible.
It also avoids nesting the other `try-except`, as in
```python
...
except ValueError:
    pass
except TransferAPIError as ex:
    if ex.code != 'BookmarkNotFound':
        raise

try:
    ...
```

sirosen:
With the exception of the exception cleanup I'd like, this is good to merge.

Also, I'm not wedded to that change if there's a reason to keep it the way it is instead.

globus_cli/services/transfer/bookmark/locate.py
corpulentcoffee: Yes, we can totally do something different.

I pushed b9260ae for a revised structure. The revised structure has the bonus of being a little bit safer in the event that `client.get_bookmark(bookmark_id_or_name.lower())` raised a `ValueError`, in which case the previous structure would have silenced that error.

If you're happy with the revised structure, I can squash it.

sirosen:
Thanks much. You would be right to point out that the nesting isn't actually any better as it turns out, but I think it's cleaner.
Merging.

globus_cli/services/transfer/bookmark/locate.py
```diff
@@ -0,0 +1,49 @@
...
+def bookmark_locate(bookmark_id_or_name):
+    """
+    Executor for `globus transfer bookmark locate`
+    """
+
+    client = get_client()
+
+    # leading/trailing whitespace doesn't make sense for UUIDs and the Transfer
+    # service outright forbids it for bookmark names, so we can strip it off
+    bookmark_id_or_name = bookmark_id_or_name.strip()
```
corpulentcoffee: I put that in there because I figured a very common case for this might be taking user input for a bookmark name, and using a bookmark name would *necessitate* quoting the user input (e.g. `globus transfer bookmark locate "$someUserInput"`), and I thought it would be super if the CLI did the right thing.

*But*, that said, maybe this is over-robust/over-polite *or* maybe we should refactor thing kind of thing into parameter type that knows to do stripping. Or maybe that's overengineering. Either way, if you don't want it I'll drop it out.

sirosen:
By the sound of things I may have jumped the gun on merging this -- if you meant to squash those commits, sorry for persisting noisy things.

If we find this kind of whitespace stripping to be generally desirable, I would say we should handle it in a parameter callback -- those are very easy to share and chain on the fly.
However, and this is important, `ParamType.convert()` happens _before_ `Parameter.callback`.
That means that an argument or option with `callback=strip_whitespace`, `type=click.INT` will insist on not seeing whitespace around the int (because that enforcement comes earlier).

If we really want to strip whitespace prior to other kinds of type enforcement (e.x. allow spaces around a UUID), it might be best to define `WithoutWhitespace` as a type that accepts, as one of its parameters, another `ParamType` to which it will chain all of its methods.
Otherwise, even if you make it a mixin class, you'll be defining new types all over the place.
That's somewhat risky because they actually have some [irritating subtleties](http://click.pocoo.org/6/api/#click.ParamType).

Issue #73, opened 2016-10-17
Add EndpointPlusPath Parameter Type


sirosen:
This type is meant to be part of the move away from required options for things like `--endpoint-id` which are very often required by commands.

We will accept inputs of the form `<endpoint_id>:<path>`.

Paths can be absolute (starts with slash), or relative to the default directory of an endpoint.

sirosen:
The parameter type should also be able to enforce `path_required=True` and `path_required=False`.

Issue #72, opened 2016-10-17
Enforce UUID Type on all Endpoint IDs


sirosen:
Currently, canonical names can make their way through the CLI and to the backend.

There's no need to support this into the future, so for the sake of future iterations of the CLI, we should strictly enforce that all endpoint IDs are UUIDs.

PR #86, opened 2016-10-18
Implement EndpointPlusPath Type, apply it in all "easy" scenarios, enforce Endpoint IDs being UUIDs

sirosen:
This is a new `click.ParamType`.
Unfortunately, a little bit of hackery is needed because of a bad behavior (which I call a bug) in `click` ( https://github.com/pallets/click/issues/674 ), so in some locations the parameter type is not used inline, but defined at the top of the file. This is actually fine, but a little annoying.

Also converts all obvious cases of `--endpoint-id` plus `--path` to this type.
Some commands, like shared endpoint creation, could use this type but (for now at least) don't.

I consider this sufficient to conclude #73, and we can work on #72 separately as an addition on top of this.

### DIFF
--- a/globus_cli/parsing/endpoint_plus_path.py
+++ b/globus_cli/parsing/endpoint_plus_path.py
@@ -47,7 +47,7 @@ class EndpointPlusPath(click.ParamType):
         # split the value on the first colon, leave the rest intact
         splitval = value.split(':', 1)
         # first element is the endpoint_id
-        endpoint_id = splitval[0]
+        endpoint_id = click.UUID(splitval[0])
 
         # get the second element, defaulting to `None` if there was no colon in
         # the original value
@@ -63,3 +63,7 @@ class EndpointPlusPath(click.ParamType):
             self.fail('The path component is required', param=param)
 
         return (endpoint_id, path)
+
+
+ENDPOINT_PLUS_OPTPATH = EndpointPlusPath(path_required=False)
+ENDPOINT_PLUS_REQPATH = EndpointPlusPath(path_required=True)
--- a/globus_cli/parsing/endpoint_plus_path.py
+++ b/globus_cli/parsing/endpoint_plus_path.py
@@ -1,0 +1,65 @@
+import click
+
+
+class EndpointPlusPath(click.ParamType):
+    """
+    Custom type for "<endpoint_id>:<path>"
+    Supports path being required and path being optional.
+
+    Always produces a Tuple.
+    """
+    name = 'endpoint plus path'
+
+    def __init__(self, *args, **kwargs):
+        # path requirement defaults to True, but can be tweaked by a kwarg
+        self.path_required = kwargs.pop('path_required', True)
+
+        super(EndpointPlusPath, self).__init__(*args, **kwargs)
+
+    def get_metavar(self, param):
+        """
+        Default metavar for this instance of the type.
+        """
+        return self.metavar
+
+    @property
+    def metavar(self):
+        """
+        Metavar as a property, to satisfy the slight brokenness of
+        click.Argument
+        Tracked against Click as an issue:
+        https://github.com/pallets/click/issues/674
+        """
+        if self.path_required:
+            return "ENDPOINT_ID:PATH"
+        else:
+            return "ENDPOINT_ID[:PATH]"
+
+    def convert(self, value, param, ctx):
+        """
+        ParamType.convert() is the actual processing method that takes a
+        provided parameter and parses it.
+        """
+        # passthrough conditions: None or already processed
+        if value is None or isinstance(value, tuple):
+            return value
+
+        # split the value on the first colon, leave the rest intact
+        splitval = value.split(':', 1)
+        # first element is the endpoint_id
+        endpoint_id = splitval[0]
+
+        # get the second element, defaulting to `None` if there was no colon in
+        # the original value
+        try:
+            path = splitval[1]
+        except IndexError:
+            path = None
+        # coerce path="" to path=None
+        # means that we treat "enpdoint_id" and "endpoint_id:" equivalently
+        path = path or None
+
+        if path is None and self.path_required:
+            self.fail('The path component is required', param=param)
+
+        return (endpoint_id, path)
--- a/globus_cli/services/transfer/async_delete.py
+++ b/globus_cli/services/transfer/async_delete.py
@@ -4,15 +4,13 @@ from globus_sdk import DeleteData
 
 
 from globus_cli.parsing import (
-    common_options, submission_id_option, EndpointPlusPath)
+    common_options, submission_id_option, ENDPOINT_PLUS_OPTPATH)
 from globus_cli.helpers import (
     outformat_is_json, print_json_response, colon_formatted_print)
 
 from globus_cli.services.transfer.helpers import (
     get_client, shlex_process_stdin)
 from globus_cli.services.transfer.activation import autoactivate
-
-path_type = EndpointPlusPath(path_required=False)
 
 
 @click.command('async-delete', short_help='Submit a Delete Task',
@@ -28,8 +26,8 @@ path_type = EndpointPlusPath(path_required=False)
               help=('Accept a batch of paths on stdin (i.e. run in '
                     'batchmode). Uses ENDPOINT_ID as passed on the '
                     'commandline, and includes any commandline PATH given'))
-@click.argument('endpoint_plus_path', metavar=path_type.metavar,
-                type=path_type)
+@click.argument('endpoint_plus_path', metavar=ENDPOINT_PLUS_OPTPATH.metavar,
+                type=ENDPOINT_PLUS_OPTPATH)
 def async_delete_command(batch, ignore_missing, recursive, endpoint_plus_path,
                          label, submission_id):
     """
--- a/globus_cli/services/transfer/async_delete.py
+++ b/globus_cli/services/transfer/async_delete.py
@@ -4,7 +4,7 @@ from globus_sdk import DeleteData
 
 
 from globus_cli.parsing import (
-    common_options, endpoint_id_option, submission_id_option)
+    common_options, submission_id_option, EndpointPlusPath)
 from globus_cli.helpers import (
     outformat_is_json, print_json_response, colon_formatted_print)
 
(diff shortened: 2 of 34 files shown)


globus_cli/parsing/endpoint_plus_path.py
```diff
@@ -0,0 +1,62 @@
...
+
+        try:
+            path = splitval[1]
+        except IndexError:
+            path = None
+            if self.path_required:
+                self.fail(('This command requires a path in addition to an '
+                           'endpoint ID. You must use the '
+                           '"<endpoint_id>:<path>" notation.'),
+                          param=param)
```
corpulentcoffee: By only checking `self.path_required` on an `IndexError` exception, you allow something like `globus transfer async-delete ddb59aef-6d04-11e5-ba46-22000b92c6ec:` to sneak through, which could definitely happen if someone combines two input variables before calling the CLI.

Maybe something like this would be nice:
```
        try:
            path = splitval[1]
        except IndexError:
            path = None
        if self.path_required and not path:
            self.fail('This command requires a path in addition to an '
                      'endpoint ID. You must use the '
                      '"<endpoint_id>:<path>" notation.',
                      param=param)
```

globus_cli/parsing/endpoint_plus_path.py
```diff
@@ -0,0 +1,63 @@
...
+        provided parameter and parses it.
+        """
+        # passthrough conditions: None or already processed
+        if value is None or isinstance(value, tuple):
+            return value
+
+        # split the value on the first colon, leave the rest intact
+        splitval = value.split(':', 1)
+        # first element is the endpoint_id
+        endpoint_id = click.UUID(splitval[0])
```
corpulentcoffee: Returning `endpoint_id` as a UUID class seems to breaks things, e.g. with `async-delete`, when it tries to activate the endpoint.

Python 2:
```
  File "build/bdist.linux-x86_64/egg/globus_cli/parsing/main_command_decorator.py", line 25, in invoke
  File "/home/mayor/Repos/globus-cli/.venv2/lib/python2.7/site-packages/click-6.6-py2.7.egg/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/mayor/Repos/globus-cli/.venv2/lib/python2.7/site-packages/click-6.6-py2.7.egg/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/mayor/Repos/globus-cli/.venv2/lib/python2.7/site-packages/click-6.6-py2.7.egg/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/mayor/Repos/globus-cli/.venv2/lib/python2.7/site-packages/click-6.6-py2.7.egg/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "build/bdist.linux-x86_64/egg/globus_cli/services/transfer/async_delete.py", line 41, in async_delete_command
  File "build/bdist.linux-x86_64/egg/globus_cli/services/transfer/activation.py", line 6, in autoactivate
  File "/home/mayor/Repos/globus-cli/.venv2/lib/python2.7/site-packages/globus_sdk-0.4.2-py2.7.egg/globus_sdk/transfer/client.py", line 286, in endpoint_autoactivate
    path = self.qjoin_path("endpoint", endpoint_id, "autoactivate")
  File "/home/mayor/Repos/globus-cli/.venv2/lib/python2.7/site-packages/globus_sdk-0.4.2-py2.7.egg/globus_sdk/base.py", line 95, in qjoin_path
    return "/" + "/".join(quote(part) for part in parts)
  File "/home/mayor/Repos/globus-cli/.venv2/lib/python2.7/site-packages/globus_sdk-0.4.2-py2.7.egg/globus_sdk/base.py", line 95, in <genexpr>
    return "/" + "/".join(quote(part) for part in parts)
  File "/usr/lib64/python2.7/urllib.py", line 1297, in quote
    if not s.rstrip(safe):
AttributeError: 'UUID' object has no attribute 'rstrip'
```

Python 3:
```
  File "/home/mayor/Repos/globus-cli/.venv3/lib/python3.5/site-packages/globus_cli-0.4.2.0-py3.5.egg/globus_cli/parsing/main_command_decorator.py", line 25, in invoke
  File "/home/mayor/Repos/globus-cli/.venv3/lib/python3.5/site-packages/click-6.6-py3.5.egg/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/mayor/Repos/globus-cli/.venv3/lib/python3.5/site-packages/click-6.6-py3.5.egg/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/mayor/Repos/globus-cli/.venv3/lib/python3.5/site-packages/click-6.6-py3.5.egg/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/mayor/Repos/globus-cli/.venv3/lib/python3.5/site-packages/click-6.6-py3.5.egg/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "/home/mayor/Repos/globus-cli/.venv3/lib/python3.5/site-packages/globus_cli-0.4.2.0-py3.5.egg/globus_cli/services/transfer/async_delete.py", line 41, in async_delete_command
  File "/home/mayor/Repos/globus-cli/.venv3/lib/python3.5/site-packages/globus_cli-0.4.2.0-py3.5.egg/globus_cli/services/transfer/activation.py", line 6, in autoactivate
  File "/home/mayor/Repos/globus-cli/.venv3/lib/python3.5/site-packages/globus_sdk-0.4.2-py3.5.egg/globus_sdk/transfer/client.py", line 286, in endpoint_autoactivate
    path = self.qjoin_path("endpoint", endpoint_id, "autoactivate")
  File "/home/mayor/Repos/globus-cli/.venv3/lib/python3.5/site-packages/globus_sdk-0.4.2-py3.5.egg/globus_sdk/base.py", line 95, in qjoin_path
    return "/" + "/".join(quote(part) for part in parts)
  File "/home/mayor/Repos/globus-cli/.venv3/lib/python3.5/site-packages/globus_sdk-0.4.2-py3.5.egg/globus_sdk/base.py", line 95, in <genexpr>
    return "/" + "/".join(quote(part) for part in parts)
  File "/usr/lib64/python3.5/urllib/parse.py", line 712, in quote
    return quote_from_bytes(string, safe)
  File "/usr/lib64/python3.5/urllib/parse.py", line 737, in quote_from_bytes
    raise TypeError("quote_from_bytes() expected bytes")
TypeError: quote_from_bytes() expected bytes
```

sirosen: Ugh. Of course. Should we fix this in the CLI (easy) or in the SDK (also easy, somewhat less direct)?
I suspect others will create `uuid.UUID`s and pass them in doing exactly this kind of validation, or even under the misapprehension that we require it.

corpulentcoffee: Fixing in the SDK might be better.

sirosen:
Marking as blocked until the SDK accepts `uuid.UUID`s.

globus_cli/services/transfer/async_delete.py
```diff
@@ -18,26 +18,24 @@
...
+                    'commandline, and includes any commandline PATH given'))
+@click.argument('endpoint_plus_path', metavar=ENDPOINT_PLUS_OPTPATH.metavar,
+                type=ENDPOINT_PLUS_OPTPATH)
+def async_delete_command(batch, ignore_missing, recursive, submission_id,
+                         endpoint_plus_path):
     """
     Executor for `globus transfer async-delete`
     """
+    endpoint_id, path = endpoint_plus_path
     if path is None and (not batch):
```
sirosen: Just to be sure that I understand you correctly, the worry is that even with the fix (which I think is a good idea), the doomsday input results in `path=""`, right?

It seems desirable that either `path is None` or it's a valid path, coming out of the `EnpointPlusPath` types.
Do you see any problem with having the type coerce empty strings into `None`?

corpulentcoffee: Also, as a sort of aside: what is the correct behavior when someone uses a path with batch mode? I thought we had talked about it prepending onto each line in the batch input (e.g. someone knows a base directory and wants to batch delete a bunch of junk inside of it), but the current state of affairs is to just add the path passed as an argument as another item to be deleted along with the items entered in the batch, which doesn't make any sense.

If users are actually NOT supposed to be able to call with a path in batch mode, maybe this should fail if both are set or both are unset (e.g. `if bool(path) is bool(batch): ... UsageError ...`).

corpulentcoffee: `EnpointPlusPath` coercing empty strings into `None` would be dandy too.

sirosen: Ah, I had incorrectly remembered the behavior you described (and which, at the time, we agreed upon) for batchmode with paths.

I think doing that correctly will be somewhat delicate, to respect absolute paths and produce normalized paths. Can we leave the behavior I put in for now and fix it before doing a release?
Or make it more strict now and change it later?
I think there are all kinds of interesting questions we'll be asking ourselves, like what to do when someone does a batch delete with `path="../../"`.

corpulentcoffee: Yeah, I can make a new issue for it,

sirosen:
With the SDK work merged, are we okay moving this forward?

corpulentcoffee:
Yes!... but it's a big changeset so I would have to finish it tomorrow morning / early afternoon.
(If someone else wants to review/vouch for everything and merge sooner, that's cool too.)

sirosen:
That's quite alright. I just found a bug in it, after all.

globus_cli/services/transfer/ls.py
```diff
@@ -76,7 +76,8 @@ def _get_ls_res(client, path, endpoint_id, recursive, depth, show_hidden):
 @click.command('ls', help='List the contents of a directory on an Endpoint',
                short_help='List Endpoint directory contents')
 @common_options
-@endpoint_id_option
+@click.argument('endpoint_plus_path', metavar=ENDPOINT_PLUS_OPTPATH.metavar,
+                type=ENDPOINT_PLUS_OPTPATH)
```
sirosen: I'm doing a rebase now, which will include that change.

globus_cli/services/transfer/share/create.py
```diff
@@ -9,15 +10,14 @@
 @click.command('create', help='Create a new Share, hosted on an Endpoint')
 @common_options
 @endpoint_create_and_update_opts(create=True, shared_ep=True)
-@click.option('--host-endpoint-id', required=True,
-              help='ID of the endpoint on which this Share is hosted')
-def share_create(host_endpoint_id, display_name, description, organization,
+@endpoint_id_arg(metavar='HOST_ENDPOINT_ID')
+def share_create(endpoint_id, display_name, description, organization,
                  contact_email, contact_info, info_link, public):
```
corpulentcoffee: We seem to be creating shares without letting the user specify the path on the host (which should be required).

corpulentcoffee: Also, this doesn't work anymore as a UUID instance (`TypeError: UUID('XXX') is not JSON serializable`). I don't know if you'd like to fix it somewhere here (e.g. your `assemble_generic_doc()` helper) or in the SDK.

sirosen: I noticed that issue with shares while doing this work, but decided to fix it later and just replicate the current behavior with the new parsing tools.
It will probably take an `ENDPOINT_PLUS_REQPATH`.

it seems pretty reasonable for `assemble_generic_doc()` -- whose sole purpose in this world is to help build Transfer document bodies -- would be able to apply the "safe stringify" to UUIDs at least.
In case of nested docs, it shouldn't try to do too much and risk accidentally storing a `str(subdict)` in the resulting JSON body.

PR #88, opened 2016-10-19
Add warning suppression via simplefilter, expose warnings with `--debug`

### LINKED ISSUE #17: Control python warnings with an env var
I think this might just be something that we hook into `GLOBUS_CLI_DEBUGMODE`.
We should suppress warnings during normal operations -- they're chatty and confusing for end users.
However, being able to see warnings is a useful feature for developers working on the project, so being able to expose them seems good.

sirosen:
This is disabled by `--debug`, but be forewarned: it means that warnings won't be enabled until we start a command's `invoke()`.
It also means that warnings sent during parsing could be suppressed even with `--debug` being passed.

Closes #17

### DIFF
--- a/globus_cli/parsing/command_state.py
+++ b/globus_cli/parsing/command_state.py
@@ -1,3 +1,4 @@
+import warnings
 import click
 
 from globus_cli import config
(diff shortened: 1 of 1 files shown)


jaswilli:
@sirosen Do you have an example command that's generating a warning so I can test this quickly?

sirosen:
I do not. To test, I hacked
```python
import warnings
warnings.warn('BAD!', Warning)
```

into a few places and made sure it showed up as expected.

PR #90, opened 2016-10-19
Save access tokens to config and use if valid

### LINKED ISSUE #83: Store and use access tokens in the configuration file
Currently every command invocation fetches a new access token from the stored refresh token. I think this is going to quickly get out of hand if and when people start making a lot of CLI calls.

Storing access tokens (and their expiration time) in the configuration file will allow us to make calls directly instead of needing to always go and get a token. Doing this should have the added benefit of making each CLI command more responsive.

I'm thinking we add the following new keys:

`<service>_access_token`
`<service>_access_token_expires_at`

jaswilli:
Closes #83

* Start saving access_tokens (and their expiration time) in the config file.
* Pass access_tokens into the authorizer to avoid always fetching a new access_token on every CLI command.
* Uses the refresh token authorizer's new `on_refresh` callback to get notified when access_token data should be updated in the config file.

Depends on https://github.com/globus/globus-sdk-python/pull/72

### DIFF
--- a/globus_cli/config.py
+++ b/globus_cli/config.py
@@ -11,13 +11,17 @@ __all__ = [
     # option name constants
     'OUTPUT_FORMAT_OPTNAME',
     'AUTH_RT_OPTNAME',
+    'AUTH_AT_OPTNAME',
+    'AUTH_AT_EXPIRES_OPTNAME',
     'TRANSFER_RT_OPTNAME',
+    'TRANSFER_AT_OPTNAME',
+    'AUTH_AT_EXPIRES_OPTNAME',
 
     'internal_auth_client',
 
     'get_output_format',
-    'get_auth_refresh_token',
-    'get_transfer_refresh_token',
+    'get_auth_tokens',
+    'get_transfer_tokens',
 
     'write_option',
     'lookup_option',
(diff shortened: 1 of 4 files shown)


jaswilli:
(this obviously needs to be held for a new release of the SDK)

globus_cli/config.py
```diff
@@ -94,6 +110,30 @@ def get_transfer_refresh_token():
     return lookup_option(TRANSFER_RT_OPTNAME)
 
 
+def get_auth_access_token():
```
sirosen: Is there a particular reason that this is still separate from loading the refresh token? (Likewise for `get_transfer_access_token()`)
It feels like a development decision that's carrying over.

Assuming there isn't a strong reason, could you fold these together into something like `get_auth_credentials`?
It's fewer names to export/import/keep-track-of.

jaswilli: That's a good point. I suppose there's no reason to have separate methods for getting the tokens from the config...

I'll consolidate them into a single method.

jaswilli:
Updated with consolidated "get tokens" methods--much better.

globus_cli/config.py
```diff
@@ -86,12 +100,30 @@ def get_output_format():
     return lookup_option(OUTPUT_FORMAT_OPTNAME)
 
 
-def get_auth_refresh_token():
-    return lookup_option(AUTH_RT_OPTNAME)
+def get_auth_tokens():
+    return {
+        'refresh_token': lookup_option(AUTH_RT_OPTNAME),
+        'access_token': lookup_option(AUTH_AT_OPTNAME),
+        'access_token_expires': int(lookup_option(AUTH_AT_EXPIRES_OPTNAME))
```
sirosen: Because `lookup_option` can return `None`, this check fails if you kill your own config file or remove the `access_token_expires` field.
Applies to Transfer as well, of course.

To confirm a fix, move your config file and try `globus auth get-identities 'abc@globusid.org' --debug`

I've run into similar issues before -- should `lookup_option` be changed to take a callable named `type`?

jaswilli: I meant using the helpers inside `lookup_option`, but yeah, since it's just two spots where we need it I'll handle it separately for now and we can always revisit if it pops up again.

jaswilli: Updated

sirosen:
@jaswilli Was your plan to hold this until we actually do an SDK release, or are you okay with merging it, and we'll make sure to do the SDK before a new CLI version? I think we're already in that situation, and wouldn't be comfortable trying to do a pure-CLI update at this point.

jaswilli:
No, I'm totally fine with merging now.

sirosen:
I'm happy with where this went. A merge you have suggested, and a merge you shall get.

PR #92, opened 2016-10-19
Expand the set of minimal tests

sirosen:
Still very much a poor man's testsuite.

Just adding things that should hopefully get us through the next rainy season.


sirosen:
Also, three cheers for `--map-http-status "*=0"`. Solving your testing problems since 2016. :laughing:

PR #93, opened 2016-10-20
Add option to set label on delete tasks

jaswilli:
* Adds `--label` option.


corpulentcoffee:
I'm pro going to labels that default to `null` (like the webapp) so that clients can come up with some sensible description of label-less tasks. Is that okay, @sirosen ?

sirosen:
I'm okay with that as well.
The thought behind the prior default label was that it might be desirable to be able to look directly at the label to differentiate CLI vs. Web submission. Since that use case is purely hypothetical, imitating the choice of the web and doing a default of `null` seems reasonable.

Issue #68, opened 2016-10-17
Transfer: missing `async-transfer` options


jaswilli:
* `preserve-mtime`
* `perf-pp`, `perf-cc`, `perf-p`, `perf-udt`
* `verify-checksum` (default: yes)
* `label` (set)

**Update**
Everything but the `perf-*` options implemented in #94.

jaswilli:
Per 10/27 status meeting we're not going to re-implement these.

PR #94, opened 2016-10-21
Add options for transfer task settings

jaswilli:
* Add `--label` to `async-transfer`
* Add `--verify-checksum` and `--no-verify-checksum` to `async-transfer` (default is to verify checksum)
* Add `--preserve-mtime` to `async-transfer`
* Add `size` to sync level documentation
* Add `--encrypt` option to `async-transfer`

Refs #68


globus_cli/services/transfer/async_transfer.py
```diff
@@ -77,6 +77,8 @@
     EXISTS: Determine whether or not to transfer based on file existence.
     If the destination file is absent, do the transfer.
 
+    SIZE: Determine whether or not to transfer based on the size of the file.
+    If destination file size does not match the source, do the transfer.
```
sirosen: Nitpick: should be followed by a blank line for spacing.

globus_cli/services/transfer/async_transfer.py
```diff
@@ -100,18 +102,27 @@
...
               type=CaseInsensitiveChoice(
                   ("exists", "size", "mtime", "checksum")),
               help=('How will the Transfer Task determine whether or not to '
                     'actually transfer a file over the network?'))
+@click.option('--preserve-mtime', is_flag=True, default=False,
+              help=('Preserve file and directory modification times.'))
+@click.option('--verify-checksum/--no-verify-checksum', default=True,
+              show_default=True,
+              help=('Verify checksum after transfer.'))
+@click.option('--encrypt', is_flag=True, default=False,
```
sirosen: This is the only default that I'm not sure makes sense.
I assume it's the same as the default for the web app, but I'd sooner have a `--no-encrypt` to turn it off than `--encrypt` to turn it on. Non-blocker for merging, however.

PR #99, opened 2016-10-22
Add tab completion as a hidden, "experimental" feature

### LINKED ISSUE #34: Add bash completion (maybe other forms of completion as well)
The awscli has fairly comprehensive completion support, which may serve as a guide:
http://docs.aws.amazon.com/cli/latest/userguide/cli-command-completion.html

It should be possible, given the structure of click commands, to do a completer as a python function or a hidden option to the CLI.

If we want completers to be pure shell functions, we can add a hidden option to generate a shell script as output, and then ship that result as part of the package data.

sirosen:
Resolves #34
I'm well aware that this isn't a priority, but it was fun to work on.

Supports both Bash and Zsh, implemented as a hidden option to the top level `globus` command. The result is that as long as `globus` can be invoked in your environment, so can `globus` tab completion. This strategy is in place of the `_GLOBUS_COMPLETE=...` approach, which doesn't appear to offer the same range of flexibility with the same level of ease.

Also includes a couple of scripts that can be used for completion support in these shells.
Sourcing them will have the desired effect.
We might want to ship these in package data in the future -- it's unclear to me at this time.

Once Bash completion (which is probably the only one that "really matters") was done, it was more or less trivial to also add zsh, which is my personal shell of choice. This is also good for keeping things well modularized to maintain the option of supporting other shells if their completion formats are simple as well.

### DIFF
--- a/globus_cli/parsing/shell_completion.py
+++ b/globus_cli/parsing/shell_completion.py
@@ -1,0 +1,195 @@
+import os
+import shlex
+import click
+
+from globus_cli.safeio import safeprint
+from globus_cli.parsing.hidden_option import HiddenOption
+from globus_cli.parsing.case_insensitive_choice import CaseInsensitiveChoice
+
+SUPPORTED_SHELLS = ['BASH', 'ZSH']
+
+
+def safe_split_line(inputline):
+    parser = shlex.shlex(inputline, posix=True)
+    parser.whitespace_split = True
+    res = []
+    # track whether or not we're looking at quoted text -- it should suppress a
+    # lot of types of completion if we see an open quote without a close quote
+    quoted = False
+    try:
+        for val in parser:
+            res.append(val)
+    # No closing quotation
+    except ValueError:
+        quoted = True
+    # grab the last token from the shlexer
+    if parser.token:
+        res.append(parser.token)
+    return res, quoted
+
+
+def get_completion_context(args):
+    """
+    Walk the tree of commands to a terminal command or multicommand, using the
+    Click Context system.
+    Effectively, we'll be using the resilient_parsing mode of commands to stop
+    evaluation, then having them capture their options and arguments, passing
+    us on to the next subcommand. If we walk "off the tree" with a command that
+    we don't recognize, we have a hardstop condition, but otherwise, we walk as
+    far as we can go and that's the location from which we should do our
+    completion work.
+    """
+    # get the "globus" command as a click.Command
+    root_command = click.get_current_context().find_root().command
+    # build a new context object off of it, with resilient_parsing set so that
+    # no callbacks are invoked
+    ctx = root_command.make_context('globus', list(args),
+                                    resilient_parsing=True)
+
+    # walk down multicommands until we've matched on everything and are at a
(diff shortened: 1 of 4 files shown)


sirosen:
I started today on an interesting experiment in handling bookmark names as complete-able things in the `bookmark locate` command. It very quickly became obvious that the extra network callouts were kiling performance to the point of being unworkable, so I threw together a quick caching layer in my homedir.

This isn't polished to the point of being merge-able, but it's an interesting toy to play with: [bookmark names cache experiment](https://github.com/sirosen/globus-cli/compare/feature/tab-completion...sirosen:experiment/tab-completion/cached-arg-completions).
Notably, it right now can't tell when all of the positional arguments have already been given, so completion "keeps going" allowing you to keep adding bookmark names after the first one.

Being able to tab-complete bookmark names has pretty significant implications for interactive use of `bookmark locate`, and therefore all endpoint operations. Namely: with this in place, bookmarks start to feel usable and useful in the CLI.

globus_cli/parsing/shell_completion.py
```diff
@@ -0,0 +1,195 @@
...
+                continue
+            for opt in param.opts:
+                # only add long-opts, never short opts to completion, unless
+                # the cur appears to be a short opt already
+                if opt.startswith('--') or (
+                        len(cur) > 1 and cur[1] != '-'):
+                    choices.append(opt)
+    # and if it's a multicommand we see, get the list of subcommands
+    elif isinstance(ctx.command, click.MultiCommand) and not quoted:
+        choices = [name for name in ctx.command.list_commands(ctx)]
```
corpulentcoffee: This list comprehension doesn't seem to be necessary.
Maybe you used to have a predicate here or maybe you just want a shallow copy (`list()` or `[:]`?).

globus_cli/parsing/shell_completion.py
```diff
@@ -0,0 +1,195 @@
...
+
+from globus_cli.safeio import safeprint
+from globus_cli.parsing.hidden_option import HiddenOption
+from globus_cli.parsing.case_insensitive_choice import CaseInsensitiveChoice
+
+SUPPORTED_SHELLS = ['BASH', 'ZSH']
+
+
+def safe_split_line(inputline):
+    parser = shlex.shlex(inputline, posix=True)
```
corpulentcoffee: Out of curiosity, is there anything we can do with this to make subshells get parsed correctly? Right now, if I use a subshell in the middle of a line (e.g. for `globus bookmark locate`) it breaks completion for the rest of the line after I finish my subshell.

globus_cli/parsing/shell_completion.py
sirosen: I used to have a predicate of some kind, I think. Can remove.

globus_cli/parsing/shell_completion.py
corpulentcoffee: We agree that if users aren't lazy and quote their subshells (which they should!) that this works correctly, so non-issue.

sirosen:
Rebased and removed the unnecessary comprehension.

PR #100, opened 2016-10-23
Join batchmode paths with commandline paths

### LINKED ISSUE #89: Batch mode and paths?
Either

- forbid it by changing the guards for async-delete and async-transfer to accept only batch mode exclusive-or path
- make it so the path specified on the command line is prepended to each line of the batch input (which might have corner cases like path normalization issues)

Today, the path from the command line is thrown in with the paths input on the batch input, which is at best not helpful or at worst will lead to unexpected behavior.

sirosen:
Per #89, slapping the commandline paths into the document body as just another pair is potentially confusing and not very good semantically.
To be more useful, join paths as given on the commandline to all of the batch inputs, letting the commandline paths be used to specify a prefix.

Implemented with a new custom ParamType for consuming/converting paths relative to a base dir.
That type has some unused features from my experiments with it which I'm leaving in, in case we want to use them in the future.

This leads logically towards an interesting potential use case for `globus transfer ls | globus transfer async-delete ...`
That's also interesting for `async-transfer`, but requires the annoying extra step of `| awk '{print $1 " " $1}'`. Thinking further about that, why not let `dest_path` be optional during transfer submission, and we'll just copy the source path, potentially joined with a commandline dest path?
This changeset includes that modification, as it actually has some minor implications for the ParamType definition.

With this improvement, it's possible to pipe and `ls` through a grep and ultimately to transfer submission. Things only get messy when you have a pipeline producing a mix of files and dirs for transfer -- in that case, a sophisticated manipulation with `awk` or similar is once again required.
Consider, however
```shell
$ globus transfer ls "$ENDPOINT_ID:/mypath" | grep '.*\.txt' | \
    globus transfer async-transfer "$ENDPOINT_ID:/mypath" "$OTHER_ENDPOINT:mypath_txts/"
```
in simple cases like this, accepting single-column input for batchmode proves very useful. (Yes, that example can be messed up by spaces in the path, but this isn't the place for building quoting tools).

Resolves #89

### DIFF
--- a/globus_cli/services/transfer/async_transfer.py
+++ b/globus_cli/services/transfer/async_transfer.py
@@ -5,7 +5,7 @@ from globus_sdk import TransferData
 
 from globus_cli.parsing import (
     CaseInsensitiveChoice, common_options, submission_id_option,
-    ENDPOINT_PLUS_OPTPATH)
+    TaskPath, ENDPOINT_PLUS_OPTPATH)
 from globus_cli.safeio import safeprint
 from globus_cli.helpers import (
     outformat_is_json, print_json_response, colon_formatted_print)
@@ -46,7 +46,7 @@ from globus_cli.services.transfer.activation import autoactivate
 
     \b
     Lines are of the form
-    [--recursive] SRCPATH DSTPATH
+    [--recursive] SOURCE_PATH [DEST_PATH]
 
     Skips empty lines and allows comments beginning with "#".
 
@@ -55,18 +55,29 @@ from globus_cli.services.transfer.activation import autoactivate
         $ cat dat
         # file 1, simple
         ~/dir1/sourcepath1 /abspath/destpath1
+
         # file 2, with spaces in dest path
-        # paths without explicit ~ implicitly use ~
+        # paths without / or ~ are relative to the default_directory of the
+        # endpoint they are on
         dir2/sourcepath2   "path with spaces/dest2"
+
         # dir 1, requires --recursive option
         --recursive ~/srcdir1/ /somepath/destdir1
+
+        # if dest path is omitted, the source path is copied
+        # the lines below are therefore equivalent to
+        #   --recursive ~/srcdir1/ ~/srcdir1/
+        #   file_to_copy file_to_copy
+        --recursive ~/srcdir1/
+        file_to_copy
     \b
         $ cat dat | globus transfer async-transfer \\
             --batch --sync-level checksum \\
-            --source-endpoint "..." --dest-endpoint "..."
+            "$SOURCE_ENDPOINT_ID" "$DEST_ENDPOINT_ID"
 
-    You can use `--batch` in addition to a single SOURCE_PATH, DEST_PATH pair
-    given in the commandline.
+    You can use `--batch` with a commandline SOURCE_PATH and/or DEST_PATH.
+    In that case, these paths will be used as dir prefixes to any paths in the
+    batch input.
 
     \b
     Sync Levels
@@ -100,7 +111,7 @@ from globus_cli.services.transfer.activation import autoactivate
               help=("Don't actually perform the transfer, print submission "
                     "data instead"))
 @click.option('--recursive', is_flag=True,
-              help=('source-path and dest-path are both directories, do a '
+              help=('SOURCE_PATH and DEST_PATH are both directories, do a '
                     'recursive dir transfer. Ignored in batchmode'))
 @click.option('--label', default=None, help=('Set a label for this task.'))
 @click.option('--sync-level', default=None, show_default=True,
@@ -119,7 +130,8 @@ from globus_cli.services.transfer.activation import autoactivate
               help=('Accept a batch of source/dest path pairs on stdin (i.e. '
                     'run in batchmode). '
                     'Uses SOURCE_ENDPOINT_ID and DEST_ENDPOINT_ID as passed '
-                    'on the commandline.'))
+                    'on the commandline. Commandline paths are still allowed '
+                    'and are used as prefixes to the batchmode inputs.'))
 @click.argument('source', metavar='SOURCE_ENDPOINT_ID[:SOURCE_PATH]',
                 type=ENDPOINT_PLUS_OPTPATH)
 @click.argument('destination', metavar='DEST_ENDPOINT_ID[:DEST_PATH]',
@@ -130,10 +142,10 @@ def async_transfer_command(batch, sync_level, recursive, destination, source,
     """
     Executor for `globus transfer async-transfer`
     """
-    source_endpoint, source_path = source
-    dest_endpoint, dest_path = destination
+    source_endpoint, cmd_source_path = source
+    dest_endpoint, cmd_dest_path = destination
 
-    if (source_path is None or dest_path is None) and (not batch):
+    if (cmd_source_path is None or cmd_dest_path is None) and (not batch):
         raise click.UsageError(
             ('async-transfer requires either SOURCE_PATH and DEST_PATH or '
              '--batch'))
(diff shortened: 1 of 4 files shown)


corpulentcoffee:
I think the behavior when you specify an absolute source path in the batch input without an explicit destination path is unexpected.

e.g.:

```
(.venv3) ~/Repos/globus-cli$ globus transfer async-transfer \
    ddb59aef-6d04-11e5-ba46-22000b92c6ec:/some-directory-on-ep1 \
    ddb59af0-6d04-11e5-ba46-22000b92c6ec:/some-directory-on-ep2 \
    --batch --dry-run
Enter transfers, line by line, as

    [--recursive] SOURCE_PATH [DEST_PATH]

Lines are split with shlex in POSIX mode: https://docs.python.org/library/shlex.html#parsing-rules
Terminate input with Ctrl+D or <EOF>

relative-123.txt
/absolute-456.txt
^D
{
  "DATA_TYPE": "transfer",
  "preserve_timestamp": false,
  "verify_checksum": true,
  "submission_id": "69bfe6d9-99fe-11e6-b0c1-22000b92c261",
  "destination_endpoint": "ddb59af0-6d04-11e5-ba46-22000b92c6ec",
  "source_endpoint": "ddb59aef-6d04-11e5-ba46-22000b92c6ec",
  "encrypt_data": false,
  "DATA": [
    {
      "destination_path": "/some-directory-on-ep2/relative-123.txt",
      "recursive": false,
      "source_path": "/some-directory-on-ep1/relative-123.txt",
      "DATA_TYPE": "transfer_item"
    },
    {
      "destination_path": "/absolute-456.txt",
      "recursive": false,
      "source_path": "/absolute-456.txt",
      "DATA_TYPE": "transfer_item"
    }
  ]
}
```

The first pair does what I would expect (`/some-directory-on-ep1/relative-123.txt` to `/some-directory-on-ep2/relative-123.txt`), but the pair from the absolute path doesn't (`/absolute-456.txt` to `/absolute-456.txt`). Given that I am omitting the batch destination path but I gave a command destination path, I would sort of expect `/absolute-456.txt` to `/some-directory-on-ep2/absolute-456.txt`.

I *think* that the behavior I would *expect* would be to take the last component off of the source absolute path (e.g. `absolute-456.txt` from `/absolute-456.txt`, `def/` from `/abc/def/`, `home/` from `/~/`) and then add it to the "command" destination path (e.g. `/some-directory-on-ep2/absolute-456.txt`, `/some-directory-on-ep2/def/`, `/some-directory-on-ep2/home/`).

sirosen:
I'm not sure I agree. Since the help says that writing
```
relative-123.txt
/absolute-456.txt
~/absolute-789.txt
```
is the same as writing
```
relative-123.txt relative-123.txt
/absolute-456.txt /absolute-456.txt
~/absolute-789.txt ~/absolute-789.txt
```
then I would *not* expect the absolute paths to be altered.

Maybe that's not clear enough in the help? Or maybe that behavior in general is wrong and the help should be amended to reflect what you're describing, and that should be the behavior?

corpulentcoffee:
Sorry, the behavior is consistent with the help text (i.e. the "if dest path is omitted, the source path is copied" bit), but it's not the behavior I would expect intuitively if I was just using it without a careful read of the help text.

sirosen:
To clarify, I have no opinion about which behavior is more correct. I just meant to highlight that this doesn't strike me as "unexpected". Unintuitive is still a fair and valid criticism.
Each way of doing it has pros and cons, and is incapable of expressing certain behaviors that explicitly specifying both paths achieves.

Do you want me to change this to join with a given dest path? What should it do with ~? If no dest is given command-line, should these paths be made relative to the default dir?
As I ask these questions, I start to worry that this might open the door to other ambiguities, but if there's a clear "right" behavior it's not an issue. We can doc it all anyway.

corpulentcoffee:
If we think that there are other ambiguities that might exist, maybe make the destination path required again, but, I think perhaps maybe we can do it definitively and document it:

1. **What should it do with ~?** `/~/` as the source path would need to be special-cased to `/destination-path/home/` or something similar.
2. **If no dest is given command-line, should these paths be made relative to the default dir?** Yes, that would be what I would expect it to do. Unless they have been configured to do otherwise, shares would receive the files in their virtual roots and other servers would be receive the files in the user's home directory.

globus_cli/services/transfer/async_transfer.py
```diff
@@ -55,18 +55,29 @@
...
+
         # dir 1, requires --recursive option
         --recursive ~/srcdir1/ /somepath/destdir1
+
+        # if dest path is omitted, the source path is copied
+        # the lines below are therefore equivalent to
+        #   --recursive ~/srcdir1/ ~/srcdir1/
+        #   file_to_copy file_to_copy
+        --recursive ~/srcdir1/
+        file_to_copy
```
corpulentcoffee: Click is rewrapping these code block "paragraphs" and it looks a mess on the `--help` output. Extra `\b` escapes at every indentation "paragraph" seem to help.

sirosen:
Dave and I discussed, and agreed to (for now) drop the single-column input to async-transfer.
There are a bunch of questions, and we don't have the bandwidth to seek answers right now.

sirosen:
I tried adding an `async-transfer --dry-run` test, but I had to back it out because it, of course, fails to get a submission ID when building the TransferData.

corpulentcoffee:
> I tried adding an `async-transfer --dry-run` test, but I had to back it out because it, of course, fails to get a submission ID when building the TransferData.

Can you manually pass a `--submission-id` parameter so it doesn't try to fetch one?

sirosen:
Past laziness bites again. `--submission-id` doesn't actually suppress the fetch of a submission ID -- it just replaces the one we fetch.
That's obviously somewhat stupid, but it was expedient at the time.

globus_cli/services/transfer/async_transfer.py
```diff
@@ -144,26 +153,26 @@ def async_transfer_command(batch, sync_level, recursive, destination, source,
...
             """
-            transfer_data.add_item(source_path, dest_path,
+            transfer_data.add_item(str(source_path), str(dest_path),
                                    recursive=recursive)
 
         shlex_process_stdin(
             process_batch_line,
             ('Enter transfers, line by line, as\n\n'
-             '    [--recursive] source-path dest-path\n'))
+             '    [--recursive] SOURCE_PATH [DEST_PATH]\n'))
```
corpulentcoffee: Need to drop the brackets indicating optionalness here. (Maybe fixup 24de9fdfcd7b52e626da49af0ef9ce4bdc50e410?)

corpulentcoffee:
One thing I would suggest for `async-transfer`: either

- disallow using `--recursive` and `--batch` together, **or**
- if they *are* used together, hide "`[--recursive]`" in the "Enter transfers, line by line" output and apply the recursive flag to every line in the batch input.

Today, `--recursive` has no effect when using `globus transfer async-transfer --batch` ...

```
(.venv3) ~/Repos/globus-cli$ globus transfer async-transfer \
    --recursive --batch --dry-run \
    cf9bcaa5-6d04-11e5-ba46-22000b92c6ec \
    ddb59af0-6d04-11e5-ba46-22000b92c6ec
Enter transfers, line by line, as

    [--recursive] SOURCE_PATH [DEST_PATH]

Lines are split with shlex in POSIX mode: https://docs.python.org/library/shlex.html#parsing-rules
Terminate input with Ctrl+D or <EOF>

/a /b
/c /d
/e /f
^D
{
  "verify_checksum": true,
  "submission_id": "fefe1741-9ad2-11e6-b0c5-22000b92c261",
  "encrypt_data": false,
  "preserve_timestamp": false,
  "source_endpoint": "cf9bcaa5-6d04-11e5-ba46-22000b92c6ec",
  "DATA": [
    {
      "destination_path": "/b",
      "recursive": false,
      "DATA_TYPE": "transfer_item",
      "source_path": "/a"
    },
    {
      "destination_path": "/d",
      "recursive": false,
      "DATA_TYPE": "transfer_item",
      "source_path": "/c"
    },
    {
      "destination_path": "/f",
      "recursive": false,
      "DATA_TYPE": "transfer_item",
      "source_path": "/e"
    }
  ],
  "destination_endpoint": "ddb59af0-6d04-11e5-ba46-22000b92c6ec",
  "DATA_TYPE": "transfer"
}
```

... yet it totally works and is totally the way to do it with `globus transfer async-delete` ...

```
(.venv3) ~/Repos/globus-cli$ globus transfer async-delete \
    --recursive --batch --dry-run \
    cf9bcaa5-6d04-11e5-ba46-22000b92c6ec
Enter paths to delete, line by line.
Lines are split with shlex in POSIX mode: https://docs.python.org/library/shlex.html#parsing-rules
Terminate input with Ctrl+D or <EOF>

/a
/b
/c
^D
{
  "DATA": [
    {
      "path": "/a",
      "DATA_TYPE": "delete_item"
    },
    {
      "path": "/b",
      "DATA_TYPE": "delete_item"
    },
    {
      "path": "/c",
      "DATA_TYPE": "delete_item"
    }
  ],
  "ignore_missing": false,
  "DATA_TYPE": "delete",
  "endpoint": "cf9bcaa5-6d04-11e5-ba46-22000b92c6ec",
  "recursive": true,
  "submission_id": "3b578f79-9ad3-11e6-b0c5-22000b92c261"
}
```

... which might lead to user confusion unless we raise an error or make the behavior consistent.

sirosen:
Allowing `async-transfer --recursive --batch ...` seems more ambiguous and prone to misinterpretation.
That should have been fixed when the interpretation of paths included on the commandline changed, and was an oversight.

I'll rectify this now, along with the docstring fix.

PR #102, opened 2016-10-24
Support for `globus transfer task cancel --all`

### LINKED ISSUE #60: Transfer: add "--all" option to task cancel
* Existing CLI has an `--all` option that will cancel all `ACTIVE` and `INACTIVE` tasks.

corpulentcoffee:
Closes #60.

### DIFF
--- a/globus_cli/services/transfer/task/cancel.py
+++ b/globus_cli/services/transfer/task/cancel.py
@@ -44,8 +44,7 @@ def cancel_task(all, task_id):
         if outformat_is_json():
             safeprint(json.dumps(
                 {
-                    'results': [client.cancel_task(task_id).data
-                                for task_id in task_ids],
+                    'results': [client.cancel_task(i).data for i in task_ids],
                     'task_ids': task_ids,
                 },
                 indent=2,
--- a/globus_cli/services/transfer/task/cancel.py
+++ b/globus_cli/services/transfer/task/cancel.py
@@ -42,8 +42,14 @@ def cancel_task(all, task_id):
             raise click.ClickException('You have no in-progress tasks.')
 
         if outformat_is_json():
-            safeprint(json.dumps([client.cancel_task(task_id).data
-                                  for task_id in task_ids], indent=2))
+            safeprint(json.dumps(
+                {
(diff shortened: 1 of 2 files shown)


globus_cli/services/transfer/task/cancel.py
```diff
@@ -10,16 +11,57 @@
...
+            task_row['task_id']
+            for task_row in client.task_list(
+                filter='type:TRANSFER,DELETE/status:ACTIVE,INACTIVE',
+                fields='task_id',
+                num_results=maxsize,  # FIXME want to ask for "unlimited" set
+            )
+        ]
+
+        if not task_ids:
+            raise click.ClickException('You have no in-progress tasks.')
```
sirosen: Is this the right behavior?
I think this should just be a successful exit.

It's vacuous, but the user is in the "good" state of having all active tasks canceled.

globus_cli/services/transfer/task/cancel.py
```diff
@@ -10,16 +11,57 @@
...
+
+        if outformat_is_json():
+            safeprint(json.dumps(
+                {
+                    'results': [client.cancel_task(i).data for i in task_ids],
+                    'task_ids': task_ids,
```
sirosen: I like this. It's kind of nice to get the IDs as a separate list. :+1:

globus_cli/services/transfer/task/cancel.py
```diff
@@ -10,16 +11,57 @@
...
+                },
+                indent=2,
+            ))
+
+        else:
+            task_count = len(task_ids)
+            safeprint('Canceling all tasks ({} total)...'.format(task_count))
+            for task_number, task_id in enumerate(task_ids, start=1):
+                safeprint('{} ({} of {}): {}'.
```
sirosen: This kind of streaming of success output is a generally applicable pattern. Do we want to take any pains to ensure that `safeprint` (perhaps optionally) does unbuffered writes?

sirosen:
We agreed to leave `--all` with no tasks as an error. `|| true` is easy.

Issue #95, opened 2016-10-21
Fix share creation to take an ENDPOINT_PLUS_REQPATH


sirosen:
Blocked on #86 

We noticed that share creation is taking an endpoint ID but not a required path.

PR #106, opened 2016-10-26
Convert most required options to parameters, reduce number of params in many places

### LINKED ISSUE #81: Required options should become positional arguments
e.g. `globus transfer bookmark show --bookmark-id=<UUID>`
should become `globus transfer bookmark show <UUID>`.

Requires #73 for commands that deal with endpoints and paths to simplify the interface of some commands (e.g. `globus transfer async-transfer`, which without any simplification would require four positional arguments instead of two).

sirosen:
This is meant to close #81 , but we may have ongoing work on this front.

Almost all required options are converted to arguments by this, but there are a couple of exceptions:

- Security principals are now, like `EndpointPlusPath`, a tuple (in this case type and value). They are built from `--identity <val>`, `--group <val>`, `--anonymous`, and `--all-authenticated`. They are not optional
- Permissions are still given as `--permissions`. This is because the meaning of a postional param with metavar `PERMISSIONS` is confusing, and there is no support for helptext on a positional. As a result the metavar `[R|RW]` is also confusing. The best solution, IMO, is to keep the helptext and make it a required option

These changes also include some cleanup to the parameter decorators to factor out the logic for applying a decorator which takes function or a set of names arguments.


globus_cli/services/transfer/endpoint/share.py
```diff
@@ -1,18 +1,18 @@
...
 from globus_cli.helpers import print_json_response
 from globus_cli.services.transfer.helpers import (
     get_client, assemble_generic_doc)
 
 
-@click.command('create', help='Create a new Share, hosted on an Endpoint')
+@click.command('share', help='Create a new Share, hosted on an Endpoint')
 @common_options
-@endpoint_create_and_update_opts(create=True, shared_ep=True)
 @endpoint_id_arg(metavar='HOST_ENDPOINT_ID')
```
corpulentcoffee: Let's do #95 soon.

globus_cli/services/transfer/endpoint/share.py
```diff
@@ -1,18 +1,18 @@
...
-def share_create(endpoint_id, display_name, description, organization,
-                 contact_email, contact_info, info_link, public):
+@endpoint_create_and_update_params(create=True, shared_ep=True)
+def endpoint_create_share(endpoint_id, display_name, description, organization,
+                          contact_email, contact_info, info_link, public):
```
corpulentcoffee: This one doesn't work.

```
(.venv3) ~/Repos/globus-cli$ globus transfer endpoint share bedc9334-97bd-11e6-b0b1-22000b92c261 "My Nifty Share"
TypeError: endpoint_create_share() missing 1 required positional argument: 'public'
```

Share endpoints must have `public=True` (or omitted) anyway.

globus_cli/services/transfer/endpoint/role/create.py
```diff
@@ -14,28 +15,25 @@
...
               type=CaseInsensitiveChoice(
                   ('administrator', 'access_manager', 'activity_manager',
                    'activity_monitor')),
               help='A role to assign.')
-def role_create(role, principal, principal_type, endpoint_id):
+def role_create(role, principal, endpoint_id):
     """
     Executor for `globus transfer endpoint role show`
     """
+    principal_type, principal_val = principal
```
corpulentcoffee: This one is busted somehow.

```
(.venv3) ~/Repos/globus-cli$ globus transfer endpoint role create 9fe702c0-9baf-11e6-b0c6-22000b92c261 --identity some-access-manager@example.com --role access_manager --debug
Traceback (most recent call last):
  File "/home/mayor/Repos/globus-cli/globus_cli/parsing/main_command_decorator.py", line 25, in invoke
    return super(TopLevelGroup, self).invoke(ctx)
  File "/home/mayor/Repos/globus-cli/.venv3/lib/python3.5/site-packages/click-6.6-py3.5.egg/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/mayor/Repos/globus-cli/.venv3/lib/python3.5/site-packages/click-6.6-py3.5.egg/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/mayor/Repos/globus-cli/.venv3/lib/python3.5/site-packages/click-6.6-py3.5.egg/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/mayor/Repos/globus-cli/.venv3/lib/python3.5/site-packages/click-6.6-py3.5.egg/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/mayor/Repos/globus-cli/.venv3/lib/python3.5/site-packages/click-6.6-py3.5.egg/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/mayor/Repos/globus-cli/.venv3/lib/python3.5/site-packages/click-6.6-py3.5.egg/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "/home/mayor/Repos/globus-cli/globus_cli/services/transfer/endpoint/role/create.py", line 28, in role_create
    principal_type, principal_val = principal
ValueError: too many values to unpack (expected 2)
```

globus_cli/services/transfer/endpoint/acl/add_rule.py
```diff
@@ -12,31 +13,30 @@
...
+def add_acl_rule(principal, permissions, endpoint_plus_path):
     """
     Executor for `globus transfer acl add-rule`
     """
+    if not principal:
+        raise click.UsageError(
+            'A security principal is required for this command')
+
     endpoint_id, path = endpoint_plus_path
+    principal_type, principal_val = principal
```
corpulentcoffee:
```
(.venv3) ~/Repos/globus-cli$ globus transfer endpoint acl add-rule 9fe702c0-9baf-11e6-b0c6-22000b92c261:/somewhere --identity=example@example.com --permissions=r --debug
Traceback (most recent call last):
  File "/home/mayor/Repos/globus-cli/globus_cli/parsing/main_command_decorator.py", line 25, in invoke
    return super(TopLevelGroup, self).invoke(ctx)
  File "/home/mayor/Repos/globus-cli/.venv3/lib/python3.5/site-packages/click-6.6-py3.5.egg/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/mayor/Repos/globus-cli/.venv3/lib/python3.5/site-packages/click-6.6-py3.5.egg/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/mayor/Repos/globus-cli/.venv3/lib/python3.5/site-packages/click-6.6-py3.5.egg/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/mayor/Repos/globus-cli/.venv3/lib/python3.5/site-packages/click-6.6-py3.5.egg/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/mayor/Repos/globus-cli/.venv3/lib/python3.5/site-packages/click-6.6-py3.5.egg/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/mayor/Repos/globus-cli/.venv3/lib/python3.5/site-packages/click-6.6-py3.5.egg/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "/home/mayor/Repos/globus-cli/globus_cli/services/transfer/endpoint/acl/add_rule.py", line 31, in add_acl_rule
    principal_type, principal_val = principal
ValueError: too many values to unpack (expected 2)
```

sirosen:
I tracked down and resolved these issues. There's a verbose commit message with details on the consternation caused by the callback processing behaviors in `click`.
In fairness to the library, the use-case of `--identity` and `--group` both wanting to write the `principal` parameter value is a little bit unusual. Part of this is just my trying to enforce that argument parsing issues should be separate from the command functionality -- there are simpler but "dirtier" solutions to the problem.

We already see this in a number of places due to our inability to enforce mutually exclusive parameter groups. If we want to replace this with something simpler, it might be best to pursue mutually exclusive params, then circle back and implement this as a collection of mutually exclusive options which can be ORed together.

Issue #103, opened 2016-10-25
Parsing errors when using --batch/shlex_process_stdin() don't abort the ultimate processing of the command


corpulentcoffee:
I ran into this while testing #100, but this bug happens on master too and wasn't introduced by that PR, so I'm opening a new issue for it.

```
(.venv3) ~/Repos/globus-cli$ globus transfer async-transfer ddb59aef-6d04-11e5-ba46-22000b92c6ec ddb59af0-6d04-11e5-ba46-22000b92c6ec --batch
Enter transfers, line by line, as

    [--recursive] source-path dest-path

Lines are split with shlex in POSIX mode: https://docs.python.org/library/shlex.html#parsing-rules
Terminate input with Ctrl+D or <EOF>

/a /b
/c /d
/e
/f /g
/h /i
/j
/k /l
^D
Usage: globus [OPTIONS] SOURCE_PATH DEST_PATH

Error: Missing argument "dest_path".
Usage: globus [OPTIONS] SOURCE_PATH DEST_PATH

Error: Missing argument "dest_path".
Message: The transfer has been accepted and a task has been created and queued for execution
Task ID: e0a58ab2-9acd-11e6-b0c5-22000b92c261
```

```
(.venv3) ~/Repos/globus-cli$ globus transfer async-transfer ddb59aef-6d04-11e5-ba46-22000b92c6ec ddb59af0-6d04-11e5-ba46-22000b92c6ec --batch --dry-run
Enter transfers, line by line, as

    [--recursive] source-path dest-path

Lines are split with shlex in POSIX mode: https://docs.python.org/library/shlex.html#parsing-rules
Terminate input with Ctrl+D or <EOF>

/a /b
/c /d
/e
/f /g
/h /i 
/j
/k /l
^D
Usage: globus [OPTIONS] SOURCE_PATH DEST_PATH

Error: Missing argument "dest_path".
Usage: globus [OPTIONS] SOURCE_PATH DEST_PATH

Error: Missing argument "dest_path".
{
  "DATA": [
    {
      "recursive": false,
      "source_path": "/a",
      "destination_path": "/b",
      "DATA_TYPE": "transfer_item"
    },
    {
      "recursive": false,
      "source_path": "/c",
      "destination_path": "/d",
      "DATA_TYPE": "transfer_item"
    },
    {
      "recursive": false,
      "source_path": "/f",
      "destination_path": "/g",
      "DATA_TYPE": "transfer_item"
    },
    {
      "recursive": false,
[...]

Issue #42, opened 2016-10-13
Remove `globus transfer endpoint autoactivate`


sirosen:
Dan mentioned this to me.
Specifically, trying to autoactivate an activated endpoint which does not support autoactivate is confusing:

``` shell
$ globus transfer endpoint autoactivate --endpoint-id danielpowers#prod01
{
  "code": "AutoActivationFailed",
  "resource": "/endpoint/danielpowers%23prod01/autoactivate",
  ...
  "message": "The endpoint could not be auto activated, fill in the returned activation_requirements and POST them back to /activate to perform manual activation.",
  ...
  "oauth_server": null,
  "subject": null
}
```

Really, the autoactivate call in this case is unhelpful / unimportant, as the activation requirements document only applies if the `expires_in` field you get back is unacceptable.
autoactivate calls on autoactivate-capable endpoints works fine and has acceptable (but noisy) output.

Additionally, this call is generally unnecessary because we automatically attempt to autoactivate any endpoint which we see, so trying an `ls` or similar operation will result in an implicit and less scary looking autoactivation.

If we keep autoactivate as an explicit operation, we need to figure out better (textmode) output for it.
If we make it only magically autoactivate, this will feed into #30 as well.

jaswilli:
I can't think of a use for `autoactivate` in the cli given that it's always going to do the autoactivate attempt behind the scenes.

Even having it output the activation requirements doesn't seem terribly useful since the cli is going to direct everyone to the web app to do endpoint activation.

I think maybe this issue should turn into "remove `transfer endpoint autoactivate` command?"

sirosen:
The only counterpoint is that it might be desirable to autoactivate something for use with the SDK.
Not that that's a major use case, but I think that's why we added it.

Let's go ahead and remove it -- it's easy enough to restore later if desired.

sirosen:
I got feedback from Dan that he's using `autoactivate` to check whether or not an endpoint requires a web activation flow.

In scripted use-cases, it's good to be able to check whether or not an endpoint will need activation within the expected timeframe of the script, and abort safely and early.
We need to make sure we continue to support that use case -- does that mean we have a course-adjustment on removing this, or that we do something else to handle that?

sirosen:
We recently agreed on `is-activated` with a `--until TIME` option.
This rename will be included with the major name changing refactor.

PR #119, opened 2016-10-29
Restructure command tree / refactor repo layout

### LINKED ISSUE #71: Command heirarchy/grouping

sirosen:
The layout of modules, since they don't have to reflect the CLI surface / usage, is largely a matter of taste. However, making them reflect that structure was the only sensible option I could work out.

So, in brief:
- all commands are now in `globus_cli.commands`
- helpers for specific services are still in `globus_cli.services`
- the commands tree itself has been restructured, per #71 
- rename `endpoint acl` to `endpoint permission`, per discussions
- rename `endpoint autoactivate` to `endpoint is-activated` with a new implementation to match those semantics

The one bit of weirdness is that I haven't come up with a better name for all of the `commands` modules for building chunks of the tree, so we end up with ugliness like `globus_cli.commands.bookmark.commands`, which is not great naming-wise.

Closes #71, #103, #42


globus_cli/commands/endpoint/is_activated.py
```diff
@@ -0,0 +1,68 @@
...
+                   .format(endpoint_id, exp_string), res)
+        click.get_current_context().exit(1)
+
+    def success(msg, *format_params):
+        _print_out(msg.format(endpoint_id, *format_params), res)
+        click.get_current_context().exit(0)
+
+    # autoactivation always means success, so short-circuit out of here
+    if res.supports_auto_activation:
+        success('{} does not require activation')
```
corpulentcoffee: I don't think that this part is correct.

Using [an endpoint that definitely requires activation](https://www.globus.org/app/endpoints/44e6ddac-e818-11e5-97d5-22000b9da45e), `is-activated` tells me it doesn't require activation:

```
~$ globus endpoint show 44e6ddac-e818-11e5-97d5-22000b9da45e
Display Name: XSEDE Gordon
ID:           44e6ddac-e818-11e5-97d5-22000b9da45e
Owner:        xsede@globusid.org
Activated:    False
Shareable:    True

~$ globus endpoint is-activated 44e6ddac-e818-11e5-97d5-22000b9da45e
44e6ddac-e818-11e5-97d5-22000b9da45e does not require activation
```

For what it's worth: the webapp decides whether or not to display the "Activate" tab in Manage Endpoints based on whether the endpoint document has an `expires_in` of `-1` on the endpoint document. This same value also appears to be available in the activation requirements document.

sirosen: Why is `"auto_activation_supported": True` on this endpoint? Do we know?

If an endpoint like `go#ep1` is used, I don't want to be telling people, incorrectly, that they need to go and activate it manually. `"expires_in": -1` will only be set if they've already activated it, which isn't right either.

Does supporting autoactivation mean something other than what I think it means?

jaswilli: > Why is "auto_activation_supported": True on this endpoint? Do we know?

A quick glance at transfer tells me that (almost?) all endpoint are going to come back with `auto_activation_supported: true`--the criteria is: shared endpoints, gcp endpoints, S3 endpoints, any endpoint with a myproxy server defined, any endpoint with an oauth server defined.

sirosen: Dave and I were just discussing this, actually. I'm going to switch this to look at `expires_in=-1`, per his suggestion. I also want a property on the SDK response that encodes that check in the future.

Apparently there is a wider class of endpoints which support autoactivate than I realized, but they don't guarantee that autoactivation will _succeed_, which seems to be the key difference.

jaswilli: Yes, I think "ok, I'll try and get a valid credential from Nexus" is a legitimate auto-activation operation, so that's why it always comes back as supported.

sirosen:
I changed the behavior per our discussion, and opened an SDK issue to add the property that I think should be available there.

globus_cli/commands/endpoint/is_activated.py
```diff
@@ -0,0 +1,68 @@
...
+                     'status 1, otherwise exits with status 0.'))
+@common_options
+@endpoint_id_arg
+@click.option('--until', type=int,
+              help=('An integer number of seconds in the future. If the '
+                    'endpoint is activated, but will expire by then, exits '
+                    'with status 1'))
+@click.option('--absolute-time', is_flag=True,
+              show_default=True, default=False,
+              help=('Treat the value of --until as a POSIX timestamp, not a '
```
corpulentcoffee: Feel free to ignore this, but it might be nice to phrase this as "POSIX timestamp (Unix time)" or "POSIX timestamp (Epoch time)" so that people know what you're talking about and don't get it confused with ISO 8601.

PR #120, opened 2016-10-31
Add initial version of logout command

### LINKED ISSUE #38: Add `globus logout` command
The inverse of `globus login`.
This can't touch the consent in Globus Auth, but it should invalidate all of your tokens by revoking them, and scrub them from the config file.

sirosen:
Revokes all tokens, and deletes them from config. Further attempts to
use the tokens naturally fail.
As an unfortunate aside, we can't do much to direct users into the
consents page right now. Maybe print some epilog help about how to go
there and remove the consent? It's a dead entity, but we can't clean it
up at all.

Resolves #38

### DIFF
--- a/globus_cli/commands/logout.py
+++ b/globus_cli/commands/logout.py
@@ -1,0 +1,43 @@
+import click
+
+from globus_cli.safeio import safeprint
+from globus_cli.parsing import common_options
+from globus_cli.config import (
+    AUTH_RT_OPTNAME, TRANSFER_RT_OPTNAME,
+    AUTH_AT_OPTNAME, TRANSFER_AT_OPTNAME,
+    AUTH_AT_EXPIRES_OPTNAME, TRANSFER_AT_EXPIRES_OPTNAME,
+    WHOAMI_ID_OPTNAME, WHOAMI_USERNAME_OPTNAME,
+    WHOAMI_EMAIL_OPTNAME, WHOAMI_NAME_OPTNAME,
+    internal_auth_client, remove_option, lookup_option)
+
+
+@click.command('logout',
+               short_help=('Logout of Globus, revoking credentials for '
+                           'the Globus CLI'),
+               help=('Remove credentials for the Globus CLI. '
+                     'Effectively stops any use of your CLI credentials as '
+                     'well'))
+@common_options
+def logout_command():
+    # build the NativeApp client object
+    native_client = internal_auth_client()
+
+    # prompt
+    confirmed = click.confirm("Are you sure you'd like to logout?")
+    if not confirmed:
+        return
+
+    safeprint('Logging out of Globus as {}'
+              .format(lookup_option(WHOAMI_USERNAME_OPTNAME)))
+
+    # remove tokens from config and revoke them
+    for token_opt in (TRANSFER_RT_OPTNAME, TRANSFER_AT_OPTNAME,
+                      TRANSFER_AT_EXPIRES_OPTNAME,
+                      AUTH_RT_OPTNAME, AUTH_AT_OPTNAME,
+                      AUTH_AT_EXPIRES_OPTNAME):
+        token = remove_option(token_opt)
+        if token:
+            native_client.oauth2_revoke_token(token)
+    for whoami_opt in (WHOAMI_ID_OPTNAME, WHOAMI_USERNAME_OPTNAME,
+                       WHOAMI_EMAIL_OPTNAME, WHOAMI_NAME_OPTNAME):
+        remove_option(whoami_opt)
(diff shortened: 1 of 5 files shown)


globus_cli/commands/logout.py
```diff
@@ -0,0 +1,43 @@
...
+    WHOAMI_EMAIL_OPTNAME, WHOAMI_NAME_OPTNAME,
+    internal_auth_client, remove_option, lookup_option)
+
+
+@click.command('logout',
+               short_help=('Logout of Globus, revoking credentials for '
+                           'the Globus CLI'),
+               help=('Remove credentials for the Globus CLI. '
+                     'Effectively stops any use of your CLI credentials as '
+                     'well'))
```
jaswilli: Technically we're not logging the user out of Globus (at least not all of Globus, since they could very well still have an active Auth session in their browser). We're probably better off just saying "Log out of the Globus CLI" or something.

Also, I'm wondering if we shouldn't avoid using "credentials" here, since it is a technical term in oauth2 and we're not really doing anything with credentials? Perhaps say we're invalidating/revoking tokens, or maybe not even give a description of what we're actually doing under the hood at all?

corpulentcoffee:
This probably needs to handle a second `globus logout` better.

```
~$ globus logout
Are you sure you'd like to logout? [y/N]: y
Logging out of Globus as shifflett@globusid.org

~$ globus logout
Are you sure you'd like to logout? [y/N]: y
Logging out of Globus as None
KeyError: 'transfer_refresh_token'

program returned status code 1
```

Maybe right at the top -- before you even call `click.confirm()` -- you just check to see if the user is even logged in to begin with?

jaswilli:
> As an unfortunate aside, we can't do much to direct users into the
> consents page right now. Maybe print some epilog help about how to go
> there and remove the consent? It's a dead entity, but we can't clean it
> up at all.

Yeah, I think putting a link to `https://auth.globus.org/consents` seems like a good idea. Not sure what language to use though. We don't want to be too dramatic ("To prevent the CLI from acquiring tokens in the future" seems sort of alarmist) and because of how named grants work, we can't even say "To prevent the CLI from getting tokens without being prompted..."

Maybe just a simple "If you wish to rescind consent granted to the CLI visit..."

jaswilli:
I think it might be useful to have a `--no-confirm` or some such option.

sirosen:
I have a version in progress that addresses these issues. For uniformity worth potential future prompting, I went with `--yes`. I assume that's inoffensive.

I went with language saying we're revoking tokens. That's consistent with developer docs because it's the most accurate description; it communicates to casual users that _something_ got revoked (they might not care what); and it can be ignored by anyone who doesn't care.

sirosen:
I think this is ready for another round of review.

globus_cli/commands/logout.py
```diff
@@ -0,0 +1,112 @@
...
+        fails.
+        """
+        if yes:
+            return
+        confirmed = click.confirm(prompt, default=default)
+        if not confirmed:
+            click.get_current_context().exit(1)
+
+    # prompt
+    _confirm("Are you sure you'd like to logout?")
```
jaswilli: Globus hates contractions...

sirosen: Well, I'm not expanding that to "you would". That's awkward and not how people talk or write.
`s/you'd like/you want/` momentarily.

globus_cli/commands/logout.py
```diff
@@ -0,0 +1,112 @@
...
+            click.get_current_context().exit(1)
+
+    # prompt
+    _confirm("Are you sure you want to logout?")
+
+    # check for username -- if not set, probably not logged in
+    username = lookup_option(WHOAMI_USERNAME_OPTNAME)
+    if not username:
+        _confirm(("Your username is not set. You may not be logged in. "
+                  "Would you like to try to logout anyway?"), default=True)
```
jaswilli: I'm not sure this check is worth anything--I feel like we should just do a best effort attempt to log them out regardless of what we find in the config file. 

Also, If we got remove it we could get rid of all the custom confirmation logic and just use `@click.confirmation_option(prompt='...')`.

sirosen: I didn't know about `confirmation_option`.
I'm fine with dropping this and instead printing a message.

I think we should say something since we can't print a username.
It also makes some sense of the numerous warnings you get about missing tokens if we say "We think you are not logged in..."

jaswilli: The no username case is already handled below (by just omitting the username from the message), which to me seems fine.

Honestly, I'm more concerned about the behavior on `logout` after a successful logout. Right now the output seems rather scary--it would be nice to be able to know that we're in a "logged-out" state and say that (along with the rescind consent hint).

This is definitely fine for now though 👍

PR #122, opened 2016-11-01
Add success confirmation output for `globus login`

### LINKED ISSUE #52: globus login, add output describing result/action after entering Authorization code
The login command could be improved with some additional explanation for what happened after the Authorization code is entered.  Below is a sample suggestion.

```
globus login
Please login to Globus here:
  https://auth.globus.org/v2/oauth2/authorize?code_challenge=<snip>

Enter the resulting Authorization Code here: <snip>

A credential (use appropriate term) for using the CLI was saved for you in (path).  It is valid until <date/time>

> (back at command prompt)
```

sirosen:
Depends on #120 -- don't merge before that.
Makes it clear that login worked, and that you can check your status with `whoami` or logout with `logout`.
Resolves #52

@stumartin as an example of the output in this new version, here's what I get:
```
$ globus login
Please login to Globus here:
---------------------------
https://auth.globus.org/v2/oauth2/authorize?code_challenge=<redacted>
---------------------------

Enter the resulting Authorization Code here: <redacted>

SUCCESS
You have logged in to the Globus CLI as sirosen@globusid.org

A set of valid tokens for accessing Globus have been saved for you in
~/.globus.cfg and are valid indefinitely unless revoked.


You can always check your current identity with
  globus whoami

Logout and invalidate these credentials at any time with
  globus logout

```
Can you confirm that this resolves the issue as you see it?

### DIFF
--- a/globus_cli/commands/login.py
+++ b/globus_cli/commands/login.py
@@ -13,6 +13,23 @@ from globus_cli.config import (
     internal_auth_client, write_option)
 
 
+_LOGIN_EPILOG = ("""\
+
+SUCCESS
+You have logged in to the Globus CLI as {}
+
+A set of valid tokens for accessing Globus have been saved for you in
+~/.globus.cfg and are valid indefinitely unless revoked.
+
+
+You can always check your current identity with
+  globus whoami
+
+Logout and invalidate these credentials at any time with
+  globus logout
+""")
+
+
 @click.command('login',
                short_help=('Login to Globus to get credentials for '
                            'the Globus CLI'),
@@ -81,3 +98,5 @@ def login_command():
     write_option(WHOAMI_USERNAME_OPTNAME, identity['username'])
     write_option(WHOAMI_EMAIL_OPTNAME, identity['email'])
     write_option(WHOAMI_NAME_OPTNAME, identity['name'])
+
+    safeprint(_LOGIN_EPILOG.format(identity['username']))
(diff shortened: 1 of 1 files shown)


globus_cli/commands/login.py
```diff
@@ -13,6 +13,23 @@
     internal_auth_client, write_option)
 
 
+_LOGIN_EPILOG = ("""\
+
+SUCCESS
+You have logged in to the Globus CLI as {}
+
+A set of valid tokens for accessing Globus have been saved for you in
+~/.globus.cfg and are valid indefinitely unless revoked.
```
jaswilli: Do we need to tell people where the tokens have been saved? It seems like we're just making it easy for someone to mess around with their config file directly. Not mentioning this also means we don't have to account for Windows not knowing what `~` means.

sirosen: Fair points. I wanted people to know that their tokens are being stored on the local filesystem though, since people could be easily confused by going out to a remote host, running the CLI, and then authenticating through a local web browser.

What about "A set of valid tokens for accessing Globus have been saved for you in your home directory."?

globus_cli/commands/login.py
```diff
@@ -13,6 +13,23 @@
...
+
+
+You can always check your current identity with
+  globus whoami
+
+Logout and invalidate these credentials at any time with
```
sirosen: Yeah, for consistency.

globus_cli/commands/login.py
bjmc: "valid indefinitely"?

How about 'will be renewed indefinitely'?

jaswilli: Personally, I'd not mention anything about it. From the perspective of the user, why do they need/want to know all the details of what's going on under the hood?

PR #126, opened 2016-11-01
Add transfer dry-run to improvrished testsuite

### LINKED ISSUE #104: Add poor man automated tests to `async-transfer` and `async-delete` using the `--dry-run` flag
As discussed in #100.

Depends on globus/globus-sdk-python#80. If that behavior is implemented such that `submission_id=None` and `submission_id=<UUID>` both work in the constructors (i.e. the former doesn't clobber the automatically-initialized value), we can just pass the `submission_id` option value along as-is (not caring at this layer about whether it's `None` or set to a UUID) to the `TransferData()`/`DeleteData()` constructors and get rid of the existing `if submission_id is not None: ...` blocks.

sirosen:
Poor man's tests grow and grow.
Resolves #104

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,9 +14,4 @@ install:
 cache: pip
 script:
   - flake8
-  # poor man's testsuite: run a bunch of commands and confirm that they don't
-  # crash and burn catastrophically
-  - globus --help  # parsing failures
-  - globus list-commands  # actually running a command
-  - globus get-identities 'abc@globus.org' --map-http-status "401=0"  # auth call (noauth)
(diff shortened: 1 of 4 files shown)


corpulentcoffee:
Supposedly this breaks the build, but it almost looks non-deterministic to me. At time of PR, 3.5-dev didn't pass, but after the merge, 3.3 and 3.4 done.

corpulentcoffee:
FWIW, if I run this repeatedly locally, I can reproduce the last command failing. The piped echo might be unreliable.

corpulentcoffee:
Okay so my guess is that the egrep call fails because the order of the dict isn't deterministic. I'll write a fix and open a PR.

PR #130, opened 2016-11-07
Use separated fgrep calls for dry-run submission

corpulentcoffee:
Because the dict key order in Python is not deterministic, this test was sporadically failing. Instead of using one big regular expression, break it into four individual calls to `fgrep` so that the non-deterministic order of the keys doesn't matter.

If you want to fix this by using the `sort_keys=True` flag on `json.dumps()`, feel free to close this.


sirosen:
This is fine. Setting `sort_keys=True` is something I want to apply to all JSON output in the future, but that's not an issue for today.

PR #131, opened 2016-11-09
Human text output for `task pause-info`

corpulentcoffee:
Adds human-readable output for both explicit activity manager pauses
and effective pause rules against an in-progress task. Revises the help
text to indicate that this command shows more than just the effective
pause rules.

Toward #35.

### DIFF
--- a/globus_cli/commands/task/pause_info.py
+++ b/globus_cli/commands/task/pause_info.py
@@ -1,14 +1,44 @@
 import click
 
+from globus_cli.safeio import safeprint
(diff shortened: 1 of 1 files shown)


sirosen:
I'm really glad I didn't try to do this myself. Not knowing the API, I'd have made a mess of it.

I'm reviewing this now. Just need to put some rules in place to do it.

sirosen:
Looks good in my testing / I was unable to break it. Merging.

PR #133, opened 2016-11-09
Move travis environment to ubuntu 14.04

jaswilli:


sirosen:
Looks like Travis is having an issue (just right now?) with 14.04 and py3.4?

Unless something is actually blocked on this, let's leave this (and the similar SDK PR) open and restart the builds tomorrow or later to see if Travis has it sorted out?

jaswilli:
Yeah, I can't figure out what could possibly just be causing the 3.4 build to fail other than something being in a bad state on Travis' end.

I'll just try restarting the build periodically and see if it resolves itself.

jaswilli:
Good news--the mystery failure has resolved itself.

PR #141, opened 2016-11-10
Require search filter when searching all endpoints

### LINKED ISSUE #140: 'globus endpoint search' help message
I typed 'globus endpoint search', expecting to get a helpful usage message, but instead got an error:

$ globus endpoint search
HTTP status:      400
request_id:       B1eYyBvVj
code:             BadRequest
message:          scope 'all' requires an additional filter parameter

jaswilli:
Closes #140

### DIFF
--- a/globus_cli/commands/endpoint/search.py
+++ b/globus_cli/commands/endpoint/search.py
@@ -26,6 +26,13 @@ def endpoint_search(filter_fulltext, filter_owner_id, filter_scope):
     """
     Executor for `globus endpoint search`
     """
+    if filter_scope == 'all' and not filter_fulltext:
+        raise click.UsageError(
+            'When searching all endpoints (--filter-scope=all, the default), '
+            'a full-text search filter is required. Other scopes (e.g. '
+            '--filter-scope=recently-used) may be used without specifying '
+            'an additional filter.')
+
     client = get_client()
 
     owner_id = filter_owner_id


sirosen:
I think this message might be potentially confusing when there are no arguments given (Steve's case in #140).

Here's how it looks:
```
$ globus endpoint search
Usage: globus endpoint search [OPTIONS] [FILTER_FULLTEXT]

Error: filter_scope "all" requires a search filter.
```

Suggestion:

```
$ globus endpoint search
Usage: globus endpoint search [OPTIONS] [FILTER_FULLTEXT]

Error: filter_scope "all" requires a search filter.Specify a filter or different filter_scope.
```

globus_cli/commands/endpoint/search.py
```diff
@@ -26,6 +26,9 @@ def endpoint_search(filter_fulltext, filter_owner_id, filter_scope):
     """
     Executor for `globus endpoint search`
     """
+    if filter_scope == 'all' and not filter_fulltext:
+        raise click.UsageError('filter_scope "all" requires a search filter.')
```
corpulentcoffee: I think this message might still be confusing unless you literally typed in `globus endpoint search --filter-scope=all` and/or are cognizant of the `--filter-scope` option.

Maybe: `When searching all endpoints (--filter-scope=all, the default), a full-text search filter is required. Other scopes (e.g. --filter-scope=recently-used) may be used without specifying an additional filter.`

sirosen: This seems better than my suggestion. Checking on how well `click` handles multiline or long messages in `UsageError`.

jaswilli: Yeah, I went with this one.

You don't get whatever fancy "short" line-wrapping goes on with other things (e.g. help output).

I force pushed an update if you want to re-pull and see for yourself.

PR #147, opened 2016-11-30
Better handling for `globus endpoint server list` when the user asks about an endpoint that isn't a GCS endpoint

### LINKED ISSUE #136: Shared endpoints and "endpoint server" commands
Should we put some logic into the `endpoint server` commands to make the experience with servers and shared endpoints better?

Currently if you do a `globus endpoint server list <endpoint_id>` on a shared endpoint the CLI will show you the list of servers backing the host endpoint(because the server list is returned by the API) but if you turn around and try to use `globus endpoint server update` you'll get an error message:

```
HTTP status:      400
request_id:       LIFg21vsH
code:             BadRequest
message:          That endpoint type doesn't have servers
```

In the web app we don't show servers for shared endpoints and I'm wondering if we shouldn't do the same in the CLI?

corpulentcoffee:
Per #136, listing servers on a share worked, but was misleading, as other commands wouldn't work. In this case, print a usage message about how to lookup information about the share or how to list the servers on the host.

Sort of related to that, listing servers on an S3 endpoint didn't raise an error but instead would just succeed and list an empty set. In this case, we can print the S3 URL instead. I could also raise an error if people prefer that.

Fixes #136.

### DIFF
--- a/globus_cli/commands/endpoint/server/list.py
+++ b/globus_cli/commands/endpoint/server/list.py
@@ -6,7 +6,8 @@ from globus_cli.safeio import safeprint
 from globus_cli.parsing import common_options, endpoint_id_arg
 from globus_cli.helpers import (colon_formatted_print, print_table,
                                 outformat_is_json)
-from globus_cli.services.transfer import get_client, print_json_from_iterator
+from globus_cli.services.transfer import (display_name_or_cname,
+                                          get_client, print_json_from_iterator)
 
 
 @click.command('list', help='List all servers belonging to an Endpoint')
@@ -22,14 +23,14 @@ def server_list(endpoint_id):
 
     if endpoint['host_endpoint_id']:  # not GCS -- this is a share endpoint
         raise click.UsageError(dedent("""\
-            {id} ({display_name}) is a share and does not have servers.
+            {id} ({0}) is a share and does not have servers.
 
             To see details of the share, use
                 globus endpoint show {id}
 
             To list the servers on the share's host endpoint, use
                 globus endpoint server list {host_endpoint_id}
-        """).format(**endpoint.data))
+        """).format(display_name_or_cname(endpoint), **endpoint.data))
 
     if endpoint['s3_url']:  # not GCS -- this is an S3 endpoint
         if outformat_is_json():
--- a/globus_cli/commands/endpoint/server/list.py
+++ b/globus_cli/commands/endpoint/server/list.py
@@ -45,4 +45,7 @@ def server_list(endpoint_id):
     if outformat_is_json():
         print_json_from_iterator(server_iterator)
     else:
-        print_table(server_iterator, [('ID', 'id'), ('URI', 'uri')])
+        print_table(server_iterator, [
+            ('ID', 'id'),
+            ('URI', lambda s: s['uri'] or "none (Globus Connect Personal)"),
+        ])
--- a/globus_cli/commands/endpoint/server/list.py
+++ b/globus_cli/commands/endpoint/server/list.py
@@ -14,6 +14,20 @@ def server_list(endpoint_id):
     """
     client = get_client()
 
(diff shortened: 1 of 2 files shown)


globus_cli/commands/endpoint/server/list.py
```diff
@@ -14,6 +17,28 @@ def server_list(endpoint_id):
...
+            "    globus endpoint show {id}\n"
+            "\n"
+            "To list the servers on the share's host endpoint, use\n"
+            "    globus endpoint server list {host_endpoint_id}".
+            format(**endpoint.data)
+        )
+
+    if endpoint['s3_url']:  # not GCS -- this is an S3 endpoint
+        if outformat_is_json():
+            safeprint(json.dumps({'s3_uri': endpoint['s3_url']}, indent=2))
```
sirosen: I noticed that this is `s3_uri` vs. `s3_url`.
Assuming this is just the result of being casual, can we use `s3_url` to be consistent with the API?

globus_cli/commands/endpoint/server/list.py
```diff
@@ -14,6 +17,28 @@ def server_list(endpoint_id):
...
+    if endpoint['host_endpoint_id']:  # not GCS -- this is a share endpoint
+        raise click.UsageError(
+            "{id} ({display_name}) is a share and does not have servers.\n"
+            "\n"
+            "To see details of the share, use\n"
+            "    globus endpoint show {id}\n"
+            "\n"
+            "To list the servers on the share's host endpoint, use\n"
+            "    globus endpoint server list {host_endpoint_id}".
+            format(**endpoint.data)
```
sirosen: As a stylistic note, I would have written this utterly differently (multiline string with `textwrap.dedent`, if it matters).
This way is fine, but should we set a convention on how long messages are represented/handled so that we have  a consistent style?

globus_cli/commands/endpoint/server/list.py
```diff
@@ -14,6 +17,28 @@ def server_list(endpoint_id):
     """
     client = get_client()
 
+    endpoint = client.get_endpoint(endpoint_id)
+
+    if endpoint['host_endpoint_id']:  # not GCS -- this is a share endpoint
+        raise click.UsageError(
+            "{id} ({display_name}) is a share and does not have servers.\n"
```
sirosen: I tested this against a Share with no `display_name` set.
Can we use `globus_cli.services.transfer.display_name_or_cname`?

You can do it with minimal noise if you add it under a name that won't conflict or as a positional arg to `format`.
Positional is probably safer.

sirosen:
Trying this on a personal endpoint gives somewhat disappointing results. e.g.
```
$ globus endpoint server list e4274365-6d04-11e5-ba46-22000b92c6ec
ID     | URI 
------ | ----
149201 | NULL
```

I'm okay merging this without a case for GCP, so long as we open an issue to track the need to update it to also handle GCPs.

corpulentcoffee:
Pushing some changes for this:

- c2cf32fe0c6d631d01ae63224b4e310a81397a9e switches the share exception message to be a multiline string that gets passed through `dedent()`
- b6bc0c136a021397aa366dab5f45b541024ead8b uses `url` consistently with S3 endpoints
- 1b4cb1cbee492aba9df85308b52a71251a593d4b uses the canonical name if display name isn't available
- c9f6933ad47ec1f5a543f9d5492c6cc787eae86a provides a better value than `NULL` for GCP servers in a `endpoint server list` and switches the columns output by `endpoint server show` [this one can be done multiple ways, given that the way Transfer decides to use or not use the servers subresource depends on if the endpoint if classic GCS, S3, or GCP]

sirosen:
This will have to be rebased off of the fix for the Auth change, but I see no further issues with it. It's good to go.

corpulentcoffee:
Rebased on the fixed `master` branch.

PR #158, opened 2017-01-18
Added examples.rst to docs to give basic use examples of Globus CLI

aaschaer:
Other than examples.rst, all files were created or changed by using "make docs".


docs-source/examples.rst
```diff
@@ -0,0 +1,250 @@
...
+    authenticated with Globus Auth. To do so run
+
+    .. code-block:: bash
+
+        $ globus login
+
+    and follow the instructions. 
+    If you are having problems with authorization, consult `Getting Started`_
+
+    .. _Getting Started: https://globus.github.io/globus-cli/# getting-started
```
sirosen: Sphinx actually offers a kind of ref that doesn't require giving a full URL.
I'd like to switch to that here. It will require modifying the `getting-started` section to have such a ref.

For an example: [Using a ref](https://github.com/globus/globus-sdk-python/commit/2498b89875afb97e949d63849453ed3761227fc5#diff-8c3a6a3481c95b3302a89e727c5dcf51R12) and [Defining a ref](https://github.com/globus/globus-sdk-python/commit/2498b89875afb97e949d63849453ed3761227fc5#diff-8c3a6a3481c95b3302a89e727c5dcf51R21)

Importantly, it renders to a relative link, and we plan to move the docs to docs.globus.org (globus/globus-sdk-python#104 applies to this repo too).

docs-source/examples.rst
```diff
@@ -0,0 +1,250 @@
...
+
+Finding an Endpoint
+-------------------
+
+Searches for a tutorial endpoint, then stores its ID in a local variable to
+allow for human-readable calls to further globus endpoint commands.
+
+.. code-block:: bash
+    
+    $ globus endpoint search "Globus Tutorial Endpoint"
```
sirosen: This applies throughout, but could you include sample (text-mode) output? I think it would make the examples more readable, especially in cases like this where we derive a value used below.

It's okay to doctor the output to look nicer (e.x. replace `ls` results with an invented set of files/dirs).

docs-source/examples.rst
```diff
@@ -0,0 +1,250 @@
...
+Globus CLI Examples
+===================
+
+.. note::
+    The following examples are all written for a Bash shell, if you are running
+    another shell you may need to modify the non globus commands and syntax
+    found in these examples
+
+    The following examples also assume the user has already successfully 
+    authenticated with Globus Auth. To do so run
```
sirosen: Very minor edit, but let's say "logged in to Globus".

`authenticated with Globus Auth` is more correct, but our communications to end-users don't draw distinctions between our services.

PR #162, opened 2017-02-01
Login command checks if the user is logged in

### LINKED ISSUE #153: `globus login` doesn't know when the user is already logged in
If the CLI already has valid tokens for the user, then "globus login" should just inmediately return success and "Already logged in".  Perhaps add a flag force re-login.

sirosen:
Check for existing credentials on disk, and skip out on login if they are found.
Add a `--force` option to login in case we're ever wrong.

Don't use validation yet -- there are still no docs or SDK support.

Closes #153


corpulentcoffee:
@aaschaer Can you test this out? Both the existing-tokens case and the `--force` flag.

aaschaer:
Trying to login with existing tokens told me I was already logged in unless the --force flag was used, and logging in without tokens still works normally. So looks like everything is working as expected to me.

corpulentcoffee:
Thanks @aaschaer!

PR #163, opened 2017-02-01
Fix `globus whoami` to have common-form output

sirosen:
For textual output delimited by colons, we commonly pad with spaces to get nice vertical alignment. However, the whoami information didn't fit the structure used by `colon_formatted_print()`, so it was done manually.

(@jaswilli, I'm not criticizing that -- it makes sense. But I care about that padding. :smile: )

Having that padding on everything is a minor cosmetic item, but easily achieved by reconstructing the whoami data as a "user document" suitable for `colon_formatted_print()`

### DIFF
--- a/globus_cli/commands/whoami.py
+++ b/globus_cli/commands/whoami.py
@@ -2,6 +2,7 @@ import click
 
 from globus_cli.safeio import safeprint
 from globus_cli.parsing import common_options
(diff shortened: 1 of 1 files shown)


jaswilli:
> I'm not criticizing that

Don't think I didn't notice `bugfix` in the branch name!

PR #214, opened 2017-02-25
Prefill named grant label on consent page

### LINKED ISSUE #176: login: auto-fill consent label with hostname
When running "globus login", on the Globus Auth consent screen it should auto-fill the "Provide a label for future reference" with the hostname.

jaswilli:
* Use the local system's hostname as the default value for the named grant label on the consent page.
* Upgrade SDK to version `0.5.1`.

screen shot 2017-02-25 at 11 31 38 am

Closes #176

### DIFF
--- a/globus_cli/commands/login.py
+++ b/globus_cli/commands/login.py
@@ -1,3 +1,5 @@
+import platform
+
 import click
 
 from globus_sdk import AuthClient, AccessTokenAuthorizer
@@ -78,8 +80,12 @@ def do_login_flow():
     # build the NativeApp client object
     native_client = internal_auth_client()
 
-    # and do the Native App Grant flow
-    native_client.oauth2_start_flow_native_app(refresh_tokens=True)
+    # and do the Native App Grant flow, prefilling the
+    # named grant label on the consent page if we can get a
+    # hostname for the local system
+    label = platform.node() or None
+    native_client.oauth2_start_flow(
+        refresh_tokens=True, prefill_named_grant=label)
 
     # prompt
     linkprompt = 'Please login to Globus here'
--- a/globus_cli/parsing/excepthook.py
+++ b/globus_cli/parsing/excepthook.py
@@ -34,14 +34,6 @@ def exit_with_mapped_status(http_status):
         sys.exit(1)
 
 
-def pagination_overrun_hook():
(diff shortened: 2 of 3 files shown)


sirosen:
Looks like bumping up to `0.5.1` results in CLI failures because we dropped the silly `PaginationOverrunError` exception type in `0.5.0` -- but the CLI handler is still importing it.

I don't mind the idea of fixing it here, or doing a separate PR + rebase if you prefer.

If this depends on `0.5.1`, then `globus_cli.version` needs to be updated to read `0.5.1.0` to stay on our current convention.

With the version updated, and whichever resolution you prefer for the current issue, this looks okay.
I'm going to look at `platform.node()` a bit and see if we can trust it everywhere (e.x. Windows?).

sirosen:
Scratch that note about checking on `platform` -- it's designed to be your portability/compatibility detection layer, so if it doesn't work on a given OS or arch, we're hosed anyway. It's been a while since I've looked at that lib.

So, yeah. Some minor kinks to work out, then :ship: :it:

jaswilli:
Yeah, I just fixed the `PaginationOverrunError` thing. For some reason the tests were passing locally even I'm definitely using SDK v0.5.1 locally.

jaswilli:
> If this depends on 0.5.1, then globus_cli.version needs to be updated to read 0.5.1.0 to stay on our current convention.

I was thinking we'd just get anything else we might want to release soon-ish merged in before we bumped the version and released. I don't much care either way though so if you'd prefer I'm happy to bump it in this PR.

sirosen:
I'm wishy-washy about bumping the version, and already changing my tune.

All of the previous changes were still depending on `0.4.6`, so, in theory, they could have been released as `0.4.6.1`. But that argument seems flimsy -- if we fully believe that this should be `0.5.1.0`, we should have incremented the version to `0.4.6.1` immediately after we released `0.4.6.0`.
It kind of makes sense to not bump the version here either, so nevermind that change.

I would like to start updating the version immediately after each release, so that `globus version` on `master` always tells you that you're ahead of PyPi. Let's start doing that this time around?

jaswilli:
Maybe `globus version --version` can return the pre-release version number. 😂

PR #216, opened 2017-02-25
Add `globus update` command

### LINKED ISSUE #154: globus update
Currently the instructions for updating the globus cli are:

source "$HOME/.globus-cli-virtualenv/bin/activate"
pip install -U globus-cli
deactivate

Can we add a "globus update" command that does this?  That will make it easier to keep the CLI updated without having to remember this virtualenv magic every time.

sirosen:
Resolves #154

In spite of some initial reservations, a second look at this leaves me with a fairly comfortable solution -- one which should be safe for all foreseeable situations.

Important attributes:
- extensive commenting
- generally verbose messages
- support for installing `master` (or any other revision) off of GitHub
- deferment of upgrade until after most application teardown
- non-strict dependency on `pip`
- latest version skips update
- basic safety checking (can we at least reach PyPi?)

I'd like this to receive fairly careful vetting, as I think it is in principle a dangerous command to expose, but a useful one.

To test, I mostly setup virtualenvs and the did `pip install . -U` repeatedly to get this version installed and/or revert its changes. It's also useful to fiddle with `globus_cli.version` to poke at different behaviors.

The install-from-GitHub options are hidden, but you can find them by looking at the changeset.

### DIFF
--- a/globus_cli/commands/main.py
+++ b/globus_cli/commands/main.py
@@ -2,6 +2,7 @@ from globus_cli.parsing import globus_main_func
 
 from globus_cli.commands.list_commands import list_commands
 from globus_cli.commands.version import version_command
+from globus_cli.commands.update import update_command
 from globus_cli.commands.config import config_command
 
 from globus_cli.commands.login import login_command
@@ -27,6 +28,7 @@ def main():
 
 main.add_command(list_commands)
 main.add_command(version_command)
+main.add_command(update_command)
 main.add_command(config_command)
 
 main.add_command(login_command)
--- a/globus_cli/commands/update.py
+++ b/globus_cli/commands/update.py
@@ -1,0 +1,108 @@
+import atexit
+import click
+
+from globus_cli.safeio import safeprint
+from globus_cli.version import get_versions
+from globus_cli.parsing import common_options, HiddenOption
+
+
+@click.command('update', help='Update the Globus CLI to its latest version')
+@common_options(no_format_option=True, no_map_http_status_option=True)
+@click.option('--yes', is_flag=True,
+              help='Automatically say "yes" to all prompts')
+# hidden options to fetch branches or tags from GitHub -- one turns this mode
+# on or off, and the other is used to set a non-master target
+@click.option('--development', is_flag=True, cls=HiddenOption)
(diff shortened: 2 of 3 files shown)


aaschaer:
Results of my attempts to break this.

I could not break this on ubuntu, all combinations of python2 python3 pypy and anaconda with every version of pip I tested ran globus update without issue.

The CSIL OSX machines I tested on are having image issues, so I was only able to test python2.7 and 3.5 with pip 9.0.1, which worked fine. I asked to be told when the issues are resolved so I can run further tests on OSX if desired, or if someone else wants to?

I did manage to break this on windows 7. 
Attempting to run globus update using pypy ran into difficulties with wheel and the pip install failed.
When using older versions of pip (6.0) globus update failed on "WinError 5 Access is denied" which might be fixable by running globus update as an admin, but since my user had ownership of the virtual env to be updated, and newer versions of pip ran fine, I'm counting this as a break.
Once I managed to get virtualenv to play nice with anaconda on windows, anaconda python2 and python3 had no issues other than the failures on older versions of pip.

I was surprised how stable this was on windows, which I guess speaks to the power of using a virtualenv for install, and think this is reasonably stable unless we want more OSX testing or more flavors of python?

sirosen:
I assume that CSIL is running the latest or second-latest OSX version.
I'm not sure if that's an issue or not. @jaswilli expressed concerns about old versions of the OS, setuptools, and anaconda -- however, remember that GCP _officially_ only supports the latest OSX.

It's arguably fine for features, especailly nonessential ones, to only be tested on our officially supported platform versions.

Given that the only condition under which it is known to break is on old versions of `pip` (6.x was last updated 2015-04-07) on Windows, I'm left curious about `pip` 6.x on Linux and OSX.

I'll see if I can test that out.

sirosen:
I just tried `pip==6.1.1` on OSX 10.9 and Debian 9.
It worked fine on both of those platforms. As I had to install virtualenv and pip without admin privs on the OSX, the configuration was a little weird, which bolsters my confidence a little bit more.

I also thought to try this with all kinds of bad settings for the `PYTHONPATH`, but nothing seemed to break it (though it is possible to break `pip install -I 'pip==6.1.1'` with a bad `PYTHONPATH`, which is how I thought of this).

Issue #63, opened 2016-10-17
Transfer: endpoint create/update functionality


jaswilli:
- Missing option to add `subscription_id`
- Support `network_use`, `location`, and `disable-verify`

sirosen:
This has been reduced to a dup of #134  Closing.

Issue #161, opened 2017-01-24
Missing endpoint update functionality


tuecke:
"globus endpoint update" is missing functionality, relative to the ssh CLI.  Specifically:

  --disable-verify      Disable checksum verification of transfers to and from the endpoint
  --no-disable-verify   Unset the --disable-verify setting
  --managed-endpoint    Set managed endpoint subscription
  --no-managed-endpoint   Unset the managed endpoint subscription
  --subscription=UUID   Specifies subscription id to use for --managed-endpoint
  --network-use=LEVEL   Set network usage level: normal, minimal, aggressive, or custom
  --max-concurrency=N   
  --preferred-concurrency=N
  --max-parallelism=N   
  --preferred-parallelism=N
  --location=LOCATION   Set location to "latitude,longitude"

jaswilli:
We need to go through this list and figure out which of these things aren't currently supported via the Transfer API and then decide what, if any, enhancements to the API we're going to push on.

A few of these options are covered in some existing issues:

* #63 (`subscription_id`, `network_use`, `location`, `disable-verify / no-disable-verify`
* #134 (`managed-endpoint / no-managed-endpoint`)

sirosen:
IIRC, `--max-concurrency`, `--preferred-concurrency`, `--max-parallelism`, and `--preferred-parallelism` are all missing API support.
I think that they are ignored on update, just like `subscription_id`, but given that the docs don't mention this, it's possible that I'm misremembering.


`--subscription` is a bigger issue because it's a piece of administrative functionality -- even if it was in the hosted CLI, it doesn't really belong in a user-facing tool.
I've floated my thoughts on a [plugin architecture](https://github.com/globus/globus-cli/issues/155) to add functions for privileged users, but I also think this could be done in a less general way, tightly scoped to a special `superuser` plugin. (i.e. detect and load a package if it is present).

corpulentcoffee:
On Wed, Jan 25, 2017 at 7:33 AM, Stephen Rosen <notifications@github.com>
wrote:

> IIRC, --max-concurrency, --preferred-concurrency, --max-parallelism, and
> --preferred-parallelism are all missing API support.
> I think that they are ignored on update, just like subscription_id, but
> given that the docs don't mention this, it's possible that I'm
> misremembering.
>
network_use (along with max_concurrency, max_parallelism,
preferred_concurrency, preferred_parallelism) have API support as long as
the endpoint is managed.

tuecke:
[...]

PR #227, opened 2017-03-07
Add endpoint create and update options requested in issue 63

aaschaer:
As requested in #63, I added network_use, location, and disable-verify options to endpoint create and update

I chose not to add subscription_id yet, as the API says it is currently only set manually by us, and I felt having it show up as an option that did nothing might be confusing. I can add it with help text explaining, or add it then comment it out if either of those are better options.

Since the API says a network_use value of custom requires max and preferred concurrency and parallelism to be set, I added --custom-concurrency and --custom-parallelism options that both take two ints. However it looks like these might be silently ignored for non managed endpoints?

To simplify user input I made location take two floats for latitude and longitude, and added a mutually exclusive --location-automatic option

I did manual testing of these options, but I was wondering how I should start making tests to confirm that my changes do what I think they do?

### DIFF
--- a/globus_cli/commands/endpoint/update.py
+++ b/globus_cli/commands/endpoint/update.py
@@ -15,11 +15,43 @@ from globus_cli.services.transfer import get_client, assemble_generic_doc
 def endpoint_update(endpoint_id, display_name, description, organization,
                     department, keywords, contact_email, contact_info,
                     info_link, public, default_directory, force_encryption,
-                    oauth_server, myproxy_server, myproxy_dn):
+                    oauth_server, myproxy_server, myproxy_dn, network_use,
+                    custom_concurrency, custom_parallelism, location_automatic,
+                    location, disable_verify):
     """
     Executor for `globus endpoint update`
     """
-    client = get_client()
+    # require custom_concurrency and custom_parallelism for custom network_use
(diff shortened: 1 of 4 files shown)


globus_cli/commands/endpoint/update.py
```diff
@@ -15,11 +15,43 @@
...
+        raise click.UsageError(
+            "cannot combine --location and --location-automatic")
+
+    # set location based on args
+    if location_automatic:
+        location = "Automatic"
+    elif location:
+        location = "{},{}".format(location[0], location[1])
+    else:
+        location = None
```
sirosen: I'm not super-ecstatic about repeating all of this logic. Can we dump a nice little helper somewhere to handle post-parsing validation/transformation on all of these options?

This is the only issue I have with this PR -- everything else seems sane/fine.

aaschaer:
Moved the option logic to parsing.shared_options, but not sure if that's the best place for it, and the way I call and return from the helper feels really clunky. Is there is a better way to do that? Pass a subset of *args or something?

aaschaer:
This will likely have a merge conflict with #240, and should include the requested functionality of #161 as well, so I'm going to close both PR's and make a new one that addresses all update/create issues for 1.0

Issue #57, opened 2016-10-17
Transfer: add options to task list


jaswilli:
- filter
- limit
- search by label (covered by "filter" above?)

PR #228, opened 2017-03-07
Add --limit and filtering options to task list

aaschaer:
Adds options to task list requested in #57

I felt it was easier on the user to split filtering up into several options rather than have them try to figure out how to make a filter string and then figure out what the API errors they were getting back meant.

I'm not ecstatic with the multi option method of passing more than one filter of the same kind, but it lets me use a click.Choice type for telling the user what values are allowed, and I think more typing is a lesser evil to unclear options?

### DIFF
--- a/globus_cli/commands/task/list.py
+++ b/globus_cli/commands/task/list.py
@@ -6,16 +6,141 @@ from globus_cli.helpers import outformat_is_json, print_table
 from globus_cli.services.transfer import print_json_from_iterator, get_client
 
 
+class ISOTimeType(click.ParamType):
+    """
+    Enforces time inputs to be in either YYYY-MM-DD or YYYY-MM-DD HH:MM:SS
+    And that each value is a valid int that falls in the correct range
+    """
+
+    name = "ISO TIME"
+
+    def convert(self, value, param, ctx):
+        try:
+            if len(value) not in [10, 19]:
+                raise ValueError
+
+            year = int(value[:4])
+            if year <= 0:
+                self.fail("Invalid year {}".format(year))
+            month = int(value[5:7])
+            if month not in range(1, 13):
+                self.fail("Month {} must be in range 1-12.".format(month))
+            day = int(value[8:10])
+            if day not in range(1, 32):
+                self.fail("Day {} must be in range 1-31.".format(day))
+
+            if len(value) == 19:
+                hour = int(value[11:13])
+                if hour not in range(24):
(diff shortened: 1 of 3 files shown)


globus_cli/commands/task/list.py
```diff
@@ -6,16 +6,141 @@
...
+                if hour not in range(24):
+                    self.fail("Hour {} must be in range 0-23.".format(hour))
+                minute = int(value[14:16])
+                if minute not in range(60):
+                    self.fail(
+                        "Minute {} must be in range 0-59.".format(minute))
+                second = int(value[17:19])
+                if second not in range(61):
+                    self.fail(
+                        "Second {} must be in range 0-60.".format(second))
```
sirosen: I don't object to us doing our own parsing logic if there's a reason, but why not use [`datetime.strptime`](https://docs.python.org/2/library/datetime.html#datetime.datetime.strptime) or [`time.strptime`](https://docs.python.org/2/library/time.html#time.strptime)?

If we have a reason that we can't use those (e.g. the tighter coupling with `click`'s `fail()`), that should be stated in a comment or docstring.

If the error messaging is the only reason, I'd contest the necessity of custom parsing. A message like "20160124 is not of the format YYYY-MM-DD or of the format YYYY-MM-DD HH:MM:SS" seems fine.

globus_cli/commands/task/list.py
```diff
@@ -6,16 +6,141 @@
 from globus_cli.services.transfer import print_json_from_iterator, get_client
 
 
+class ISOTimeType(click.ParamType):
+    """
+    Enforces time inputs to be in either YYYY-MM-DD or YYYY-MM-DD HH:MM:SS
+    And that each value is a valid int that falls in the correct range
+    """
+
+    name = "ISO TIME"
```
sirosen: Can we make this `ISO_TIME`? The Type name will sometimes show up as a metavar in help output, and we've had a past complaint that spaces in metavars were confusing.
(I know it's silly, but humor me? :smile: )

aaschaer:
Made requested changes.
The code feels much cleaner now, thank you for the suggestions.

sirosen:
Yeah, I think relying on the stdlib to handle parsing times is a big improvement in terms of readability.

We could even parameterize `ISOTimeType` over format strings in the future, if we need/want to support different time formats in other parts of the CLI.

I don't know that applying all these filters is ever going to get more understandable than it is at this point, absent extensive docs.

Issue #58, opened 2016-10-17
Transfer: add options to task event-list


jaswilli:
- filter
- limit

PR #229, opened 2017-03-08
Add --limit and filtering options to task event-list

aaschaer:
Addresses issue #58


sirosen:
I think there's potential for someone to get confused by passing both filter options. I'll open another issue to track that longer term.

PR #230, opened 2017-03-09
RetryingTransferClient

aaschaer:
Preliminary attempt to make the CLI more robust on network errors, any feedback is much appreciated.

I tested the retry function itself and that resources are not breaking using the class, but would like to set up a mock TransferClient that failed predictably to really test this.

One issue, even though task submission is safe on retry, the Duplicate and Accepted responses still differ slightly, how should we address this?

### DIFF
--- a/globus_cli/services/transfer.py
+++ b/globus_cli/services/transfer.py
@@ -1,13 +1,57 @@
 import uuid
 import json
+import random
+import time
 
 from globus_sdk import TransferClient, RefreshTokenAuthorizer
+from globus_sdk.exc import NetworkError
 
 from globus_cli import version
 from globus_cli.config import (
     get_transfer_tokens, internal_auth_client, set_transfer_access_token)
 from globus_cli.safeio import safeprint
 from globus_cli.helpers import print_table
+
+
+class RetryingTransferClient(TransferClient):
+    """
+    Wrapper around TransferClient that retries safe resources on NetworkErrors
+    """
+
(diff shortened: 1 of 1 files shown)


sirosen:
Personally, I'm pretty comfortable with this as it is, and allowing the Task submission responses to change subtly.

Given the potential dangers of this change, I'd like to get some feedback from @corpulentcoffee and @jaswilli. Not only for the extra eyes, but also because they have experience using the Transfer API, and may be aware of corner cases or "gotchas" of which I am not.
Having some basic, reasonable handling for the occasional network error is highly desirable (necessary?) for v1, but we haven't discussed possible avenues before now.

jaswilli:
I'm not entirely convinced that the added complexity is worth it. If someone's network connection is really this bad, I'm not sure how this is going to make the final outcome much better (i.e., I suspect a whole bunch of other things are going to fall over on them...).

That said, I don't see anything obviously wrong with the implementation and it's certainly pretty well isolated. So given that we should be able to remove it fairly painlessly down the road if we want, this seems fine to me if no one else has any objections.

sirosen:
I was the one who set Aaron down this path, following my own opinion (above) that it's important.

I'd like to hear @corpulentcoffee weigh in to share any strong feelings or lack thereof. :neutral_face: 
Assuming he has no objections, let's go ahead for now. As noted, it's easy enough to back out the change if necessary.

PR #233, opened 2017-03-10
Testing framework, expansion of basic tests, and a config error fix needed for testing

aaschaer:
Added a testing framework and moved and expanded the basic_tests into a python file to make sure the framework itself is working.

I choose to handle refresh tokens / simulating having the Cli Test User logged in by modifying the config file to use the testing refresh tokens and Cli Test User data, then restoring the previous config state during CliTestCase class tear down.

One thing I'd especially like looked at is my changes to config.py. I ran into an error with lookup_option still finding options even after they had been cleared from the config file. My guess is that this is because lookup_option was using the sdk's GlobusConfigParser unlike the rest of config so it could find values stored inside the GlobusConfigParser even if those options no longer existed. I think I fixed this in a way that won't break anything, but would like at least one more pair of eyes.


aaschaer:
Travis is failing on some assumptions I made about the local environment, I'll fix this and make a new PR

globus_cli/config.py
```diff
@@ -87,8 +86,14 @@ def get_config_obj(system=False):
...
-    p = globus_sdk.config._get_parser()
-    return p.get(option, section=section, environment=environment)
+    conf = get_config_obj()
+    try:
+        if environment:
+            return conf["environment " + environment][option]
+        else:
+            return conf[section][option]
+    except KeyError:
+        return None
```
sirosen: You said this is the part that makes you nervous. I think this change is okay, and probably for the better long-term.

It _is_ a subtle behavioral change, and could break some esoteric setups, but it's acceptable.
The SDK config parser uses the standard python `ConfigParser`, and reads from `~/.globus.cfg` and `/etc/globus.cfg`. The CLI config obj is from `configobj`, which, because it thinks of config as something which can be flushed back to disk, can't handle multiple files.

So, if someone had something in `/etc/globus.cfg` which they were relying on in the CLI, it would break.
No one normally will, and I'm okay with breaking those who do -- they'll be primarily internal if they even exist.

PR #234, opened 2017-03-13
Testing framework, basic tests in python, and config bug fix needed for testing

aaschaer:
Re-making testing framework PR.

Should I try to get appvoyer running on this repo, or should we use the same account used for the SDK?


sirosen:
I think this is fine to merge as-is. It may suffer from the same issue I was working on in globus-sdk-python#156 , but we might as well fix that in the SDK and then port the fix here.

AppVeyor is not a necessity at present -- we'll configure it at some later date. I'm sure it will break in an interesting and spectacular way, which we really can't deal with prior to GW.

Issue #226, opened 2017-03-05
Make `-v/--verbose` a global option


sirosen:
I'm thinking the following:

`-v` is a "counting" option. i.e. we set `verbosity=N` where `N` is the number of occurrences of the option.
Allows `-vvv`, etc. a.k.a. "angry scripting mode", since it often means "Why won't you say what's wrong?!"

We allow a config parameter, `verbosity`, which is an int. In actual fact, `verbosity` is `max(config_verbosity, command_line_verbosity)`
Default is, of course, 0.

`verbosity` is attached to the `CommandState`, for easy access.

In addition to being able to get it as an int, have a bool helper function, `is_verbose()` for commands which only have a simple notion of verbosity.

sirosen:
I realized that we'll also want a way of tuning verbosity down. Even if you set a high verbosity in your config, you might want less verbose output in a script.

Consider the current behavior of `whoami`.

There are two ways I can think of handling this. Maybe there are others?

1. Add `-q/--quiet` as the inverse of `-v`
2. Add an option to explicitly set the int verbosity level

sirosen:
After discussion we agreed on something which should have been obvious: the problem caused by verbosity being configurable but us wanting to be able to force `verbosity=0` is one of my own making.

Verbosity should _not_ be a configurable option.
Even so, it should be a global option, it should be a counting style of option, and an `is_verbose()` helper would not go amiss.

PR #235, opened 2017-03-15
Add local http server login flow

aaschaer:
As we've run into the "cannot automate login testing" issue again I tested this manually both locally and over ssh, and everything seems to be working.

Not a functionality issue, but the redirect page on the local server is *very* bare-bones, is that something worth fancying up a bit for 1.0?

Also, once #226 has been addressed we might want to use the enable_requests_logging() helper function on certain levels of verbosity?


globus_cli/commands/login.py
```diff
@@ -97,7 +115,50 @@ def do_login_flow():
...
+    exchange_code_and_store_config(native_client, auth_code)
+
+
+def do_local_server_login_flow():
+    """
+    Starts a local http server, opens a browser to have the user login,
+    and gets the code redirected to the server (no copy and pasting required)
+    """
+    # start local server and create matching redirect_uri
+    for port in range(8000, 8999):
```
jaswilli: I think instead of trying to play this game we should let the OS give us a port at runtime. Something like:

```
server = start_local_server(listen=('127.0.0.1', 0))
_, port = server.socket.getsockname()
redirect_uri = 'http://localhost:{}'.format(port)
```

Passing in `0` tells the underlying socket builder to just ask the OS for a free port.

aaschaer: That is much cleaner, thank you.

jaswilli:
I think we should probably make "local server" login the default and then fall back to the copy-paste method if the remote session check comes back true. We can still provide a flag to opt-out of the behavior (`--no-local-server` or something).

aaschaer:
Made local server default as requested.

jaswilli:
This seems like a good first pass at this feature. I'll merge and then come back around on it with a better final destination page.

Issue #225, opened 2017-03-02
Make `bookmark locate` a behavior of `bookmark show`


sirosen:
And improve the helptext.

We agreed that it makes more sense to talk about "resolving a bookmark to a path".

sirosen:
Updated issue title. The default behavior of `bookmark show` should be the current behavior of `bookmark locate`.
`bookmark show --verbose` should have the current behavior of `bookmark show`

Issue #172, opened 2017-02-03
get-identities: output only identity id or name


tuecke:
get-identities does not appear to have an easy way to get back just the identity id for a username.  For example, what I want is:

$ globus get-identities tuecke@uchicago.edu --xxx
b04b1e76-d274-11e5-bf36-07c9790e3d20

Perhaps --xxx should be "--format short".  Or "--fields identity_id --format csv".  

Likewise 
$ globus get-identities --identities b04b1e76-d274-11e5-bf36-07c9790e3d20 --xxx
tuecke@uchicago.edu

For multi-valued input, have one output per line, the same order as they appear on the command line.  For example:

globus get-identities tuecke@uchicago.edu tuecke@globusid.org --xxx
b04b1e76-d274-11e5-bf36-07c9790e3d20
b0497184-d274-11e5-bf33-130e800f88c2

sirosen:
Rather than mucking about with another textual format definition, I'd suggest making the current output the result of `--verbose` and defaulting to this output type.

That's simple, and consistent with the behavior of `globus ls --long` and `globus whoami --verbose`.

aaschaer:
closing as resolved by #253

PR #236, opened 2017-03-15
Add global verbosity flag, move bookmark locate behavior into bookmark show

aaschaer:
Addresses issues #225 and #226 

I figured since having a --verbose flag was a prerequisite of changing bookmark show's --verbose output I might as well combine these into one PR.

I picked arbitrary but hopefully reasonable levels for warnings and logging for the different verbosity levels. I was wondering if maybe -v should be reserved for commands that change their outputs with any verbosity, and only start logging/warning at -vv, but having -vvv be the "angry scripting" level felt better than pushing it to -vvvv, and warnings and error logs shouldn't come up much in the first place?

I went ahead and removed bookmark locate, since we were considering renaming it anyways and its redundant presence seemed potentially confusing.

Since the resolve_id_or_name bookmark helper could return either a GlobusHTTPResponse or a dict, I updated print_json_response to check type, but also had it fall back to just safeprint if it didn't see either of those types. Maybe this should raise a value error instead?


aaschaer:
Adding issue #172 to this PR since it also requires a --verbose option. If this is getting to be too much for one PR I can take out #172 and #225 and make them as separate PRs after the verbosity flag gets merged.

sirosen:
I don't have an issue with including the bookmark changes here -- those seem fine, but can we separate the `get-identities` changes?

I see strange behaviors that we need to resolve, like
```shell
$ globus get-identities sirosen@globusid.org
ae332d86-d274-11e5-b885-b31714a110e9
$ globus get-identities sirosen@globusid.org ae332d86-d274-11e5-b885-b31714a110e9
ae332d86-d274-11e5-b885-b31714a110e9
sirosen@globusid.org
```

aaschaer:
Moved out the get identities changes as requested, we should probably talk about how to address that issue separately sometime.

Looks like my rebase -i skills still need sharpening as the task list and iso time commits are showing up on this PR now? If there is trouble merging let me know and I'll just remake the PR.

Issue #207, opened 2017-02-10
`globus task wait` doesn't respect `--format JSON`


sirosen:
Task wait accepts the `--format` option, but it doesn't actually produce JSON output.

I'm not sure that it needs to support `--heartbeat --format JSON` because streaming JSON output is messy. We could print heartbeats to stderr.

When done, there should at least be a result document with the final task status.

PR #239, opened 2017-03-16
Add json outputs to task wait, and move heartbeats to stderr

aaschaer:
Addresses issue #207 

Also made the timeout of the SDK call equal to the --polling-interval option to make prevent the SDK calls from timing out before the CLI polling interval was reached.

I also moved the output of --meow to stderr, and would like to thank whoever made this option for making my day~


sirosen:
`--meow` is the most important option in the entire CLI. This all looks good; nothing particularly noteworthy to talk about.

Thanks for taking care of this!

Issue #187, opened 2017-02-08
endpoint share vs. endpoint create --shared


tuecke:
Currently the CLI uses "globus endpoint create ..." to create GCS and GCP endpoints, but "globus endpoint share ..." to create shared endpoints.  I'm debating if endpoint share should be folded into endpoint create:

globus endpoint create --shared HOST_ENDPOINT_ID:HOST_PATH DISPLAYNAME

What was the reason for separating it out?

sirosen:
This was what we reached in #10 a long time back.
That decision was based in part on the API's model, which is that shared endpoint creation is a separate factory method because its parameters are so different.

At the time, we were doing long opts for everything and there were no positional arguments, so it was possible to consider having these folded together. However, at present I'm not sure it's feasible or that the results would be desirable.
In the current CLI, it would be hard, maybe impossible, to inject a new positional argument at parsing time based on the presence of `--shared`, and there are messy implications for `globus endpoint create --help`.

i.e. the first positional is `DISPLAY_NAME` for `globus endpoint create` but is an `endpoint_plus_path` type for `globus endpoint share`.
I don't know how this would work if we tried to fold them back down together.

Do we want to discuss further, or just close this out?

tuecke:
I'm not sure I understand the positional argument issue, if the argument is "--shared HOST_ENDPOINT_ID:HOST_PATH", not just "--shared".  

> That decision was based in part on the API's model, which is that shared endpoint creation is a separate factory method because its parameters are so different.

I don't recall the reasons why /shared_endpoint is a separate factory method in the API from /endpoint, but the parameters really aren't so different.  The doc says a shared endpoint document is just a subset of the full endpoint document.  I think that's really true of GDP endpoints also.

There is also a terminology issue.  We generally do not use the phase "create a Globus share".  It is a "shared endpoint".  A "share" is not a noun we use.  The noun is "endpoint", with "shared" being a qualifier.  

My leaning is to fold shared endpoint creation into "globus endpoint create".

sirosen:
> the argument is "--shared HOST_ENDPOINT_ID:HOST_PATH", not just "--shared".

That makes sense of this.

Yeah, I've got no problem with converting this to `--shared` as an option on create. I thought you wanted to be able to `globus endpoint create id:path name --shared`, which we can't do with the way parsing works.

The decision we made in #10 was based in part on the notion that we want consistent modeling across the API, SDK, and CLI, and that the API models shared endpoint creation as a distinct operation.
[...]

PR #240, opened 2017-03-16
Move endpoint share to a --shared option for endpoint create

aaschaer:
Addresses issue #187 

This required some pretty big changes to parsing since the create/update options can no longer know if they are dealing with a share or not. I tried to make an option validation function for both endpoint create and update to make up for this, but only endpoint create knows what kind of endpoint it is dealing with before making calls to the API, so I'm only validating endpoint create for now. I guess we could make a get call to find the endpoint type, validate the option parameters, then make the update call?


aaschaer:
Realized this should probably be in one PR with #227 to avoid merge conflicts.

PR #244, opened 2017-03-21
Endpoint create and update

aaschaer:
Addresses issues #63 #187 and #161 

As it is unclear whether or not subscription_id actually is actually supported by the API, I made it a hidden option for now.

Since moving endpoint share into endpoint create --shared removed some option sanitization logic, I added a validation function that makes sure users aren't passing options that don't work with the given endpoint type.

I'm uncomfortable with the process of figuring out what kind of endpoint we are dealing with since --globus-connect-server works as a "not-globus-connect-personal" flag, and is ignored if the user also passes --shared. It would line up more with the API if we make GCS the default that is then overridden by --globus-connect-personal xor --shared, but that seems more confusing for the user?

Since the changes to these commands were pretty major, I added some unit tests for them to make sure I hadn't broken everything. These required a few testing framework updates to make SDK calls inside tests and to clean up old files in the .globus/sharing folder.


globus_cli/commands/endpoint/update.py
```diff
@@ -11,23 +13,35 @@
...
+                    myproxy_server, oauth_server, force_encryption,
+                    default_directory, subscription_id, network_use,
+                    max_concurrency, preferred_concurrency,
+                    max_parallelism, preferred_parallelism):
     """
     Executor for `globus endpoint update`
     """
+    # validate params. Requires a get call to check the endpoint type
+    args, _, _, values = inspect.getargvalues(inspect.currentframe())
+    params = dict((k, values[k]) for k in args)
```
sirosen: I'm curious why you're using `inspect` for this purpose. It seems to me much cleaner to take the keyword args dict, `**kwargs`, for all of the arguments above.

`kwargs` then replaces all usage of `params` below.

globus_cli/commands/endpoint/create.py
```diff
@@ -25,32 +27,49 @@
...
-        contact_info=contact_info, info_link=info_link,
-        force_encryption=force_encryption, public=public,
-        default_directory=default_directory,
-        myproxy_server=myproxy_server, myproxy_dn=myproxy_dn,
-        oauth_server=oauth_server)
+    # validate params
+    if shared:
+        endpoint_type = "shared"
+    args, _, _, values = inspect.getargvalues(inspect.currentframe())
+    params = dict((k, values[k]) for k in args)
```
sirosen: As below, I think taking `**kwargs` to the function is nicer than using `inspect`.

sirosen: Come to think of it, using `**kwargs` also saves us from the repetition between these function definitions and the option set. Just another advantage of switching over to that approach.

aaschaer: I think ran into some issue using **kwargs on my first iteration of this and switched to inspect as a workaround, given that I can't remember what the issue was and this has gotten much cleaner since then I'll switch back to **kwargs.

aaschaer:
Changed inspect to kwargs as requested.

I'd still like to work a little on the endpoint create endpoint_type logic. Do we definitely want to keep GCP as the default? I would prefer requiring one of --shared, --globus-connect-server or --globus-connect-personal (maybe shorten those to "server" and "personal") to prevent confusion as to what kind of endpoint a user was making.

sirosen:
> I would prefer requiring one of --shared, --globus-connect-server or --globus-connect-personal (maybe shorten those to "server" and "personal") to prevent confusion as to what kind of endpoint a user was making.

That sounds much better than the current behavior. We've heard the complaint that defaulting to `personal` is confusing / non-obvious.

Do you want to do that in this PR as well?

---

With `inspect` replaced by `**kwargs` usage, I think this is looking pretty good. We can hold off for a bit to add requirement of an explicit endpoint type, or merge as-is.

aaschaer:
This PR already has too much in it, I'll make another PR for endpoint type selection.

Issue #242, opened 2017-03-21
Add `globus endpoint activate`, accepting MyProxy credentials


sirosen:
Supersedes #150 

I think `globus endpoint is-activated` is still useful/good. Maybe we can discuss the future fates of these commands post-GlobusWorld and see if some kind of merging is reasonable.
Meantimes, we need to resurrect an old command: `globus endpoint activate`
This has PM signoff, so we should go ahead and implement.

Proposed Usage:
```
globus endpoint activate ENDPOINT_ID \
  -U/--myproxy-username USERNAME \
  -P/--myproxy-password PASSWORD
```

- Should attempt autoactivation if no credentials are given
- Only needs to support autoactivation and myproxy activation (for now)
- Helptext should note that commandline creds are for "MyProxy", not "OAuth for MyProxy"
- `-U` requires `-P` and vice-versa
- Optional: allow `globus config set cli.default_myproxy_username USERNAME` to set a default for `-U`

PR #245, opened 2017-03-22
endpoint activate

aaschaer:
First pass at addressing issue #242 

Upon manual testing on GCS everything appears to be working, but we've run into the "how to automate tests with passwords" issue again.

I went ahead and added web activation since that's far easier to do than myproxy anyways.

I wasn't sure if we should always attempt autoactivation or not. It could make a user's life easier, but also cause confusion if a user was expecting manual activation. Maybe always attempting autoactivation unless a --no-autoactivate flag is passed would fix this problem?

I considered making the password a hidden prompt input akin to the click.password_option, but felt this might make it seem like we are being more secure with these passwords than we really are. If a false sense of security (along with the disclaimer about how myproxy activation handles passwords I assume we have somewhere) is okay I can add this.

### DIFF
--- a/globus_cli/commands/config/init.py
+++ b/globus_cli/commands/config/init.py
@@ -4,7 +4,8 @@ import click
 
 from globus_cli.safeio import safeprint
 from globus_cli.parsing import CaseInsensitiveChoice, common_options
-from globus_cli.config import write_option, OUTPUT_FORMAT_OPTNAME
+from globus_cli.config import (
+    write_option, OUTPUT_FORMAT_OPTNAME, MYPROXY_USERNAME_OPTNAME)
 
 
 @click.command('init',
@@ -14,7 +15,9 @@ from globus_cli.config import write_option, OUTPUT_FORMAT_OPTNAME
 @click.option('--default-output-format',
               help='The default format for the CLI to use when printing.',
               type=CaseInsensitiveChoice(['json', 'text']))
-def init_command(default_output_format):
+@click.option("--default-myproxy-username",
+              help="The default username to use when activating via myproxy.")
+def init_command(default_output_format, default_myproxy_username):
     """
     Executor for `globus config init`
     """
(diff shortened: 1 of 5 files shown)


aaschaer:
Manual testing on GCS worked just fine, but some of the messages might be confusing to non API fluent users, notably "The endpoint could not be auto activated, fill in the returned activation_requirements and POST them back to /activate to perform manual activation." requires more API fluency than we should ever require anyone using the CLI. 

Should we overwrite that one specifically or stop relying on the message field all together?

sirosen:
> I considered making the password a hidden prompt input akin to the click.password_option, but felt this might make it seem like we are being more secure with these passwords than we really are. If a false sense of security (along with the disclaimer about how myproxy activation handles passwords I assume we have somewhere) is okay I can add this.

I think we should make it a hidden prompt. No one who's serious about security will be using this without some sense of what it's doing.
There's a risk that someone who knows basically nothing -- perhaps isn't even activating a MyProxy endpoint and is just confused -- will see it and think we're bad at our jobs.

> I wasn't sure if we should always attempt autoactivation or not. It could make a user's life easier, but also cause confusion if a user was expecting manual activation. Maybe always attempting autoactivation unless a --no-autoactivate flag is passed would fix this problem?

That sounds excellent as a solution. I think the WebApp policy is to always try autoactivation before anything else happens, so we can follow that lead.

> Should we overwrite that one specifically or stop relying on the message field all together?

Long-long term, we'll probably want a collection of custom error messages mapped from Transfer errors, falling back on `message` when we don't have a good error.
For now, I think coughing up an inscrutable string on occasion is acceptable.

---

Quality-wise, I think this is fine, and it covers all of the "hard parts".

I'd like us to try autoactivation and make `--myproxy-password` a hidden prompt.

aaschaer:
Needed a little bit of redesign to have passwords be hidden inputs.
Users can now use endpoint activate <UUID> --myproxy which will prompt them for their myproxy username/password. They can skip the prompts by passing these values with --myproxy-username and --myproxy-password (which I made a hidden option but mentioned after a warning in help) or set a default myproxy username in config.

Autoactivation is now always attempted unless --no-autoactivate is given, and in the process I added a check if the endpoint is already active which can be bypassed with the --force option.

sirosen:
I'm not sure I 100% agree with requiring `--myproxy` if `-U/--myproxy-username` or `-P/--myproxy-password` is given and making `-P/--myproxy-password` hidden. However, I also can't say that I definitively disagree with those choices.

Given that `-P` can prompt, and `-U` can have a default in config, `--myproxy` makes sense  to include as an option.

`--web` wasn't something I thought necessary, but it's a nice bell on this.

PR #246, opened 2017-03-22
Local server login improvements

jaswilli:
* Adds a post-login attempt landing page that either displays the error or a link to the CLI docs.
* Handle the case where a user may deny consent.

### DIFF
--- a/globus_cli/helpers/local_server.py
+++ b/globus_cli/helpers/local_server.py
@@ -1,6 +1,5 @@
 import logging
 import os
-import ssl
 import threading
 
 try:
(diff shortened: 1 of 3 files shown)


globus_cli/helpers/local_server.py
```diff
@@ -40,38 +90,49 @@ def is_remote_session():
...
+                        'ascii'))
+                self.server.return_code(code)
+            else:
+                msg = query_params.get(
+                    'error_description', query_params.get('error'))
+
+                self.wfile.write(
+                    bytes(
+                        HTML_TEMPLATE.substitute(post_login_message=msg),
+                        'ascii'))
```
sirosen: On Python2, this blows up with `TypeError: str() takes at most 1 argument (2 given)`

Because 2-to-3 and encoding and reasons.

I think you can rely on `six.b` for this, but I'm not 100% sure. Checking that now.

sirosen:
With the py2 issue fixed, I think this is good to merge.

Issue #144, opened 2016-11-23
`task event-list` has a bad text output format


sirosen:
Right now, it tries to table-print its results. However, one of the columns it's printing is an arbitrary JSON blob, so it doesn't fit into the tabular format.
Furthermore, the table printing logic tries to use the maximum length of that field as the width of the column, which makes for a bad header as well.

We should probably do something other than the tabular format.
I can't think of a way of making that table work.

Ideally, it would be consistent in look and feel with the other non-tabular output which we produce.

sirosen:
This could be handled, maybe, by omitting that field from the results. It makes the text output kind of useless though...

sirosen:
@ranantha, is it okay to punt on this for 1.0?

We have no solution ready, and JSON output works fine.
(The output "works", but is so ugly as to be useless)

sirosen:
Discussed. Agreed that this can wait until post 1.0

PR #251, opened 2017-03-27
Replace all printing with new OutputFormatter

sirosen:
The `OutputFormatter` is intended to be a layer at which we control/dispatch on output types, styles, and apply all of the bells and whistles which we find necessary.

It replaces all calls to check the output format, and all direct prints.
Accepts several well-known strings to describe textual output formats, and converters and othersuch for handling inputs which don't match up with assumptions.
Text output can also be completely overwritten with custom strings, decorated with preambles and epilogs, or replaced with custom print functions.

JSON output is pretty simple, but does accept and use a json_converter func which transforms non-dict data into a dict. That lets us convert iterables and othersuch easily.

This is a necessary first stage of preparation for:
- jmsepath
- filtered/selectable table output

I don't know that I personally really believe in that second case, but jmsepath is definitely a real thing I want to add.

---

There's a murky distinction between initializing the formatter and calling it. Which options go where is not obvious to me, but it seems desirable to model this as a class? I'm open to discussion on whether or not this is the right way to go.

There may be simplifications to this which we can make. I added features as they became necessary, not following any grand plan.


sirosen:
Rebased to remove conflict with #253, and updated to fix the failing test (`transfer --dry-run` now respects `--format [text|json]`, so we have to specify `-F json` in the tests).

corpulentcoffee:
I'm OK with the approach overall, but I don't love the class-based interface.

Aside from `__init__`, the class has three public methods, of which only one is actually called publicly (`print_response`). Generally, callers are instantiating the class with some arguments, calling `print_response()` on the new instance with some other arguments, and then discarding it. The `commands.endpoint.is_activated` module is an exception that actually saves the instance and then calls `print_response()` in different ways conditionally.

As a caller to this, I think I would prefer just having `globus_cli.safeio` expose a single callable that combines the arguments that go to `__init__` with those that go to `print_response` (so I don't have to remember which ones go where) along with the various constants (e.g. `FORMAT_TEXT_TABLE`).

sirosen:
Yeah, I wasn't a huge fan of it and thought it was confusing by the time I was done with it. I wasn't certain enough that it needed fixing until seeing this feedback. It's been refactored into a helper, and is somewhat simpler.

This sprawling world of output handling could probably be tidied up even more on the inside, but I'm not sure how right now.

corpulentcoffee:
I didn't dig into why yet, but showing the details of a bookmark breaks in this PR:

```
$ globus bookmark show cc3fd750-e99f-11e6-b9b1-22000b9a448b
ddb59aef-6d04-11e5-ba46-22000b92c6ec:/~/

$ globus bookmark show cc3fd750-e99f-11e6-b9b1-22000b9a448b -v
ValueError: max() arg is an empty sequence
```
This is OK on `master`:

```
~$ globus bookmark show cc3fd750-e99f-11e6-b9b1-22000b9a448b
ddb59aef-6d04-11e5-ba46-22000b92c6ec:/~/

~$ globus bookmark show cc3fd750-e99f-11e6-b9b1-22000b9a448b -v
ID:          cc3fd750-e99f-11e6-b9b1-22000b9a448b
Name:        Home on Globus Tutorial Endpoint 1
Endpoint ID: ddb59aef-6d04-11e5-ba46-22000b92c6ec
Path:        /~/
```

sirosen:
Fixed that, and added a test for `bookmark show -v`. Unfortunately, that required a bit of work for testing bookmarks in general.
The error is something I can audit for in general -- failing to pass `fields` when requesting tabular or "record-like" output. (Aside: I don't love calling it "record", but that's what Steve called it a while back and I have no better name ready-to-hand.)

Ideally, we'd have tests that at least run every command with `-F text`, `-F json`, combined with `-v`, `-vv`, and `-vvv`. Even if I were about to go and write that, I don't want to muddle this PR with it.

sirosen:
> I can see us running into this over and over.

Agreed.

> Should we maybe add logic to the output formatter internals where `fields` that gets defaulted (empty) gets all fields (i.e., `named_fields = [(k, k) for k in res.data]`)?

This was my first instinct, but a moment of thought makes me think it could be wrong.
Not everything in `res.data` is guaranteed to serialize nicely. Imagine that default formatting applied to a list or object field. (more like #144 )

I think I'm going to go with errors if `fields` is absent.

sirosen:
If no fields are supplied and we have a text format which requires them, it's now an error.

I'm checking for other misses and haven't found any so far.

PR #253, opened 2017-03-27
Get identities

aaschaer:
Addresses #172

Default output is one resolved uuid or username per line in the same order as the inputs with "NO_SUCH_IDENTITY" for missing values.

Verbose is now tabular and json is unchanged.

### DIFF
--- a/globus_cli/commands/get_identities.py
+++ b/globus_cli/commands/get_identities.py
@@ -1,34 +1,36 @@
 import base64
 import uuid
-
 import click
 
 from globus_sdk import GlobusResponse
 
+from globus_cli.safeio import safeprint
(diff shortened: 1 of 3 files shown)


sirosen:
This matches up with what we discussed yesterday. Good to see test coverage expanding. Merging.

jaswilli:
Post-merge nitpick: If you're going to do whitespace/quote changes it's nice to do it in a standalone commit, otherwise it's hard to pick out the real changes in the diff.

sirosen:
Yes, that is true. Though I don't mind the occasional 1-2 line change sneaking in, there was a little bit much in this changeset.

Commits are practical artifacts which we use primarily to review changes. So, whatever makes them the most convenient for that purpose works.

PR #254, opened 2017-03-30
Support for JMESPath expressions via --jmespath

sirosen:
Possible now that #251 is merged.

This option takes a JMESPath expression, coerces the output format to JSON, and post-processes the output results with the given expression. Compilation happens at option callback evaluation time, so no calls are made if we're given a malformed expression.

Adds dependency on jmespath==0.9.2 (latest release)

Includes a couple of simple tests, nothing too fancy.

### DIFF
--- a/globus_cli/parsing/command_state.py
+++ b/globus_cli/parsing/command_state.py
@@ -1,5 +1,6 @@
 import warnings
 import click
+import jmespath
 
 from globus_cli import config
 from globus_cli.parsing.case_insensitive_choice import CaseInsensitiveChoice
@@ -16,6 +17,8 @@ class CommandState(object):
     def __init__(self):
         # default is config value, or TEXT if it's not set
         self.output_format = config.get_output_format() or TEXT_FORMAT
+        # a jmespath expression to process on the json output
+        self.jmespath_expr = None
         # default is always False
         self.debug = False
         # default is 0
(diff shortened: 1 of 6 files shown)


globus_cli/parsing/command_state.py
```diff
@@ -36,17 +39,35 @@ def is_verbose(self):
...
+        state.jmespath_expr = jmespath.compile(value)
+        return state.jmespath_expr
+
+    f = click.option(
         '-F', '--format',
         type=CaseInsensitiveChoice([JSON_FORMAT, TEXT_FORMAT]),
         help='Output format for stdout. Defaults to text',
         expose_value=False, callback=callback)(f)
+    f = click.option(
+        "--jmespath",
```
jaswilli: How about `--query, -q`?

sirosen: I don't really mind `--query`, but I worried that it might have bad optics on commands like `globus endpoint search` where "query" is a meaningful verb. Is that not a big deal or not as confusing as I worry?

The other thing on my mind with its naming is general skittishness about having a short option for this. JMESPath queries aren't for the faint hearted -- a malformed expression will blow up, sometimes rather opaquely. Should it have a short opt?

jaswilli: I really want the option to be `--jq` but there's the obvious problem with that...

I'd just prefer the option name to be short and easy to type. Typing `--jmespath` over and over doesn't sound fun.

sirosen:
I added the `--jq` alias. Convenience outweighs the slim potential for confusion here, IMO. No one seems to disagree strongly.

PR #255, opened 2017-04-03
Add --deadline option to transfer and delete.

aaschaer:
This is not currently documented in the [API docs](https://docs.globus.org/api/transfer/task_submit/) as far as I can tell, but came up in a support ticket for the SDK and is one less option we're missing from the hosted CLI.


sirosen:
I like the new `task_submission_options` sharing. Looks good.

PR #257, opened 2017-04-05
Revoke any previously existing tokens on login

aaschaer:
While writing login docs I realized that when using --force on login existing tokens were just being overwritten and not revoked.

Also removed some depreciated tokens that had a todo to remove them.

I accidentally revoked the clitester's tokens during a quick manual test (it worked), so there are some new ones here, if we don't want to make this change I'll make a quick PR to fix that.


sirosen:
No objection. I agree that we should revoke on `--force`d login.

PR #272, opened 2017-04-07
Fix task pause-info exit status

### LINKED ISSUE #260: Task pause-info has inconsistent exit codes
Take an unpaused task, e.g. a completed task.

`globus task pause-info TID -Ftext` will exit 1.
`globus task pause-info TID -Fjson` will exit 0.

Either both should exit 0 or both exit 1.

sirosen:
Even if a task is not paused, exit(0) when text printing. Otherwise, exit code differs by output format.
Fixes #260

### DIFF
--- a/globus_cli/commands/task/pause_info.py
+++ b/globus_cli/commands/task/pause_info.py
@@ -1,7 +1,7 @@
 import click
 
 from globus_cli.parsing import common_options, task_id_arg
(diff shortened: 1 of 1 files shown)


aaschaer:
Upon manual testing task pause-info is exiting with 0 as expected on un-paused tasks. Merging.

Issue #177, opened 2017-02-03
config help


tuecke:
The only way I could figure out what the various "config" commands do was to play with it.  The help text wasn't much help.

The "globus config -h" text should probably say that the config file is used to store various config parameters such as access tokens automatically, but that there are some config options that the user can set to configure the default behavior of the CLI.  Then list those user-definable options (e.g., cli.output_format).

It was not clear from the "config init" help that it would interactively prompt me to set user-definable configuration values, namely for output_format.  But is the "config init" even needed?  It seems redundant with "config set", aside from its interactive nature.  But I suppose if there are more user-definable options, that would be useful.

Perhaps have a common example that you use across all of the config help messages.  For example, in "config set" show how to set cli.output_text to json.  In "config remove", show how to remove the cli.output_text.  In "config show", show how to show the cli.output_text.

sirosen:
We haven't paid much attention to these commands as it's not clear whether or not we should keep them.
There's a case to be made that even the `output_format` option isn't necessary, and no other config parameters have been added or surfaced as desirable.

I'll put this and the other `globus config` issues into the `1.0` milestone, but we should take a moment to discuss the future of these commands.

---

> It was not clear from the "config init" help that it would interactively prompt me to set user-definable configuration values, namely for output_format. But is the "config init" even needed? It seems redundant with "config set", aside from its interactive nature. But I suppose if there are more user-definable options, that would be useful.

My desire for that command was based on my own experience with `knife`. The `knife init` command is more or less essential, as there are six or seven options which you must set to even get started with the tool, and without the interactive prompts I can't imagine doing it correctly every time.

The same rules may not apply to us; I'm not sure.

sirosen:
@aaschaer, I'd like you take point on this, since we've had no new insights and we're short on time.

Let's remove `globus config edit`, and try to improve docs on the rest.

I think these commands should say, in their helptext, that they are for advanced users only.
We'll hold off on really changing them significantly for now.

aaschaer:
Sounds good.

PR #273, opened 2017-04-07
Improve config helptext, remove config edit, and change default to cli

aaschaer:
Addresses #177 

In addition to removing edit and trying to improve help text, I also changed the default location from "general" to "cli" since the only values in "general" are depreciated access tokens.

I added a few notes that these commands are for advanced users as suggested.


sirosen:
Those tokens are deprecated too! ;-)

I think the general->cli switch is correct/natural. Good call.
Taking a look at content now.

sirosen:
I have some very minor gripes with the text. Mostly "Globus API" vs "Globus APIs" and other minor edits.

None of them are important enough to block on. Meeting.

PR #276, opened 2017-04-10
fixed some broken links in index

teresas:
links were prefixed with globus_ but filenames were not, so links were broken on index page


aaschaer:
Ah, sorry for missing that. Merging.

PR #286, opened 2017-04-11
Validate tokens on login

sirosen:
Adds calls to `validate_token()` on `globus login`. Long-standing TODO item.

### DIFF
--- a/globus_cli/commands/login.py
+++ b/globus_cli/commands/login.py
@@ -84,7 +84,13 @@ def check_logged_in():
             whoami_email is None or whoami_name is None):
         return False
 
-    # TODO: check that the tokens are valid
+    # check that tokens are valid
+    native_client = internal_auth_client()
+    for tok in (transfer_rt, auth_rt):
+        res = native_client.validate_token(tok)
+        if not res['active']:
+            return False
+
     return True
 
 
--- a/globus_cli/commands/login.py
+++ b/globus_cli/commands/login.py
@@ -87,7 +87,7 @@ def check_logged_in():
     # check that tokens are valid
     native_client = internal_auth_client()
     for tok in (transfer_rt, auth_rt):
-        res = native_client.validate_token(tok)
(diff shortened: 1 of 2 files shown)


globus_cli/commands/login.py
```diff
@@ -84,7 +84,13 @@ def check_logged_in():
             whoami_email is None or whoami_name is None):
         return False
 
-    # TODO: check that the tokens are valid
+    # check that tokens are valid
+    native_client = internal_auth_client()
+    for tok in (transfer_rt, auth_rt):
+        res = native_client.validate_token(tok)
```
jaswilli: The method name is `oauth2_validate_token()`.

tests/unit/test_login_command.py
```diff
@@ -0,0 +1,22 @@
...
+
+from tests.framework.cli_testcase import CliTestCase
+
+from globus_cli.config import AUTH_RT_OPTNAME, TRANSFER_RT_OPTNAME
+
+
+class EndpointCreateTests(CliTestCase):
+    @mock.patch('globus_cli.commands.login.internal_auth_client')
+    def test_login_validates_token(self, get_client):
+        ac = mock.MagicMock()
```
jaswilli: MagicMock strikes again!

sirosen: 😭

sirosen:
@jaswilli, I just pushed a fix for this. Not only did I fix the method name, but I also added a `spec=globus_sdk.NativeAppAuthClient` to ensure that it can't sneak back in.

We should probably be `spec=...`ing more aggressively, because of the problems with Mock.
The Mock docs also point out the fun possibility of `mock.assert_called_once_with(...)` being mispelled as `mock.assret_called_once_wth(...)`, which would pass silently...

jaswilli:
> mispelled as mock.assret_called_once_wth(...), which would pass silently...

Ha. I did notice that one later on as well.

Issue #129, opened 2016-11-01
Add protection on `globus delete` against unsafe globs?


sirosen:
This was discussed on #69 
Deletes through the new CLI are always `--unsafe` per hosted CLI rules.

Should we add client-side protections against trailing globs?

PR #288, opened 2017-04-13
rm command

### LINKED ISSUE #196: rm missing
The ssh CLI has an "rm" command that emulates the Unix rm.  It supports file globbing, and blocks waiting for completion.  The new CLI should have this command also.

aaschaer:
First pass to resolve #196, I'll add some unit tests soon.

It felt a little weird having the task_submission_options on this command since the point of rm is to pretend like a delete task is synchronous, but since the hosted cli has --label and --deadline I figured --dry-run and --submission-id wouldn't hurt?

I left out the -D option from the hosted cli since users can now just use 'globus rm EP:PATH &' if they want to make their synchronous wrapper of an asynchronous task run asynchronously again for some reason.

I decided to add a --timeout that defaulted to a minute, because I dislike the idea of a user submitting a task that becomes inactive or takes longer than expected time and getting stuck. However since the point of rm is blocking I wouldn't mind removing this or making it not a default.

Should I add a --no-globbing option in case people happen to be rming files that have a bunch of * ? [ or ] in their names and don't want to have to escape their characters?

Issue #129 also applies here.

### DIFF
--- a/globus_cli/commands/main.py
+++ b/globus_cli/commands/main.py
@@ -12,6 +12,7 @@ from globus_cli.commands.whoami import whoami_command
 from globus_cli.commands.get_identities import get_identities_command
 from globus_cli.commands.ls import ls_command
 from globus_cli.commands.delete import delete_command
+from globus_cli.commands.rm import rm_command
 from globus_cli.commands.transfer import transfer_command
 from globus_cli.commands.mkdir import mkdir_command
 from globus_cli.commands.rename import rename_command
@@ -40,6 +41,7 @@ main.add_command(ls_command)
 main.add_command(mkdir_command)
 main.add_command(rename_command)
 main.add_command(delete_command)
+main.add_command(rm_command)
 main.add_command(transfer_command)
 
 main.add_command(endpoint_command)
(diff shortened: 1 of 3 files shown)


sirosen:
This has been sitting open for a long time, but at some point we will need to talk about it.
It's generally a bad idea to have multiple ways of performing the same task, and I believe that #196 would be a misfeature.

IMO, `globus task wait $(globus delete $src $dst)` is preferable because it clearly delineates the waiting options from the deletion options, and there's no confusion about which behavior is being controlled by a given option. `globus rm` would have to graft together numerous options from `globus delete` and `globus task wait`, and I'm not sure that's a good idea at all.

Similitude to local shell commands is a mirage which we should not necessarily chase. Already, we have to contend with locally interpreted glob patterns being applied to the target endpoint, and other "brokenness" that we aren't prepared to resolve. Making our commands less obviously unique/dissimilar may be very confusing for users.

sirosen:
Since this is being revived...

@aaschaer, I think this is a viable basis, but needs to be revisited in light of `--unsafe/--star-silent` being added to `globus delete`.
Probably we need to define a common option set for `rm` and `delete`.

There might be a viable refactor I can do.
I'll try to remember to look tomorrow or over the weekend.

I think it should be a fresh PR though; not reopening this one.

PR #295, opened 2017-04-27
whoami use userinfo and add --linked-identities option

### LINKED ISSUE #115: whoami command changes + move to userinfo endpoint
- [ ] change the backing API from the Auth `whoami` endpoint to the future Auth `userinfo` endpoint
     requires auth:view_identity_set scope
- [ ] add support for returning the linked identities (which may require requesting an additional scope for those to be returned via the `userinfo` Auth API endpoint)

aaschaer:
Resolves #115

Probably shouldn't be merged until globus/globus-sdk-python/issues/202 is resolved and the view_identity_set scope is publicly documented, but any changes in the SDK will need to preserve access to the top level fields, so this should continue to work even if a more pythonic way of getting the identity_set is added.

### DIFF
--- a/adoc/whoami.1.adoc
+++ b/adoc/whoami.1.adoc
@@ -10,14 +10,16 @@ globus whoami - Show the currently logged-in identity.
 
 == DESCRIPTION
 
-The *globus whoami* command will display information for the currently logged
-in user.
-
-The default output is just the username, but if --verbose is given
-Name, ID and email are displayed as well.
+The *globus whoami* command will display information for the currently
+logged-in user.
 
 
 == OPTIONS
+
+*--linked-identities* ::
+
+Show the full identity set of the currently logged in identity, which contains
+the primary identity and any identities linked to it.
 
 include::include/format_option.adoc[]
 
@@ -25,12 +27,36 @@ include::include/help_option.adoc[]
 
 include::include/verbose_option.adoc[]
 
+
+== OUTPUT
+
+If no options are given the default output is just the preferred
+username of the logged in identity.
+
+If *--linked-identities* is given the output will be each username in the
+logged-in user's identity set.
+
+If *--verbose* is given, the following fields will be output, either in
+a record format or a table format if *--linked-identities* is also given.
+
+- 'Username'
+- 'Name'
+- 'ID'
+- 'Email'
+
+
 == EXAMPLES
 
 Display multiple fields of the current user's information:
 
 ----
 $ globus whoami -v
+----
+
+Display each username in the current user's identity set:
+
+----
+$ globus whoami --linked-identities
 ----
 
 include::include/exit_status_no_http.adoc[]
--- a/globus_cli/commands/login.py
+++ b/globus_cli/commands/login.py
@@ -1,6 +1,5 @@
 import platform
 import webbrowser
-
 import click
 
 from globus_sdk import AuthClient, AccessTokenAuthorizer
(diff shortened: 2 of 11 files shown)


sirosen:
Especially in light of #354, I've been looking at this again.
I'd like to get it merged at some point, but we need to add a little bit of compatibility glue to keep `globus whoami` sane in cases where the user has already done a `login` with the older scopes.

A crude spec of what we need to add:
- `globus login` should explicitly delete obsoleted config data on success
- if the `whoami` info is present in config, `globus login` should behave as though it got `--force`
- `globus whoami` checks config data and works the old way...
  - unless you gave `--linked-identities`, in which case it fails and demands a fresh login
  - or if there is no config data, pursues the new codepaths

I'd also like input from @jaswilli on what other cases we might need to handle. This was all I could think of.

aaschaer:
Okay updated this PR, it just so happens this also resolves #412

Note this is still technically a breaking change since the underlying json for whoami has changed, so anyone scripting off jmespath might run into unexpected errors.

jaswilli:
This looks good to me. 

Do we want to add something to remove the `WHOAMI_*` fields from the config file?

aaschaer:
I guess the the `WHOAMI_*` fields could cause some confusion, and it shouldn't be too hard to just delete them on config load. Unless someone has strong objections I'll add that in?

sirosen:
I approve of deleting them. It feels like cleaning up after ourselves and being good citizens.
We can drop the deletion code after a while, when we think most people will have updated.

> Note this is still technically a breaking change since the underlying json for whoami has changed, so anyone scripting off jmespath might run into unexpected errors.

I'm not really all that worried -- practically speaking nobody will notice; it's doubtful anyone is currently using this -- but it seems kind of bad in principle.
Can we maybe retain the old out by using the `whoami` endpoint unless we're passed `--linked-identities`?
We could clean it up to all use userinfo in v2, and I think the login/logout code herein remains good.

Is that worth the effort? I know it bulks up `whoami`, but I'm willing to accept that in exchange for "true" (a.k.a. irritating) backwards compatibility.

aaschaer:
So evidently `/p/whoami` will return your full identity set regardless of if you've requested the `view_identity_set` scope.

1. Should I switch to `p/whoami` to simplify the failure on wrong scopes logic?

2. Either way the json blob we get will not match the old whoami values, but I could repackage the primary identity into the old format we were using for whomai for full backwards compatibility?

jaswilli:
Since this is a user command line tool first and foremost I'm not terribly inclined to take a hardline on any sort of semver compatibility. I think modified JSON output from an underlying API call is fair game in this case.

Really, the JSON output could change at any point if an underlying service changes...

sirosen:
I'd rather switch to the publicly documented API if possible.

> I could repackage the primary identity into the old format we were using for whomai for full backwards compatibility?

If this doesn't seem too onerous, let's go with that?

Another option:
Declare that I'm being far too picky and we should just let the JSON output change subtly and we'll field any complaints which arise.

jaswilli:
> let the JSON output change subtly and we'll field any complaints which arise.

👍

aaschaer:
I also would like to avoid having completely different json keys between `globus whoami` and `globus whoami --linked-identities`

I suppose if sessions is going to be v2 we could save this "breaking" change for then?

sirosen:
> Really, the JSON output could change at any point if an underlying service changes...

Yes, our output and compatibility promise is only as strong as the underlying services. Completely agree on this.

The only reason I hesitate at all is that we have this under our control and we're initiating the change, but I'm convinced that it's okay to let this one slide.

Let's just go forward with this as-is, with the subtle change in output.
Looks like it's ready and can merge once tests pass.

PR #296, opened 2017-04-28
Improve autoactivation failure messages

### LINKED ISSUE #30: Produce more useful error on endpoints where autoactivation fails to activate
Right now, the error output is not prescriptive, and just notes that an endpoint needs to be activated.
Example from Steve:

``` shell
$ globus transfer ls --endpoint-id 9d6d994a-6d04-11e5-ba46-22000b92c6ec
HTTP status:      400
request_id:       9sHmaySoX
code:             ClientError.ActivationRequired
message:          The endpoint '9d6d994a-6d04-11e5-ba46-22000b92c6ec' is not activated
```

This is actually the LS call returning the `ActivationRequired` error, not a processed activation requirements doc.

What we really need to do is this: autoactivate any endpoint that we see, and capture any activation requirements document that we get back.
If the response has a success code, proceed as usual.
However, if we get `NotActivated`, we need to parse and present the `ActivationRequirements`.
I think in that case, we should exit with status 1 -- the command failed, after all.

As for the error presentation, it depends a bit on the required activation type.
For web activation, we just need to show the activation link, that's pretty easy.
For CLI activation, we need some text that clarifies that we mean the hosted CLI.
Possibly something like this:

``` shell
$ globus transfer ls --endpoint-id 9d6d994a-6d04-11e5-ba46-22000b92c6ec
Globus CLI Error: A Globus Endpoint Requires Activation
code:             ClientError.ActivationRequired
message:          The endpoint '9d6d994a-6d04-11e5-ba46-22000b92c6ec' is not activated

To activate 9d6d994a-6d04-11e5-ba46-22000b92c6ec , you must use the Hosted Globus CLI.
Follow this guide to connect <link>
Then run
  endpoint-activate 9d6d994a-6d04-11e5-ba46-22000b92c6ec
```

Note that I omit some of the HTTP detail (it's a success there, after all), but still want the `ClientError.ActivationRequired` code noted.

This may have some unforeseen impact on #21, so we need to think about that.

Ideally, the SDK would handle parsing the activation requirements document a bit, and we could inspect an `activation_type` property on it to see CLI vs. Web activation.

aaschaer:
Resolves #30 and improves the autoactivation failure messages in endpoint activate and is-activated as well.

As the issue says it might be nice if some of the activation_requirements parsing were done in the SDK, but it seemed reasonable enough to do it in the CLI for now.

I didn't write any automated tests for these without a stable endpoint that doesn't accept auto-activation to test against, but manual testing against my own GCS endpoint worked fine.

### DIFF
--- a/globus_cli/commands/endpoint/is_activated.py
+++ b/globus_cli/commands/endpoint/is_activated.py
@@ -1,7 +1,8 @@
 import click
 
 from globus_cli.parsing import common_options, endpoint_id_arg
-from globus_cli.services.transfer import get_client
+from globus_cli.services.transfer import (
+    get_client, activation_requirements_help_text)
 from globus_cli.safeio import formatted_print
 
 
@@ -29,14 +30,11 @@ def endpoint_is_activated(endpoint_id, until, absolute_time):
     def fail(deadline=None):
         exp_string = ''
         if deadline is not None:
-            exp_string = ' or will expire before {}'.format(deadline)
+            exp_string = ' or will expire within {} seconds'.format(deadline)
 
-        formatted_print(
-            res, simple_text=(
-                '{0} is not activated{1}\n'
-                'To activate, please go to the following page:\n\n'
-                '  https://www.globus.org/app/endpoints/{0}/activate\n'
-                ).format(endpoint_id, exp_string))
+        message = ("The endpoint is not activated{}.\n\n".format(exp_string) +
+                   activation_requirements_help_text(res, endpoint_id))
+        formatted_print(res, simple_text=message)
         click.get_current_context().exit(1)
 
     def success(msg, *format_params):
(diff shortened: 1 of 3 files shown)


globus_cli/commands/endpoint/is_activated.py
```diff
@@ -29,14 +29,25 @@ def endpoint_is_activated(endpoint_id, until, absolute_time):
...
+
+            "The endpoint supports Web{} activation:".format(
+                " and MyProxy" if myproxy else ""),
+
+            ("Use 'globus endpoint activate {} --web' "
+                "for Web activation.".format(endpoint_id)),
+
+            "{}".format(("Use 'globus endpoint activate {} --myproxy' "
+                         "for Myproxy activation."
+                         .format(endpoint_id)) if myproxy else "")
```
sirosen: You can just use the parenthetical expression, no need to `"{}".format(x)`, since that's the same as `x`.
I can totally see how with all of the tricks and games we play with `format()` this kind of thing sneaks up on you.

aaschaer:
Refactored this to handle delegate proxy.

I currently am displaying delegate proxy as an available activation method for an endpoint even if M2Crypto is not imported, but instead of giving a command I say M2Crypto is required and give a link to the docs. Would it be better to not even display it as an available activation method if MCrypto isn't available?

jaswilli:
Assuming we still want this it looks like it needs to be updated after the switch to `cryptography`.

PR #297, opened 2017-04-28
Add OneUseOption and decorator

### LINKED ISSUE #33: Define a new option type which strictly requires that it is only passed once
We're currently using the Click default behavior when parsing options with `multiple=False`.
That parses all instances of an option and returns the last one given.

So `globus transfer ls --endpoint-id 'abc123' --endpoint-id 'def456'` is entirely valid, and results in `endpoint_id="def456"`.

This is fine/safe for options where allowing an override doesn't change the service-side semantics of commands, like `--format`, but for IDs and paths, it could be the result of user confusion and not result in the desired action.
For example, a `globus transfer async-transfer` with two sources, two destinations, and two paths is inherently ambiguous, and likely means that the user has wholly misunderstood usage of that command.

To correctly implement a change in this behavior, I think we need to add a new class of option.
Such opts can internally be specified with `multiple=True` and then assert that `len(value) <= 1` where `value` is the resulting tuple.
I'd suggest that this be a custom callback, but I worry about future opts that need or want a custom callback _and_ this behavior.
The custom option class doesn't add much more complexity, and properly handles that case.

aaschaer:
Resolves #33 and adds a decorator for easy substitution of existing click.option decorators.

I feel like most of the options this issue was created for are now arguments, but the endpoint create options seemed like a reasonable enough use case for trying this out. If desired I can go add this to any options that probably shouldn't be used more than once.

### DIFF
--- a/globus_cli/commands/endpoint/create.py
+++ b/globus_cli/commands/endpoint/create.py
@@ -1,7 +1,7 @@
 import click
 
 from globus_cli.parsing import (
-    common_options, endpoint_create_and_update_params,
+    common_options, endpoint_create_and_update_params, one_use_option,
     validate_endpoint_create_and_update_params, ENDPOINT_PLUS_REQPATH)
 from globus_cli.services.transfer import (
     autoactivate, get_client, assemble_generic_doc)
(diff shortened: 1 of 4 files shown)


sirosen:
I kept meaning to read this and never did. It looks good, and checks out okay.

> I feel like most of the options this issue was created for are now arguments

That may be, but I think `--shared` is a perfect example of something which is technically constrained to be an option, but which we really want only one of. These cases will still exist, and it's nice to be able to respect the fact that if you're specifying certain options more than once, it's probably a mistake.

Issue #37, opened 2016-10-10
Check if user has authenticated before attempting authenticated calls


sirosen:
This is a bit of an open question.

Right now, if you attempt an authenticated action like `globus transfer ls ...`, you get a 401 error back.
Ideally, for any call that cannot be made unauthenticated, the relevant commands would give you a "no auth" error of some kind and direct you to `globus login`.

The big question here is where and how this gets enforced.
Some calls are available unauthenticated, but I'm not sure that it's worth extra burden to support that mode of usage.
If all calls which talk to the service are forced to be made with credentials, then the problem is solved neatly and easily.
However, if we do want to support this case, then we need to figure out where credentials are and are not required.

When HTTPS endpoints enter the ecosystem with possible "public read" ACLs, the desire for unauthenticated calls to be supported _may_ (not necessarily _will_) increase.

aaschaer:
Was #299 sufficient to close this?

PR #299, opened 2017-05-01
Improve missing / invalid token error handling

aaschaer:
Addresses #37, although rather than explicitly checking that a user is authenticated per call that requires authentication, I figured it would be more maintainable and require fewer network calls to just have specific exception handling for when we get back errors from missing or invalid tokens?

Unable to make automated tests for invalid fresh tokens, but manual testing worked as expected.

### DIFF
--- a/globus_cli/safeio/errors.py
+++ b/globus_cli/safeio/errors.py
@@ -35,8 +35,7 @@ class PrintableErrorField(object):
                 name, spacer, spacer.join(self.value.split('\n')))
 
 
-def write_error_info(error_name, fields):
-    message = None
+def write_error_info(error_name, fields, message=None):
     if outformat_is_json():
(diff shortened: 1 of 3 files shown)


globus_cli/safeio/errors.py
```diff
@@ -35,17 +35,18 @@ def __str__(self):
...
             dict(
                 [('error_name', error_name)] +
                 [(f.name, f.value) for f in fields]),
             indent=2)
-    else:
-        message = 'A {0} Occurred.\n{1}'.format(
+
+    if not message:
+        message = 'A{0} {1} Occurred.\n{2}'.format(
+            "n" if error_name[0] in "aeiouAEIOU" else "",
```
sirosen: Hahaha! I would probably have just tried to avoid the indefinite article, but this works too.

PR #304, opened 2017-05-09
Upgrade pip in Makefile

jaswilli:
Ensure we have an up-to-date version of pip.

**update:** I guess we also need to update `setuptools`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,14 +8,14 @@ python:
   - "3.6"
   # allow failures on CPython dev and pypy
   # we want to be warned about these, but they aen't critical
-  - "3.6-dev"
+  - "3.7-dev"
   - "pypy"
   - "pypy3"
 matrix:
   allow_failures:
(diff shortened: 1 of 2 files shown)


sirosen:
Annoying thing about Travis: even when in the py3 parts of the build matrix, `virtualenv` still sets up a py2 virtualenv. That's why the travisfile wasn't using `make test/opts`.

You can confirm which python is being used by looking at which site-packages folder is used (inside of `.../python2.7/...`). I would _love it_ if you can figure out a way to have the Makefile be Travis-friendly, but that blew up in my face when I last tried.

jaswilli:
New plan based on all the various constraints we've run into:

* New make target, `travis` which will not use a virtualenv.
* Install the latest pip and setuptools, except for in the `travis` target, where we install the latest version less than 30, since we need to support the version of pypy3 that Travis makes available.

PR #306, opened 2017-05-10
Add --delegate-proxy option to endpoint activate

### LINKED ISSUE #303: Add delegate_proxy activation command; potentially an optional behavior of `endpoint activate`
This has been presented by PM as high-priority work. globus/globus-sdk-python#182 has some of the original thoughts on this topic.

Specifics about the implementation, that I think would be good for a first pass:
- Add an install extra to the CLI by modifying `setup.py` -- you can see `globus-sdk[jwt]` for an example -- `globus-cli[delegate-proxy]`, which installs `m2crypto==0.26.0`
- The `delegate_proxy` behavior should only even _appear_ if `m2crypto` is installed; do this via `try: ... / except ImportError: ...`
- Don't make any changes to the SDK, but factor out helpers operating purely on the activation requirements document, so they can be moved "upstream" easily in the future if desired
- Like the `python-jose` dependent SDK tests, have the tests runnable via `make tests` or `make tests/opts` and only the latter installs `globus-cli[delegate-proxy]`

@jaswilli, @corpulentcoffee I'd like one of you to review the resultant work, along with me.

aaschaer:
Resolves #303 

Testing this was a bit of an ordeal.

First of all, it seems M2Crypto and pip don't get along on Ubuntu, so I was only able to confirm that optional dependency logic in setup.py / Makefile break on pip install. Fortunatly Ubuntu is aware of this issue and has a package `python-m2crypto` that can be installed with apt-get. This breaks our install via virtualenv model though...

Second of all, default GCS doesn't work with delegate_proxy activation. You need to use a Gridmap instead of the default MyProxyGridmapCallout for authorization. If anyone else wants to test this feel free to ask me for more details.

I did manage to manually activate and then ls an endpoint though, so this *is* working as far as I can tell. I will try to add some automated tests tomorrow, but I suspect that I won't be able to do much but confirm that the option is hidden if M2Crypto hasn't been imported unless I find a way around the pip incompatibility.


jaswilli:
Curious about the m2crypto issue, as I/we have installed it via pip before. Let's talk tomorrow.

aaschaer:
Well I'm happy enough with the code itself, but getting M2Crypto to work with Travis and Nose2 continues to be an issue. It seems to currently be in a state where pip claims that it has "Successfully installed M2Crypto-0.26.0" (other than the pypy builds, which fail on building the M2Crypto wheel) but the tests are skipping on M2Crypto import failures.

sirosen:
Do local runs of the testsuite work okay? I might consider punting on getting the Travis issues worked out until later, if we all can test locally.

Reviewing code changes now.

adoc/endpoint_activate.1.adoc
```diff
@@ -33,17 +35,31 @@ Myproxy activation requires your username and password for the myproxy
...
 cli.default_myproxy_username".
-
 For security it is recommended that you only enter your password when
 prompted to hide your inputs and keep your password out of your command
 history, but you may pass your password with the --myproxy-password
 or -P options.
 
+NOTE: Delegate Proxy activation requires the optional dependency of
+M2Crypto. This dependency is optional as it frequently requires additional
+non-python packets to be installed on your machine. Consult
```
sirosen: Typo alert: "packets"

aaschaer:
Added a `--proxy-lifetime` option (also hidden unless M2Crypto imported) to configure the lifetime of the proxy being generated and set the default lifetime to 12 hours.

Changed "packets" to "packages" to fix my typo.

And yes, the tests run fine locally. When I run `make travis` in a virtualenv on my machine everything passes and the test that checks that `--delegate-proxy` is hidden if M2Crypto is not imported is skipped as expected.

adoc/endpoint_activate.1.adoc
```diff
@@ -33,17 +35,31 @@ Myproxy activation requires your username and password for the myproxy
...
 or -P options.
 
+NOTE: Delegate Proxy activation requires the optional dependency of
+M2Crypto. This dependency is optional as it frequently requires additional
+non-python packages to be installed on your machine. Consult 
+https://gitlab.com/m2crypto/m2crypto/blob/master/INSTALL.rst[M2Crypto's Install Page]
+for details
+
+To use Delegate Proxy activation use the --delegate-proxy option with a
+file containing an 5.X09 certificate as an argument (e.g. an X.509
```
sirosen: Typo: `5.X09`

aaschaer:
Pushed the copy-edits if we want to go ahead and merge this to get the release out. I should have the other edits done before too long though.

PR #312, opened 2017-05-23
Switch from M2Crypto to cryptography for delegate proxy activation

aaschaer:
Hopefully this will make our dependencies less painful.

This should address #311 since cryptography uses one function to parse all private key types, and I switched to using a regex to parse the PEM files to make use of this. Haven't tested yet, and don't know what would happen if someone gave us a non RSA private key.

Includes some security updates, the old code was using SHA1, and X.509 v2 (yet still managed to add extensions somehow?) new code is using SHA256 and X.509 v3

I added a sentence about `pip install globus_cli[delegate-proxy]` to the activation adoc, do we want to add this to our install page as well?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ travis:
 	pip install --upgrade "setuptools>=29,<30"
 	pip install -r test-requirements.txt
 ifeq (, $(findstring pypy, $(TRAVIS_PYTHON_VERSION)))
-	pip install "M2Crypto==0.26.0"
+	pip install "cryptography>=1.8.1,<2.0.0"
 endif
 	python setup.py develop
 	flake8
--- a/adoc/endpoint_activate.1.adoc
+++ b/adoc/endpoint_activate.1.adoc
@@ -45,10 +45,12 @@ history, but you may pass your password with the --myproxy-password
 or -P options.
 
 NOTE: Delegate Proxy activation requires the optional dependency of
-M2Crypto. This dependency is optional as it frequently requires additional
-non-python packages to be installed on your machine. Consult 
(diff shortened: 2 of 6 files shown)


aaschaer:
Looks like we still have some travis pypy issues.

Cryptography is only compatible with pypy >= 2.6, and travis's  default versions are 2.5 for pypy and 2.4 for pypy3. Travis does allow specifying pypy versions, but doing so requires sudo, and even after activating that my builds were breaking on the pypy install.

I did manage to get travis half-working by only turning on sudo and not specifying a pypy version, but it seems only pypy updated automatically, where as pypy3 was still stuck at 2.4.

I decided the simplest solution was to keep skipping cryptography on pypy runs, but if someone who's better at travis than me wants to get an updated version working, delegate proxy *should* work with pypy >= 2.6 now.

jaswilli:
Fully on board with just continuing to skip pypy until we can reasonably get current releases on travis.

sirosen:
This all looks good to me, barring one small typo ("Gloubus" in an adoc file).

I'm going to test it out on my own machine and see if I can get it to work before merging.

aaschaer:
Fixed and squahsed the typo (And added Globus to my spell-check dictionary)

sirosen:
Works for me, clearly works for everyone.

Thanks for fixing up the typo. ( :laughing: about you adding it to your dictionary. I've come to ignore spellcheck on anything technical. My spellchecker doesn't even recognize "spellcheck" as a word!)

Like Jason, I'm cool with skipping crypto tests on pypy. If we can fix it in the future, that would be great, but no harm leaving it as-is.

PR #313, opened 2017-05-30
Improve utf-8 handling and add encoding tests

### LINKED ISSUE #65: Verify that we're handling encoding (utf8) correctly
Comment from Karl:

> What is the byte/unicode encoding on mkdir/rename/transfer/ls paths.  What if a path is illegal utf8.

aaschaer:
Resolves #65

Click is very good at handling byte string/unicode conversion issues, but recommends against importing unicode literals from future in modules that import click (and gives a warning if you do so). I doubt unicode literals would break anything since our APIs only use unicode for text, but I figured better to be safe and used the `u""` marking to make any potentially unsafe format strings unicode.

The only places that needed additional changes to handle non-ascii strings were the test framework and the `PrintableErrorField` which I re-wrote a little to `safe_stringify` inputs and not use `__str__` to make sure everything was unicode internally.

The tests are mostly re-written versions of the SDK encoding tests, but I added some additional error handling tests and a test on escaping quotation marks.

I also did some manual testing with ubunutu terminal using `en_US.UTF-8` encoding to make sure the click test runner wasn't playing significantly nicer than a real shell.

### DIFF
--- a/globus_cli/commands/bookmark/helpers.py
+++ b/globus_cli/commands/bookmark/helpers.py
@@ -31,8 +31,8 @@ def resolve_id_or_name(client, bookmark_id_or_name):
                        if bookmark_row['name'] == bookmark_id_or_name)
 
         except StopIteration:
-            safeprint('No bookmark found for "{}"'.format(bookmark_id_or_name),
-                      write_to_stderr=True)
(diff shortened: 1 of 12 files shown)


sirosen:
Took a while to review this. Awesome!

PR #316, opened 2017-06-12
Work around concurrent test runs

jaswilli:
- Drop fields from the tested output that can be affected
  by concurrent test runs.


sirosen:
You could also use `assertDictEqual()`, but I don't care much either way.

Waiting for tests to pass, then merging.

jaswilli:
Oh yeah, sorry. On autopilot 😄

PR #317, opened 2017-06-12
v1.1.1

jaswilli:
* Replace `M2Crypto` with `cryptography` (delegate proxy support).
* Bump SDK to v1.1.1

@ranantha We're going to hold this release pending possibly communication to our existing delegate proxy users about the switch.

### DIFF
--- a/changelog.adoc
+++ b/changelog.adoc
@@ -1,6 +1,23 @@
 = CLI Changelog
 
 
+== 1.1.1
+
+Enhancements:
+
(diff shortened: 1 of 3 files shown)


ranantha:
I can reach out to the groups who had expressed interest in this. But let’s update the documentation for this here: https://docs.globus.org/cli/reference/endpoint_activate/ <https://docs.globus.org/cli/reference/endpoint_activate/>. We can update the note to say when you update to v1.1.1 cryptography will be used and link to that page, and leave the existing note for older versions. 

Rachana

PR #320, opened 2017-06-13
Make `globus login` respect Ctrl+C

### LINKED ISSUE #319: `globus login` incorrectly captures KeyboardInterrupt in fancy browser mode
Easy to reproduce:

1. `globus login` to start the login flow with the browser interactivity.
2. Go back to the shell where `globus login` is running
3. Ctrl+C

Nothing happens.
The login command should die upon receipt of an interrupt.

I can workaround this by `Ctrl+Z; jobs; kill %n`, but that's silly.

sirosen:
The behavior of Python's multithreaded Queue.get() is different when block=True/False. Apparently, Queue.get(block=True) cannot be interrupted because it's waiting on a mutex.
Instead, using a polling interval (*any* polling interval) forces Python to use interruptible primitives. Everything else nicely falls into place.

Fixes #319

### DIFF
--- a/globus_cli/helpers/local_server.py
+++ b/globus_cli/helpers/local_server.py
@@ -139,7 +139,17 @@ class RedirectHTTPServer(HTTPServer, object):
         self._auth_code_queue.put_nowait(code)
 
     def wait_for_code(self):
-        return self._auth_code_queue.get(block=True)
+        # use a non-blocking get() on the queue so that keyboard interrupts
+        # during the wait will be respected
+        # use a 1 second polling interval, though the exact value doesn't
+        # matter
+        # relevant Python issue discussing this behavior:
+        # https://bugs.python.org/issue1360
+        while True:
+            try:
+                return self._auth_code_queue.get(timeout=1)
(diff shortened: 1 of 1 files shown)


globus_cli/helpers/local_server.py
```diff
@@ -139,7 +139,17 @@ def return_code(self, code):
...
+        # during the wait will be respected
+        # use a 1 second polling interval, though the exact value doesn't
+        # matter
+        # relevant Python issue discussing this behavior:
+        # https://bugs.python.org/issue1360
+        while True:
+            try:
+                return self._auth_code_queue.get(timeout=1)
+            except Queue.Empty:
+                pass
```
jaswilli: I just tested this using block+timeout and the workaround behavior is still in effect. Doing it that way we could just do a simple:
`return self._auth_code_queue.get(block=True, timeout=3600)`

and then future maintainers wouldn't have to figure out what we were up to...

jaswilli:
Nice work @sirosen ⭐️

PR #323, opened 2017-06-14
Change 'login to' to 'log into'

### LINKED ISSUE #322: Typo: "Login to" should be "Log into"
minor typo in the CLI "login    Login to Globus to get credentials for the Globus CLI" ... "Login to" should be "Log into". It's pretty minor though, some people are adamant that login is not a verb because 'loginned' is not a word.

aaschaer:
Quick regex to fix #322


sirosen:
Sure, whatevs.

PR #327, opened 2017-06-15
Support provisioning flag for get-identities calls

### LINKED ISSUE #326: Creating an endpoint ACL/role or doing an endpoint search by owner using an identity username that doesn't exist in Auth yet raises an IndexError
If a CLI user tries `globus endpoint role create --identity`, `globus endpoint permission create --identity`, or `globus endpoint search --filter-owner-id` and they supply a username that doesn't exist in Auth yet, `_lookup_identity_field()` will raise an `IndexError` when calling `client.get_identities()` because the `provision=True` flag isn't passed to the SDK.

I think that probably

- CLI codebase callers of `maybe_lookup_identity_id()` need to indicate whether to provision
- `endpoint endpoint search` shouldn't provision
- `endpoint endpoint permission` and `globus endpoint role` should either provision *or* expose a flag for the CLI user to indicate that they want to provision an identity should the one that they passed not exist

### LINKED ISSUE #328: Error when roles or permission list contains an entry that has a non-existent identity
`globus endpoint role list <uuid>` and `globus endpoint permission list <uuid>` assume that every entry in the list will resolve to an identity so the `username` can be printed. However, Transfer allows ACLs and roles to be assigned to arbitrary UUIDs.

Example:
screen shot 2017-06-19 at 9 31 05 am

CLI output:
```
globus endpoint role list 78ff1934-4638-11e7-bd4f-22000b9a448b
IndexError: list index out of range
```

Will fix as part of #327.

jaswilli:
Closes #326
Closes #328 

* Adds option `--provision-identity` to create endpoint role and permission commands, mutually exclusive with the `--identity` option.
* Fixes error condition during a list operation where Transfer has a role or permission "principal" that does not yet exist in Auth (Transfer allows roles/permissions to be created for arbitrary UUIDs).

### DIFF
--- a/globus_cli/commands/endpoint/permission/create.py
+++ b/globus_cli/commands/endpoint/permission/create.py
@@ -13,7 +13,8 @@ from globus_cli.services.transfer import get_client, assemble_generic_doc
 @click.command('create', help=('Create an access control rule, allowing new '
                                'permissions'))
 @common_options
-@security_principal_opts(allow_anonymous=True, allow_all_authenticated=True)
+@security_principal_opts(
+    allow_anonymous=True, allow_all_authenticated=True, allow_provision=True)
 @click.option('--permissions', required=True,
               type=CaseInsensitiveChoice(('r', 'rw')),
               help='Permissions to add. Read-Only or Read/Write')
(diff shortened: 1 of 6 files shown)


jaswilli:
This is ready for review/test/merge.

corpulentcoffee:
I'm happy with this. Is it okay for merging @sirosen?

sirosen:
Just finished my review. Yeah, looks great.
Thanks @jaswilli for taking care of this.

PR #332, opened 2017-06-26
Mock config file for testing

aaschaer:
As discussed, mocks out the whole config layer by patching calls to `get_config_obj`.

This should make it more difficult to accidentally revoke the testing tokens, and adds the ability to specify config per call of run_line for further testing of login/logout later.

### DIFF
--- a/tests/framework/cli_testcase.py
+++ b/tests/framework/cli_testcase.py
@@ -58,6 +58,10 @@ def default_test_config(*args, **kwargs):
     """
     user_data = get_user_data()["clitester1a"]
 
+    # create a ConfgObj from a dict of testing constants. a ConfigObj created
+    # this way will not be tied to a config file on disk, meaning that
+    # ConfigObj.filename = None and ConfigObj.write() returns a string without
+    # writing anything to disk.
     return ConfigObj({"cli": {
         AUTH_RT_OPTNAME: CLITESTER1A_AUTH_RT,
         AUTH_AT_OPTNAME: "",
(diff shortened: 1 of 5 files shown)


tests/framework/cli_testcase.py
```diff
@@ -46,12 +51,14 @@ def clean_sharing():
...
+def default_test_config(*args, **kwargs):
     """
-    Writes known constants and refresh tokens into config for testing
+    Returns a ConfigObj with the clitester's refresh tokens and whoami info
+    as if the clitester was logged in and a call to get_config_obj was made.
     """
     user_data = get_user_data()["clitester1a"]
-    testing_constants = {
+
+    return ConfigObj({"cli": {
```
sirosen: Can you add a comment here? The behavior of `ConfigObj`, when initialized this way, is kind of unintuitive.

Specifically, explain that:
- You can create a `ConfigObj` from a `dict`, which sets `ConfigObj.filename = None`
- When `ConfigObj.filename = None`, then `ConfigObj.write()` returns a string and writes nothing to disk

I had to go to the configobj docs site to know that this was safe.

That said, this behavior seems perfect for testing.

tests/unit/test_login_command.py
```diff
@@ -5,20 +5,22 @@
...
 import globus_sdk
 
-from tests.framework.cli_testcase import CliTestCase
+from tests.framework.cli_testcase import CliTestCase, default_test_config
 
 from globus_cli.config import AUTH_RT_OPTNAME, TRANSFER_RT_OPTNAME
 
 
-class EndpointCreateTests(CliTestCase):
+class LoginCommandTests(CliTestCase):
```
sirosen: Nice catch. :sweat_smile:

sirosen:
I dropped some inline comments. If you add the comments requested, and tests pass, I'll merge.

PR #333, opened 2017-06-27
Rate limited BFS for --recursive ls

### LINKED ISSUE #46: Recursive dir listing has potentially dangerous implications service-side -- limit it or remove it
The case we're concerned about is the pathological case of hundreds of nested directories expanding to huge numbers of files (m/billions).
Per notes from Karl, we'd need to switch traversal to be breadth-first and unbuffered output to handle that case at all -- makes perfect sense, though unbuffered JSON output will require careful handling and some testing to ensure that the results are valid JSON.

To resolve, we have three paths forward (there may be more?):
- Remove recursive `ls` from the CLI
- Add it to the SDK and limit the total number of dirs to `ls` to some "small" but reasonable number -- e.x. 50
- Add it to the SDK and enforce an N second sleep every M dirs, where N and M are both small (guessing 1s every 25 dirs)

This is only worth discussing as an SDK addition (and not purely as a CLI feature) because we've made it very easy to build your own version of recursive `ls`.
Unless we make it easier to use the "correct" version than to build your own, we haven't actually solved the problem.

Removing it from the CLI is viable but may be worse than adding a `recursive=True` option to the SDK `ls` -- I just don't trust that people won't go building this themselves. It therefore seems a poor way of protecting the service.

Given the choice between a hard limit and a sleep, the sleep has much better semantics.
It doesn't mess with error modes at all -- an SDKLimitExceeded error will be confusing to people, and is likely to break scripts unexpectedly as their inputs scale up.
The sleep behavior also sets the tone correctly for a future universe in which that kind of rate limiting is enforced service side in addition to being done client-side.

There's already been some discussion of these, but I'm moving it into issues to better track status and decisions.
I'd appreciate input from @karlito-go and @ranantha on this one.

aaschaer:
Ports https://github.com/globus/globus-sdk-python/pull/214 into the CLI as suggested on that PR.

This should be replaced whenever Transfer eventually supports native recursive ls on endpoints, but for now this should resolve #46

### DIFF
--- a/globus_cli/commands/ls.py
+++ b/globus_cli/commands/ls.py
@@ -1,5 +1,4 @@
 import click
-import re
 
 from globus_cli.parsing import common_options, ENDPOINT_PLUS_OPTPATH
 from globus_cli.safeio import formatted_print
@@ -50,41 +49,15 @@ def ls_command(endpoint_plus_path, recursive_depth_limit,
     else:
         res = client.operation_ls(endpoint_id, **ls_params)
 
-    def formated_item_name(item):
-        # determine if the formated item name should end in / or not
-        final_slash = ("/" if item["type"] == "dir" else "")
-
-        # ignore any trailing slashes for easier logic
-        if path and path[-1] == "/":
-            formatted_path = path[:-1]
-        else:
-            formatted_path = path
-
-        # if we are doing a recursive ls we want to display the relative path
-        # from the start of the ls.
-        if recursive:
-
-            # if the ls command has a starting path,
-            # give everything relative to that path
-            if formatted_path:
-                relative_path = re.match(u".*{}/(.*{})".format(
-                    re.escape(formatted_path), re.escape(item["name"])),
-                    item["path"]).group(1)
-            # otherwise give everything after the first /~/ or /
-            else:
-                relative_path = re.match(u"(/~)*/(.*{})".format(
-                    re.escape(item["name"])), item["path"]).group(2)
-
-            return relative_path + final_slash
-        else:
-            return item["name"] + final_slash
+    def cleaned_item_name(item):
+        return item['name'] + ('/' if item['type'] == 'dir' else '')
 
     # and then print it, per formatting rules
     formatted_print(
         res, fields=[('Permissions', 'permissions'), ('User', 'user'),
                      ('Group', 'group'), ('Size', 'size'),
                      ('Last Modified', 'last_modified'), ('File Type', 'type'),
-                     ('Filename', formated_item_name)],
+                     ('Filename', cleaned_item_name)],
         simple_text=(None if long or is_verbose() or outformat_is_json() else
-                     "\n".join(formated_item_name(x) for x in res)),
+                     "\n".join(cleaned_item_name(x) for x in res)),
         json_converter=iterable_response_to_dict)
--- a/globus_cli/commands/ls.py
+++ b/globus_cli/commands/ls.py
@@ -60,19 +60,20 @@ def ls_command(endpoint_plus_path, recursive_depth_limit,
         else:
             formatted_path = path
 
-        # if the item has a path field, it is recursive, and we want
-        # to display the relative path from the start of the ls.
-        if item.get("path"):
+        # if we are doing a recursive ls we want to display the relative path
+        # from the start of the ls.
+        if recursive:
 
             # if the ls command has a starting path,
             # give everything relative to that path
             if formatted_path:
                 relative_path = re.match(u".*{}/(.*{})".format(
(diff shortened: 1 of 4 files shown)


globus_cli/commands/ls.py
```diff
@@ -101,19 +38,52 @@ def ls_command(endpoint_plus_path, recursive_depth_limit,
...
+
+        # if the item has a path field, it is recursive, and we want
+        # to display the relative path from the start of the ls.
+        if item.get("path"):
+
+            # if the ls command has a starting path,
+            # give everything relative to that path
+            if formatted_path:
+                relative_path = re.match(u".*{}/(.*{})".format(
+                    formatted_path, item["name"]), item["path"]).group(1)
```
sirosen: I think you need to pass these values through `re.escape`. Otherwise, this will fail on paths with things like literal `?` chars.

The same goes for the expression below, in its use of `item["name"]`.

globus_cli/commands/ls.py
```diff
@@ -101,19 +38,52 @@ def ls_command(endpoint_plus_path, recursive_depth_limit,
...
+
+        # ignore any trailing slashes for easier logic
+        if path and path[-1] == "/":
+            formatted_path = path[:-1]
+        else:
+            formatted_path = path
+
+        # if the item has a path field, it is recursive, and we want
+        # to display the relative path from the start of the ls.
+        if item.get("path"):
```
sirosen: This is a little bit nitpicky, but while we're making changes:
It would be cleaner if you did this detection based on the value of `recursive`. There's no rule that Transfer couldn't add an attribute which happens to also be named `path` in the future, and it reads a bit better/more obviously.

globus_cli/commands/ls.py
sirosen: Do we need the leading `.*` in the expression, actually? I'm not comfortable merging something that munges the output like this unless I know what it's doing.

As an alternate form of this question:
What's wrong with `relative_path = item["path"][len(formatted_name)+1:]`?

aaschaer:
Made suggested changes, thanks for the review!

globus_cli/commands/ls.py
```diff
@@ -101,19 +38,53 @@ def ls_command(endpoint_plus_path, recursive_depth_limit,
...
+            # if the ls command has a starting path,
+            # give everything relative to that path
+            if formatted_path:
+                relative_path = re.match(u".*{}/(.*{})".format(
+                    re.escape(formatted_path), re.escape(item["name"])),
+                    item["path"]).group(1)
+            # otherwise give everything after the first /~/ or /
+            else:
+                relative_path = re.match(u"(/~)*/(.*{})".format(
+                    re.escape(item["name"])), item["path"]).group(2)
```
sirosen: I'm looking at these regexes again, and I'm not yet convinced that they're right/safe.

Questions:
- Why start the recursive case with `.*`? What could appear before `formatted_path`, and why can it not, itself, contain multiple matches for `formatted_path`?
- What is the `.*` before `item["name"]` in the recursive case? Isn't the formal case for this that there's nothing? When does that match anything?
- In the non recursive case, why do any regex alteration at all?
- How will this handle S3 (today) and Google Drive (future) paths which contain literal `~` chars?

sirosen:
I was able to create some files on a test S3 endpoint to prove out that the path munging is not safe, and, equally importantly, that what we have today works.

Create the following files on an S3 endpoint:
- `/~`
- `/~/\/~` (`\` is a dir name, since S3 is a funky place)
- `/~/~/~/~`

---

Old Version of `ls -r /` | New Version of `ls -r /`
---|---
`~/` | `~/`
`~` | `~`
`~/\/` | `\/`
`~/~/` | `~/`
`~/\/~` | `\/~`
`~/~/~/` | `~/`
`~/~/~/~` | `~`

Though the old `ls -r` output isn't pretty (how could it be in this case?) it's correct.

---

Closely related to this is a question of whether `globus ls` should behave more like `ls` or like `find`. The `find`-style output is easier on us, as maintainers, as we don't have to munge the output very much at all.

@jaswilli, @corpulentcoffee, do you have strong opinions about our `ls` command being `find`-like vs. `ls`-like? (As in, would producing output like `find` offend your sensibilities?)

aaschaer:
I originally used regexes to minimize changes to the BFS implementation, but since this is just for the CLI now  I see no reason to remove regexes, add the relative path to the BFS queue, and go back to updating the `name` field to the relative path we want to display.

sirosen:
I can confirm that the new version works on my pathological-case, and it is a lot more predictable.

Without knowing what the future will hold (e.g. when will we even have recursive `ls` in the API?), building this in the easiest way for the CLI seems the best course of action. Merging.

PR #334, opened 2017-06-30
Update Travis builds to run on Trusty

sirosen:
Per [a recent update](https://blog.travis-ci.com/2017-06-19-trusty-updates-2017-Q2) Travis' new Ubuntu 14.04 infrastructure is rolling out.
It's very unclear from their blog entries whether or not this applies to our builds (`sudo: false`), but I'm guessing that _something_ changed.

See if this builds correctly, and if so, let's switch over?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 sudo: false
+dist: trusty
+
 language: python
 python:
   - "2.7"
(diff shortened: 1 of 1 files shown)


sirosen:
Looks like this fails to build on `pypy` and `pypy3`, both with `sudo: required` and `sudo: false`.
Not really sure what's going on at this point, but I guess we could go with `sudo: required` and `apt-get install pypy` to get some testing in on _some_ pypy interpreter?

I think I'd sooner drop the pypy builds -- they're only there because they're easy, and give us confidence that we'll work on _any_ random python version.

jaswilli:
Yeah, I'd be fine with dropping pypy at least until it gets better support on Travis or until we hear a report of someone having trouble with pypy.

PR #335, opened 2017-07-06
Managed endpoint options

### LINKED ISSUE #134: Add `--(no-)managed-endpoint` flag or equivalent to `endpoint update`
Eric Blau raised this as an issue, and we should be sure to track it somewhere.

Currently, XSEDE is relying on the hosted CLI's
```shell
$ endpoint-modify <endpoint name> --managed-endpoint
```

This is the kind of functionality that seems reasonable to add to `endpoint update`, perhaps under a new name, but there is no API support.
If we can get an API (and corresponding SDK) resource to use, we should add this.

I'm partial to `--managed/--no-managed` as the flag, since `-endpoint` seems redundant.

aaschaer:
Resolves #134 as we are no longer blocked as of transfer 5.6

Adds `--managed/--no-managed` and un-hides `--subscription-id` from endpoint create and update.
I made`--managed/--no-managed` and `--subscription-id` mutually exclusive as that seemed the least likely to cause confusion.

`--managed` passes the magic "DEFAULT" value as the subscription id as discussed on slack.

`--no-managed` passes `None` which required a workaround since `assemble_generic_doc` removed all fields with a value of `None`. We may want to allow passing None to remove other fields as well?

Unit tests aren't feasible without giving public access to a subscription ID, but all of my manual tests worked as expected.

### DIFF
--- a/adoc/include/endpoint_options.adoc
+++ b/adoc/include/endpoint_options.adoc
@@ -45,7 +45,7 @@ Set this endpoint to be (in)visible to other users searching for endpoints.
 (Globus Connect Server endpoints only, all Globus Connect Personal endpoints 
 are private and all shared endpoints are public)
 
-*--myproxy-dn* 'DOMANIN_NAME'::
+*--myproxy-dn* 'DOMAIN_NAME'::
 
 Set the MyProxy Server domain name for this endpoint.
 (Globus Connect Server endpoints only)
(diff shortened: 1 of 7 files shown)


sirosen:
Jason and I agreed that we don't really like the `include_nones` workaround. I just pushed a change with a different, arguably (or at least, in my opinion) cleaner solution to the `None` vs. `null` issue.

I need to do some (re)testing of this, but it basically looks good.

sirosen:
I also did some minor fiddling to allow `--managed --subscription-id FOO` (which is harmless) and disallow `--no-managed --subscription-id FOO` (which is not harmless).

@jaswilli is going to review, and we're going to try to ship this out soon.

Issue #341, opened 2017-08-18
Expand adoc description of batch transfer mode + `--recursive` flag


sirosen:
This relates a little to globusonline/docs.globus.org#180

Someone mentioned in a ticket that he found the "Using the CLI" doc inconsistent with the reference.
If my guess is right, this is because we're not sufficiently clear in the reference documentation that command-line `--recursive` and batch-mode `--recursive` are different and mutually exclusive.

We may want to further clarify with a one or two examples on the reference page, as in
```shell
$ globus transfer --batch ...
xyz def
--recursive a b
c d --recursive
```

aaschaer:
Upon starting to clarify this behavior in the adoc I realized that a similar clarification for the `--symlink` option is  also needed, so I'm now clarifying both in #343

corpulentcoffee:
Closing this per #379.

PR #343, opened 2017-08-18
CLI Symlink support

### LINKED ISSUE #321: Symlink Support
Here is my proposal for how we should add Transfer's upcoming support of symlinks to the CLI:

**Task Submission** 

Add a `--recursive-symlinks [keep | copy | ignore]` option to `globus transfer` that gets passed through the SDK as the `recursive_symlinks` query parameter for Transfer.

Add a `--symlink` option to `globus transfer` with usage similar to `--recursive` on both single item and batch transfers to mark that a `transfer_symlink_item` should be used instead of a `transfer_item` to transfer a symlink rather than the target of a symlink. If both `--symlink` and `--recursive` are given for an item, the CLI should throw a usage error.

**Task Monitoring:**

Add `Symlinks` and `Recursive Symlinks` to the `TRANSFER_FIELDS` for output used by `globus task show`
that will display the new `symlinks` and `recursive_symlinks` fields in the task document.

**Directory Listing:**

For `--long` output of `globus ls`, change the behavior of the File Type and Filename output fields to also look at the `link_target` response field. If `link_target` is not None, display "`name` -> `link_target`" as the Filename, and append " (symlink)" to `type` for File Type output if the `type` is not `invalid_symlink`.

e.g. a valid symlink pointing to a file at /tmp/file.txt would have a File Type of `file (symlink)` and a Filename of `test.txt -> /tmp/file.txt`.

For `--recursive` usage of `globus ls` add a `--follow-symlinks` flag that will also follow valid symlinks to directories while making recursive calls to `operation_ls`. We may want to deal with #46 before we do this.

**Symlink Creation**

Add a `globus symlink` command with usage `globus symlink [OPTIONS] ENDPOINT_ID TARGET_PATH LINK_PATH`
that just passes `ENDPOINT_ID`, `TARGET_PATH`, and `LINK_PATH` to the SDK's new `TransferClient.operation_symlink` method.

aaschaer:
Resolves #321 as discussed in the issue.

@ranantha I made the changes you requested and added examples in the adoc files if you would like to review them.

### DIFF
--- a/adoc/transfer.1.adoc
+++ b/adoc/transfer.1.adoc
@@ -24,12 +24,12 @@ consistent, deterministic location.  The contents of a source directory will
 be placed inside the dest directory.  A source file will be copied to
 the dest file path, which must not be an existing  directory.  All
 intermediate / parent directories on the dest will be automatically
-created if they don't exist. 
+created if they don't exist.
 
-If the files or directories given as input are symbolic links, they are
-followed.  However, no other symbolic links are followed and no symbolic links
-are ever created on the dest.  
-
+By default, if the file or directory at SOURCE_PATH is a symbolic link, it is
+followed, and no other symbolic links are followed, transfered, or created on
+the destination endpoint. This behavior can be changed with the '--symlink' and
+'--recursive-symlinks' options.
 
 include::include/cli_autoactivate.adoc[]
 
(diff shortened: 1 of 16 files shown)


aaschaer:
Since clarifying the command / batch distinction for the `--recursive` option issue from #341 also applies to the `--symlink option` I have updated transfer.1.adoc here to clarify the issue for both options.

adoc/transfer.1.adoc
```diff
@@ -24,56 +24,86 @@ consistent, deterministic location.  The contents of a source directory will
...
 
 *-r, --recursive*::
 
 If SOURCE_PATH and DEST_PATH are both directories, do a recursive transfer
 of all files and directories inside of the source directory to the destination
-directory. Note that for batch input, each directory must be marked as
-recursive individually unlike 'globus delete'.
+directory. For batch input each directory must use the '--recursive' option at
+the batch level, as this option is mutually exclusive with
+'--batch' at the command level. This option is also mutually
```
sirosen: I think you mean "mutually exclusive with `--recursive`"?

The growing potential for confusion here makes me wonder if we should split `transfer --batch` and `transfer srcep:src dstep:dst` into entirely separate commands.

sirosen:
I had one point in the docs that I think is confusing/needs clarification.
Let me know when you get to it, and we can merge.

Let's do a release soon, to get this into people's hands.

ranantha:
@sirosen there are still some issues being ironed out with support for symlinks in Globus Connect Server. Please hold off on releasing this until that is sorted out, and the CLI has been tested with the updated server.

sirosen:
I'm cleaning up stale PRs on a few projects. I don't think this ever landed in a state in which we could ship the relevant CLI change, so I'm closing.

Feel free to ping me or reopen if I'm mistaken or things change.

PR #354, opened 2017-10-30
Set encoding=utf-8 on ConfigObj

sirosen:
This is used for loading and writing `~/.globus.cfg`
Setting the encoding allows ConfigObj to correctly handle unicode strings and does not conflict with existing file contents in ascii-only files.
Tested manually -- it's not clear at present how this should be added to the testsuite, since tests currently run against _real_ configuration data.

### DIFF
--- a/globus_cli/config.py
+++ b/globus_cli/config.py
@@ -84,7 +84,7 @@ def get_config_obj(system=False):
     else:
         path = os.path.expanduser("~/.globus.cfg")
 
-    return ConfigObj(path)
+    return ConfigObj(path, encoding='utf-8')
 
 
 def lookup_option(option, section='cli', environment=None):
--- a/globus_cli/safeio/output_formatter.py
+++ b/globus_cli/safeio/output_formatter.py
@@ -1,3 +1,5 @@
+from __future__ import unicode_literals
+
 import json
 import six
 


sirosen:
Added a fix for output formatting, which handles `globus whoami -v` on py2.

Example output:
```
$ chardet ~/.globus.cfg
/home/sirosen/.globus.cfg: utf-8 with confidence 0.99

$ globus whoami -v
Username: sirosen7@globusid.org
Name:     Стивэн Rosen
ID:       1e5a510b-59f9-46d9-893d-347da4ef4179
Email:    sirosen+20171030@globus.org

$ globus whoami -F json
{
  "Username": "sirosen7@globusid.org", 
  "Name": "\u0421\u0442\u0438\u0432\u044d\u043d Rosen", 
  "Email": "sirosen+20171030@globus.org", 
  "ID": "1e5a510b-59f9-46d9-893d-347da4ef4179"
}
```

There may be additional printing commands which are vulnerable to unicode explosions, but this is not trying to solve all of them -- it's purely focused on the fullname issue.

---

Important note about all of this: there's an argument for explicitly encoding in UTF-8 before writing to disk and decoding upon loading data, but _that's not what configobj is doing_.
It's producing a UTF-8 file, which other tools could choke on. However, given that `configparser` in the SDK doesn't mind this, I'm not sure that we care.

PR #356, opened 2017-10-31
Fix testsuite 400s

sirosen:
Transfer didn't used to validate the format of some servername fields, but it has started to. Put the hostnames all under `example.com` to pass validation.

Also, the use of a `try-finally` to fix linting was a misreading of the prior logic. `finally` runs even if the `try` succeeds, of course, but `self._clean()` should only run if things fail -- `tearDown` handles running it otherwise.

Fixing these two issues _should_ get us a passing testsuite.


aaschaer:
Looks good to me, and tests are passing, merging.

PR #360, opened 2017-11-07
Cert validity timestamps now use UTC

FreifunkerEZ:
If you live in GMT-plus-something land (Europe or Asia) the certs generated by create_proxy_cert() will not be valid for some hours because it inserts the local time. But GMT+N is for UTC "The Future". Having a "not yet valid" cert sucks really hard. Now it will use UTC to generate the cert and everything is rainbows.

### DIFF
--- a/globus_cli/helpers/delegate_proxy.py
+++ b/globus_cli/helpers/delegate_proxy.py
@@ -147,9 +147,9 @@ def create_proxy_cert(loaded_cert, loaded_private_key,
     builder = builder.serial_number(serial)
 
     # set the new proxy as valid from now until lifetime_hours have passed
-    builder = builder.not_valid_before(datetime.datetime.today())
+    builder = builder.not_valid_before(datetime.datetime.utcnow())
     builder = builder.not_valid_after(
-        datetime.datetime.today() + datetime.timedelta(hours=lifetime_hours))
+        datetime.datetime.utcnow() + datetime.timedelta(hours=lifetime_hours))
 
     # set the public key of the new proxy to the given public key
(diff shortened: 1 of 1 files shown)


sirosen:
Good catch, thank you!

Since Globus is mostly US-based, we typically don't see the bad behavior. There may be other bugs of this flavor -- incorrectly treating localtime like UTC time -- I'll make a note for us to investigate.

Tests have passed, so I'm merging.

FreifunkerEZ:
Wow, that was fast. Thanks. I feel very honored :)
Well, living in GMT-minus-land, you will not see "not valid yet" but you should be having problems with "not valid anymore" because the whole window-of-validity moves a bit into the past, moving the end-date closer.

I thought I'd grep the whole source and it seems like this is the only place to use datetime
```
 >grep -r datetime * 
 helpers/delegate_proxy.py:import datetime

```
but at least one use of "time" (in parsing/iso_time.py)
Cheers!
chris :)

sirosen:
> Wow, that was fast.

Well, don't worry. Watch how long it takes for us to actually release the fix. 😉 

> Well, living in GMT-minus-land, you will not see "not valid yet" but you should be having problems with "not valid anymore" because the whole window-of-validity moves a bit into the past, moving the end-date closer.

Yes, we'll see earlier than expected expiry. If we're generating certificates that are valid for more than 5 or 6 hours in a US TZ nothing will be "obviously" broken -- you'll just need to get a new cert more often.

Thanks for doing that grep. `globus_cli.parsing.iso_time` looks fine -- no danger of timezone shenanigans there.

PR #361, opened 2017-11-07
Avoid user confusion around --network-use=custom

corpulentcoffee:
Submissions around the endpoint network performance parameters are tricky and can lead to user confusion if we do not validate that the set of parameters used will actually be accepted/persisted by the Transfer service.

cc @ranantha 
cc @karlito-go

For updates, I'm not passing the old `network_use` value (like is done with `managed`) as the Transfer service doesn't do patchy updates on `network_use` -- if you want to set the `network_use` fields, you need to do them altogether -- if we wanted to make this so the user could specify fewer parameters, we could do so, it would probably just mean some munging in the update function before this validation one gets called.

### DIFF
--- a/globus_cli/parsing/shared_options.py
+++ b/globus_cli/parsing/shared_options.py
@@ -188,23 +188,28 @@ def endpoint_create_and_update_params(*args, **kwargs):
             "--network-use", default=None,
             type=click.Choice(["normal", "minimal", "aggressive", "custom"]),
             help=("Set the endpoint's network use level. If using custom, "
-                  "the endpoint's max and preferred concurrency must be set "
+                  "the endpoint's max and preferred concurrency and "
+                  "parallelism must be set "
                   "(Managed endpoints only) (Globus Connect Server only)"))(f)
         f = click.option(
             "--max-concurrency", type=int, default=None,
-            help=("Set the endpoint's max concurrency "
+            help=("Set the endpoint's max concurrency; "
+                  "requires --network-use=custom "
                   "(Managed endpoints only) (Globus Connect Server only)"))(f)
         f = click.option(
             "--preferred-concurrency", type=int, default=None,
-            help=("Set the endpoint's preferred concurrency "
+            help=("Set the endpoint's preferred concurrency; "
+                  "requires --network-use=custom "
                   "(Managed endpoints only) (Globus Connect Server only)"))(f)
         f = click.option(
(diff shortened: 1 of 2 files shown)


globus_cli/parsing/shared_options.py
```diff
@@ -244,6 +244,33 @@ def validate_endpoint_create_and_update_params(endpoint_type, managed, params):
...
+    # because the Transfer service doesn't do network use level updates in a
+    # patchy way, *both* endpoint `POST`s *and* `PUT`s must either use
+    # - `network_use='custom'` with *every* other parameter specified, or
+    # - a preset/absent `network_use` with *no* other parameter specified (in
+    #   this case, Transfer will accept but ignore the others parameters if
+    #   given, leading to user confusion if we don't do this validation check)
+    custom_network_use_params = ('max_concurrency', 'preferred_concurrency',
+                                 'max_parallelism', 'preferred_parallelism')
+    if params['network_use'] == 'custom':
+        for option in custom_network_use_params:
```
karlito-go: This part of the "if" is already checked by transfer-api.  It will fail if custom doesn't pass all parameters.  Probably not harmful if you check it yourself, but hosted CLI does not.

globus_cli/parsing/shared_options.py
```diff
@@ -244,6 +244,33 @@ def validate_endpoint_create_and_update_params(endpoint_type, managed, params):
...
+            if params[option] is None:
+                raise click.UsageError(
+                    "When using --network-use=custom, you must specify all "
+                    "{} parameters: {}".
+                    format(len(custom_network_use_params),
+                           ", ".join("--" + option.replace("_", "-")
+                                     for option in custom_network_use_params))
+                )
+    else:
+        for option in custom_network_use_params:
```
karlito-go: this looks fine.  The hosted CLI does this same sanity check.

globus_cli/parsing/shared_options.py
corpulentcoffee: Given our recent re-learned lessons about duplicating validation client-side and service-side, I am inclined to agree. Unless @sirosen wants me to keep this for some reason, I will drop this part and just run the check below `if params['network_use'] != 'custom'`.

sirosen: If the API rejects this usage, then yes, I agree with both of you that we should let the command through and allow for the API error.

PR #367, opened 2017-12-07
Add --skip-activation-check option

### LINKED ISSUE #352: Add `--skip-activation-check` option for Task submission
Transfer now supports skipping the check that the endpoints used by a task are activated, to allow post-facto activation.
It needs to stay optional to avoid accidentally queuing tasks for inactive endpoints without warning people.

aaschaer:
Resolves #352 

This functionality can be tested automatically without re-visiting the current inability of the test suite to test against any non auto-activatable endpoints, but manual tests worked as expected.

### DIFF
--- a/globus_cli/commands/delete.py
+++ b/globus_cli/commands/delete.py
@@ -28,7 +28,8 @@ from globus_cli.services.transfer import get_client, autoactivate
 @click.argument('endpoint_plus_path', metavar=ENDPOINT_PLUS_OPTPATH.metavar,
                 type=ENDPOINT_PLUS_OPTPATH)
 def delete_command(batch, ignore_missing, recursive, endpoint_plus_path,
-                   label, submission_id, dry_run, deadline):
+                   label, submission_id, dry_run, deadline,
+                   skip_activation_check):
     """
     Executor for `globus delete`
     """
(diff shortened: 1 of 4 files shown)


globus_cli/commands/delete.py
```diff
@@ -38,14 +39,16 @@ def delete_command(batch, ignore_missing, recursive, endpoint_plus_path,
             'delete requires either a PATH OR --batch')
 
     client = get_client()
-    autoactivate(client, endpoint_id, if_expires_in=60)
+    autoactivate(client, endpoint_id, if_expires_in=60,
+                 skip_failure=skip_activation_check)
```
jaswilli: What's the rationale behind still doing autoactivation if skip activation is set?

aaschaer: I was thinking "just in case the endpoint can be autoactivated, why not try?"

The answer to my question is "because it takes another round trip on the network, and no-one will use this on autoactivatable endpoints"

I'll take this out.

jaswilli: I'm not super concerned with the extra request, I was more curious if there was still some reason we needed to call autoactivate.

But yeah, I think dropping it if we can is the way to go. It's confusing if nothing else.

PR #375, opened 2018-02-23
Update `endpoint update` helptext

### LINKED ISSUE #339: Help text for endpoint update is ambiguous
The  help text for ```globus endpoint update``` says:

<pre>
--subscription-id UUID            Set the endpoint as a managed endpoint with
                                  the given subscription ID. Mutually
                                  exclusive with --managed/--no-managed.
--managed / --no-managed          (Un)set the endpoint as a managed endpoint.
                                  Requires the user to be a subscription
                                  manager. If the user has multiple
                                  subscription IDs, --subscription-id must be
                                  used instead of --managed to specify one.
</pre>
It should break out the ```--no-managed``` option separately since, for those users with multiple subscription IDs, you cannot specify a subscription ID if you want to unset the managed flag on an endpoint.

sirosen:
Resolves #339
Almost the same as the text in that issue, but don't state that `--managed/--no-managed` are exclusive -- they take on last-one-wins semantics in this usage. Making them exlusive doesn't seem important.

### DIFF
--- a/globus_cli/parsing/shared_options.py
+++ b/globus_cli/parsing/shared_options.py
@@ -175,12 +175,14 @@ def endpoint_create_and_update_params(*args, **kwargs):
         # Managed Endpoint options
         f = click.option(
             "--managed", "managed", is_flag=True, flag_value=True,
+            default=None,
             help=("Set the endpoint as a managed endpoint. Requires the "
                   "user to be a subscription manager. If the user has "
                   "multiple subscription IDs, --subscription-id must be used "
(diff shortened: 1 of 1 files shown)


sirosen:
I thought I had tested exactly that case, but I guess not.
Will take a look and work out a fix.

sirosen:
Put together a simple fix: my refactor changed the default from `None` to `False`, so just change it back (don't change any downstream logic).
After more careful testing this time around, it seems good.

PR #376, opened 2018-02-24
Add config file path command

### LINKED ISSUE #178: config filename
Somewhere in the config commands there should be an easy way to get back the config filename.  Its printed as part of the output message on various commands, but I couldn't find a command to just get back the name.  Options:

globus config --filename
globus config filename
globus config show --filename

The output should simply be the file name, with no other messages around it.

jaswilli:
Closes #178


sirosen:
Works for me. I don't see any issues with this, since the original behavior of the helper is preserved.

PR #378, opened 2018-02-26
Add logout link text

### LINKED ISSUE #340: `globus logout` doesn't appear to link to the auth.globus.org logout page
Unless I'm missing something, `globus logout` fails to provide a link to `auth.globus.org/v2/web/logout`.

Am I forgetting some subtlety or reason that we did this, or is it just a bug?

sirosen:
Closes #340 

Per our chat, does this text look good?

### DIFF
--- a/globus_cli/commands/logout.py
+++ b/globus_cli/commands/logout.py
@@ -26,6 +26,10 @@ Manage Consents Page:
 
 _LOGOUT_EPILOG = """\
 You are now successfully logged out of the Globus CLI.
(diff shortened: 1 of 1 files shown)


corpulentcoffee:
This isn't `GLOBUS_SDK_ENVIRONMENT`-aware, which sometimes confuses people when they're doing testing outside of production and don't realize that certain links have been statically set.

Would it be helpful for the base `AuthClient` in the SDK to support an `oauth_get_logout_url()` (or `get_logout_url()`) method?

PR #382, opened 2018-02-27
Add globus endpoint local-id command

### LINKED ISSUE #371: Globus Connect Personal UUID
GCP has a file containing the UUID of its endpoint. Unix installs it goes in `~/.globusonline/lta/client-id.txt` and on Windows it goes in `%LocalAppData%\Globus Connect\client-id.txt`.

So we do have the ability to provide a command to show the UUID to a user which I think we certainly want to do.  The question is really, should this just be implemented directly in the CLI or should we also add SDK methods?

jaswilli:
Closes #371 

I don't feel strongly about the name of the command. Suggestions welcome.

```
(venv) jason-mbp:globus-cli jtw$ globus endpoint local-gcp-uuid
No Globus Connect Personal installation found.
```

```
(venv) jason-mbp:globus-cli jtw$ globus endpoint local-gcp-uuid
62a5c95a-1b5b-11e8-b6eb-0ac6873fc732
```

### DIFF
--- a/globus_cli/commands/endpoint/commands.py
+++ b/globus_cli/commands/endpoint/commands.py
@@ -14,6 +14,7 @@ from globus_cli.commands.endpoint.is_activated import endpoint_is_activated
 from globus_cli.commands.endpoint.deactivate import endpoint_deactivate
 from globus_cli.commands.endpoint.my_shared_endpoint_list import (
     my_shared_endpoint_list)
+from globus_cli.commands.endpoint.local_id import local_id
 
 
 @globus_group(name='endpoint', help='Manage Globus endpoint definitions')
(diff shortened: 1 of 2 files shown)


sirosen:
The content we had in our doc was literally
```
`globus local-endpoint-id --personal` ? (Allows for --server later)
```

I kind of like the idea of `local-endpoint-id` with `--personal` as the default.

sirosen:
Per our chat, I favor `globus endpoint local-id --personal` / `globus endpoint local-id`.

If you switch to that I'm happy -- I'd rather have one command for both GCP and GCS down the road.

jaswilli:
Updated with above `globus endpoint local-id --personal` with `--personal` being the default if no flag is specified (so you can use the command as `globus endpoint local-id`.

Issue #344, opened 2017-08-28
Enable parallelism and concurrency tuning on transfer submission


sirosen:
The hosted CLI allows managed endpoint transfers to be submitted with customized performance options. Specifically, you can tune parallelism and concurrency on a per-transfer basis.

This has no current API support, so it's a parity issue blocked on Transfer being updated.

PR #383, opened 2018-02-27
Add --notify and performance task options

### LINKED ISSUE #368: Add `--notify` option to Transfer Task submission
The Transfer API supports notification settings whenever a Task is submitted:
`notify_on_succeeded`, `notify_on_failed`, and `notify_on_inactive`

Exposing this in the CLI could go down a number of routes, but here's one option:
`--notify off`: no notifications, all set to False
`--notify on`: (default) all set to True
`--notify "succeeded,failed"`: don't notify when tasks become inactive
`--notify failed`: notify on when tasks fail
`--notify "off,succeeded"`: UsageError
`--notify "succeeded,failed,inactive"`: same as default, but explicitly set these to True (vs. omitting from API doc)

I don't want to get too fancy about having options which are, themselves, parsed (vs. `--notify` being an option where we allow multiple usage), but `--notify off --notify failed` seems inherently ambiguous. Not allowing multiple uses and following `click`'s typical "last use wins" logic 

Relates to globusonline/koa#62

sirosen:
`--notify` is for both Transfer and Delete Tasks.

`--perf-cc`, `--perf-pp`, `--perf-p`, and `--perf-udt` are all available as hidden options on `globus transfer`, since we don't really want to encourage their usage.

Per #368, `--notify` allows `--notify off`, `--notify on`, `--notify failed,inactive` all doing relatively sensible things. `--notify on,failed` is valid (but takes the stronger, `on`) and `--notify off,failed` is invalid.

Closes #368, #344


jaswilli:
Looks like this might require some test adjustments.

sirosen:
Fixed in cb71d95b06c5686b5b20bf5572cfbb161b17a9ca .
I had written the typical thing for option callbacks of
```
if not value:
    return None
```
but the correct thing in this case is more subtle and explicit
```
if value is None:
    return {}
```

PR #387, opened 2018-03-07
Improvements and fixes to completion

### LINKED ISSUE #386: Make shell completion supported, change recommended sourcing method
cc @ranantha, I'd like us to make the current completion support an officially supported part of the CLI for GlobusWorld. I think that would be a nice time to add it in; it's a bit of a showy feature which has been done for a while and therefore requires basically no additional work.

The code has been sitting almost entirely untouched (and working) for over a year.
I've been using the current shell completion code for a while now and haven't had any issues with it. It works really well, especially in ZSH where the completion support is a bit superior.

A lot of tools use a nice trick for getting their shell completion stuff into people's bashrc:
`mytool --completion` prints the completer, and users are told to `eval "$(mytool --completion)"`

I like this. Could we add it, maybe move shell completion into the packaged CLI source dir?
We could then document this as a full feature of the CLI.

One thing that we'd need to add, I think, is some way to determine which shell to use. I think `bash` is a sane default, so:
`globus --completion` can do bash completion, and
`globus --completion zsh` can do `zsh` completion

@corpulentcoffee, @jaswilli, I know you guys have been less keen on shell completion than I. Does the change I'm suggesting (putting the current completion scripts into a hidden `--completion` option and documenting it) sound acceptable?

sirosen:
This doesn't quite resolve #386 , as that requires a full release and documentation updates, but this is the main body of code for it.

- Add descriptions of flags and commands to zsh completion.
  zsh and other newer, hipper shells support "help strings" of some kind attached to completion results. For now, we just have zsh support, because I care.
  This is cool -- it makes our completion look very showy and professional in zsh.

- Fix bash completion handling of `globus --format=text endp<tab>`
  Switching from `COMP_WORDS` to `COMP_LINE` basically solves this -- by default, bash splits `=` incorrectly and breaks completion.

- Add `Parameter.secondary_opts` to completion choices.
  Click sorts out `--foo/--no-foo` flag declarations into `Parameter.opts` and `Parameter.secondary_opts`, so we need to collect both sets of options for completion.

- Replace completer scripts with `eval "$(...)"` pattern of completion usage. Convert current scripts to shims.

One unfortunate thing: I wanted to have `globus --completer` and `globus --completer zsh`, but this doesn't work that well. Instead, we have `--completer` and `--zsh-completer`, since that gives us distinct options (and therefore no parsing conflict). `--bash-completer` is an alias of `--completer`.

### DIFF
--- a/globus_cli/parsing/shell_completion.py
+++ b/globus_cli/parsing/shell_completion.py
@@ -77,6 +77,13 @@ def get_completion_context(args):
 
 
 def get_all_choices(completed_args, cur, quoted):
+    """
+    This is the main completion function.
+    Inputs:
+    - completed_args: a list of already-completed arguments
+    - cur: the current "word in progress" or None
+    - quoted: is cur part of a quoted string?
+    """
     ctx = get_completion_context(completed_args)
     if not ctx:
         return []
@@ -122,12 +129,13 @@ def get_all_choices(completed_args, cur, quoted):
             # skip hidden options
             if isinstance(param, HiddenOption):
                 continue
-            for opt in param.opts:
-                # only add long-opts, never short opts to completion, unless
-                # the cur appears to be a short opt already
-                if opt.startswith('--') or (
-                        len(cur) > 1 and cur[1] != '-'):
-                    choices.append((opt, param.help))
+            for optset in (param.opts, param.secondary_opts):
+                for opt in optset:
+                    # only add long-opts, never short opts to completion,
+                    # unless the cur appears to be a short opt already
+                    if opt.startswith('--') or (
+                            len(cur) > 1 and cur[1] != '-'):
+                        choices.append((opt, param.help))
     # and if it's a multicommand we see, get the list of subcommands
     elif isinstance(ctx.command, click.MultiCommand) and not quoted:
         choices = [(cmdname, ctx.command.get_command(ctx, cmdname).short_help)
--- a/globus_cli/parsing/shell_completion.py
+++ b/globus_cli/parsing/shell_completion.py
@@ -6,7 +6,7 @@ from globus_cli.safeio import safeprint
 from globus_cli.parsing.hidden_option import HiddenOption
 from globus_cli.parsing.case_insensitive_choice import CaseInsensitiveChoice
 
-SUPPORTED_SHELLS = ['BASH', 'ZSH']
+SUPPORTED_SHELLS = ('BASH', 'ZSH')
 
 
 def safe_split_line(inputline):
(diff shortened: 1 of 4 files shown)


globus_cli/parsing/shell_completion.py
```diff
@@ -168,12 +186,71 @@ def do_zsh_complete():
...
+        """
+        Replace ' with '"'"'
+
+        Because we'll put these single quote chars in '...'
+        quotation, we need to do
+        '   -- end single quotes
+        "'" -- single quote string (will concatenate in ZSH)
+        '   -- start single quotes again
+        """
+        return helpstr.replace("'", "'\"'\"'")
```
corpulentcoffee: I think *something* needs to be done here for backticks in help strings as well; right now, zsh tries to expand them when you return to it the `_arguments` structure.

The help strings used by the various options to `globus ls` contain a lot of backticks, so `globus ls --<TAB>` spits out errors like

```
(eval):.:1: not enough arguments
(eval):1: command not found: --recursive
```

along with backtick-expanded help output like

```
--long                         -- For text output only. Do long form output, kind of like total 84\ndrwxr-xr-x  4 mayor users  409
```

If we had any help strings with `$`, then that would be similarly problematic (but we don't).

sirosen: I was a little bit nervous about this but thought it was okay. I can confirm that I see the same issue.

I'm a bit confused though -- this is all inside of a single quoted string, so shouldn't backticks and `$` be suppressed? Have I miscounted the number of layers of escaping or something?

globus_cli/parsing/shell_completion.py
```diff
@@ -168,12 +186,71 @@ def do_zsh_complete():
...
+      local IFS=$'\\t'
+      if type globus > /dev/null; then
+        COMPREPLY=( $( env COMP_LINE="$COMP_LINE" COMP_POINT="$COMP_POINT" \\
+                       globus --shell-complete BASH ) )
+      else
+        COMPREPLY=( )
+      fi
+      return 0
+    }
+    complete -F _globus_completion -o nospace globus;
```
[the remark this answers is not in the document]
sirosen: Ah, yes, I should revert this to the prior behavior.

I was thinking about the case of `globus --format=<TAB>`. However, that's actually not a complete match anyway, so the behavior under `-o default` is actually correct/just fine.

globus_cli/parsing/shell_completion.py
corpulentcoffee: > I'm a bit confused though -- this is all inside of a single quoted string, so shouldn't backticks and `$` be suppressed? Have I miscounted the number of layers of escaping or something?

I am not really a user of zsh let alone familiar with its internals, but I *think* that whatever you pass to `_arguments` might be re`eval`uated?
https://github.com/zsh-users/zsh/blob/master/Completion/Base/Utility/_arguments#L423

sirosen: I looked at the `_arguments` documentation again ( http://zsh.sourceforge.net/Doc/Release/Completion-System.html ) because that source... nearly made my head explode at first.
After rereading, it looks like that's the processing for descriptions of the form `((item1\:desc1 ...))`. (Since you found this, you've presumably already worked that out -- I'm trying to catch up here and confirm.)

The docs don't make it _at all_ clear that what we're passing is going to be `eval`ed, but that's definitely what's happening.
I've got a fix I'm playing with and trying to break now. I'll push it if I can't find a way of causing harm.

Issue #388, opened 2018-03-08
CLI Parity


aaschaer:
Here are the remaining gaps with the hosted cli other than the rm command which we already have an issue for. I don't think anything here is worth not deprecating the hosted cli for but @ranantha should confirm this.

Missing API support:
- the ability to get subtasks of successful delete tasks *
- `retrying` `code` `faults` and `subtask_id` fields on `globus task show -t` *
- `expansions` field on `globus task show`
- `credential_subject` on `globus endpoint show`

(* both of these have a disclaimer about potentially not being available in future versions of the cli)

Not missing API support:
- `--myproxy-lifetime option` for `globus endpoint activate`
- ability to delete servers by hostname with `globus endpoint server delete`
- `globus task wait` doesn't show a progress bar, files completed, or mb/s

sirosen:
The subtask functionality is all deprecated. (Hence the notes about not being available in the future.)
We can just strike those from the list.

---

Task `expansions` and Endpoint `credential_subject` are both simple fields.
If they're added to the API, we'll add CLI support, of course. Not much to say about those.

---

`--myproxy-lifetime` is very straightforward. We should just knock that out.

---

`globus endpoint server delete` by hostname is a little fiddly and not important enough to consider blocking.
We'd have various specialized cases to handle (multiple matches, no matches, etc), and users can already `endpoint server list` followed by `endpoint server delete`. It's not hard, but unless someone comes to us with a need for it, it's unnecessary complexity.

If someone were scripting endpoint server deletion by hostname, that would be legitimate, but I haven't heard of anyone doing that.

---

`globus task wait` can't show a progress bar. We don't have the necessary information to render one. I doubt that the one shown by the hosted CLI is accurate, or it may be based on information not exposed through the API.

Files completed and mb/s , both of which can be seen with a simple `globus task show`, are not important enough to disrupt the currently simple/clean output.
Printing `.` to indicate that the wait is still polling (and that the command didn't hang somewhere) is good and useful.

sirosen:
After doing a quick search of hosted CLI usage history, it looks like `server-remove` by hostname is the most common way that command is used.
So I was utterly wrong.

It may be worth doing this. Unfortunately, I can't think of any particularly clean way of exposing that in the API. It seems to conflict a little bit with the REST model.

We can do the current `server-remove -p` behavior with pretty low effort by just doing a list, check for full URI matches, check for `hostname:port` matches, and then finally check for `hostname` matches.
[...]

PR #389, opened 2018-03-22
Allow delete of Endpoint Servers by hostname, uri

sirosen:
Backwards compatible.
Prior usage:
`globus endpoint server delete EPID SERVER_ID`
New usage:
`globus endpoint server delete EPID SERVER [--mode id|hostname|uri]`

The `--mode` defaults to "id", meaning `SERVER` is a server ID.
`mode=hostname` means that `SERVER` is a `HOSTNAME` or `HOSTNAME:PORT` (easy enough to check).
`mode=uri` means that `SERVER` is `SCHEME://HOSTNAME:PORT`

I considered making the `--mode` determined by whether or not `SERVER` is an int or contains `://`.
This seems potentially fragile and I don't think it gets us significant benefit -- the caller already knows what they're passing, after all.

As a necessary step, refactor some of `endpoint server list` into a useful helper. I made the signature on it a little funky (returns string or response object depending on results) but I don't think it's so weird as to been unconscionable.

Thoughts much appreciated.

Contributes towards #388

### DIFF
--- a/globus_cli/commands/endpoint/server/delete.py
+++ b/globus_cli/commands/endpoint/server/delete.py
@@ -69,9 +69,21 @@ def server_delete(endpoint_id, server):
 
     mode = _detect_mode(server)
 
-    if mode != 'id':
-        endpoint, server_list = get_endpoint_w_server_list(endpoint_id)
+    # list (even if not necessary) in order to make errors more consistent when
+    # mode='id'
+    endpoint, server_list = get_endpoint_w_server_list(endpoint_id)
 
+    if server_list == 'S3':
+        raise click.UsageError('You cannot delete servers from S3 endpoints.')
+
+    # we don't *have to* raise an error in the GCP case, since the API would
+    # deny it too, but doing so makes our errors a little bit more consistent
+    # with deletes against S3 endpoints and shares
+    if endpoint['is_globus_connect']:
+        raise click.UsageError(
+            'You cannot delete servers from Globus Connect Personal endpoints')
+
+    if mode != 'id':
         matches = _spec_to_matches(server_list, server, mode)
         if not matches:
             raise click.UsageError('No server was found matching "{}"'
@@ -80,7 +92,7 @@ def server_delete(endpoint_id, server):
             raise click.UsageError(dedent("""\
                 Multiple servers matched "{}":
                     {}
-            """).format(server, [x['id'] for x in server_list]))
+            """).format(server, [x['id'] for x in matches]))
         else:
             server = matches[0]['id']
 
--- a/globus_cli/commands/endpoint/server/delete.py
+++ b/globus_cli/commands/endpoint/server/delete.py
@@ -1,30 +1,29 @@
 from textwrap import dedent
 import click
 
-from globus_cli.parsing import (
-    common_options, endpoint_id_arg, CaseInsensitiveChoice)
+from globus_cli.parsing import common_options, endpoint_id_arg
 from globus_cli.safeio import formatted_print, FORMAT_TEXT_RAW
 
 from globus_cli.services.transfer import get_client, get_endpoint_w_server_list
 
 
-def _spec_to_matches(server_list, server_spec, style):
+def _spec_to_matches(server_list, server_spec, mode):
     """
-    style is in {uri, hostname, hostname_port}
+    mode is in {uri, hostname, hostname_port}
 
     A list of matching server docs.
     Should usually be 0 or 1 matches. Multiple matches are possible though.
     """
-    assert style in ('uri', 'hostname', 'hostname_port')
+    assert mode in ('uri', 'hostname', 'hostname_port')
 
     def match(server_doc):
-        if style == 'hostname':
+        if mode == 'hostname':
             return server_spec == server_doc['hostname']
-        elif style == 'hostname_port':
+        elif mode == 'hostname_port':
             return (server_spec == '{}:{}'.format(
                         server_doc['hostname'],
                         server_doc['port']))
-        elif style == 'uri':
+        elif mode == 'uri':
             return (server_spec == '{}://{}:{}'.format(
                         server_doc['scheme'],
                         server_doc['hostname'],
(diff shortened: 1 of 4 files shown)


corpulentcoffee:
> I considered making the `--mode` determined by whether or not `SERVER` is an int or contains `://`.
This seems potentially fragile and I don't think it gets us significant benefit -- the caller already knows what they're passing, after all.

If you're at all on the fence about this, my vote would be to just auto-detect this for the caller. I don't *think* it would be fragile enough to justify making people type out `--mode=hostname`, if it's true that "`server-remove` by hostname" was the most common way people did this with the hosted CLI.

sirosen:
I am all on the fence about it.
To be clear, we're thinking of having basically the same underlying mode logic decided by a check like
```python
if isinstance(server, int):
    mode = 'id'
elif '://' in server:
    mode = 'uri'
elif ':' in server:
    mode = 'hostname_port'
else:
    mode = 'hostname'
```

Is that right? I suppose `id`s will always be ints?

corpulentcoffee:
> ```python
> elif ':' in server:
>     mode = 'hostname_port'
> ```

I guess this wouldn't work if a GCS server were specified by IPv6 address, but it looks like Transfer doesn't support doing that.

> I suppose `id`s will always be ints?

Well, if, in the future, Transfer switches to UUIDs, those would be easy to detect too.

I would expect everything you proposed here to work just fine in today's reality, though. You could future-proof things if you wanted to, though (e.g. checking for UUIDs, checking for two or more colons / brackets for IPv6).

sirosen:
I don't want to be overly paranoid. If you're comfortable with that detection logic, I am.

We can act when the first IPv6 bug report comes in -- at that stage, there will be a valid workaround (just use the ID) to hold them until we can add support.

I agree that Transfer changing IDs is easy enough to handle. I doubt that would ever happen though.

I'll make the change.

sirosen:
Sorry this is taking so long. We've been somewhat lazy about test coverage, but I felt that this is important client-side logic to actually have under test coverage.
I've written some minimal tests, and will push my changes assuming that they all work out okay.

sirosen:
Changes with new tests are now up.

globus_cli/commands/endpoint/server/delete.py
```diff
@@ -1,21 +1,90 @@
...
     Executor for `globus endpoint server show`
     """
     client = get_client()
-    server_doc = client.delete_endpoint_server(endpoint_id, server_id)
 
-    formatted_print(server_doc, text_format=FORMAT_TEXT_RAW,
+    mode = _detect_mode(server)
+
+    if mode != 'id':
+        endpoint, server_list = get_endpoint_w_server_list(endpoint_id)
```
corpulentcoffee: You may want to raise a usage error here if the `server_list` returned by `get_endpoint_w_server_list` returns `'S3'` (otherwise `_spec_to_matches` will be operating on a string instead of the iterable response it is actually expecting):

```
~$ globus endpoint server delete cf9bcaa5-6d04-11e5-ba46-22000b92c6ec example.com
TypeError: string indices must be integers

program returned status code 1
```

corpulentcoffee: Adding to this, if you do decide to have a `'S3'` check here: you *could* go ahead and call `get_endpoint_w_server_list` regardless of the `mode` so that the user always gets consistent messaging:

```
~$ globus endpoint server delete 0f27d74c-1b41-11e8-b6eb-0ac6873fc732 123456
HTTP status:      400
request_id:       UEaLoAYy5
code:             BadRequest
message:          That endpoint type doesn't have servers

program returned status code 1

~$ globus endpoint server delete 0f27d74c-1b41-11e8-b6eb-0ac6873fc732 example.com
Usage: globus endpoint server delete [OPTIONS] ENDPOINT_ID SERVER

Error: 0f27d74c-1b41-11e8-b6eb-0ac6873fc732 (My Test Share) is a share and does not have servers.

To see details of the share, use
    globus endpoint show 0f27d74c-1b41-11e8-b6eb-0ac6873fc732

To list the servers on the share's host endpoint, use
    globus endpoint server list ddb59aef-6d04-11e5-ba46-22000b92c6ec


program returned status code 2
```

globus_cli/commands/endpoint/server/delete.py
```diff
@@ -1,21 +1,90 @@
...
+
+        matches = _spec_to_matches(server_list, server, mode)
+        if not matches:
+            raise click.UsageError('No server was found matching "{}"'
+                                   .format(server))
+        elif len(matches) > 1:
+            raise click.UsageError(dedent("""\
+                Multiple servers matched "{}":
+                    {}
+            """).format(server, [x['id'] for x in server_list]))
```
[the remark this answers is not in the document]
sirosen: Yes, I did. But I only tested in cases where everything is a match so... whoops.

globus_cli/commands/endpoint/server/delete.py
sirosen: Playing with this more, I *really* dislike having the different error modes based on what you pass. Agree with your suggestion -- we'll always list, even if we don't need it.

I started to believe myself a little bit too much with the tests I wrote and forgot entirely to check on the S3 case (which the testsuite can't do because S3 endpoint creation is hosted CLI only). Will rectify along with the aforementioned fix.

globus_cli/commands/endpoint/server/delete.py
```diff
@@ -69,9 +69,22 @@ def server_delete(endpoint_id, server):
...
+    endpoint, server_list = get_endpoint_w_server_list(endpoint_id)
+
+    if server_list == 'S3':
+        raise click.UsageError('You cannot delete servers from S3 endpoints.')
 
+    # GCPs will have one server in their list, with a null hostname
+    # we don't *have to* raise an error in this case, since the API would deny
+    # it too, but doing so makes our errors a little bit more consistent
+    # with deletes against S3 endpoints and shares
+    if not all(x['hostname'] is None for x in server_list):
```
corpulentcoffee: Is this check correct?

I think you want `server_list and all(x['hostname'] is None for x in server_list)` or maybe `len(server_list) == 1 and server_list[0]['hostname'] is None` or maybe even just `endpoint['is_globus_connect']`.

sirosen: You caught me before I squashed this.
I meant `all()` as opposed to `not all()`, yes.

`endpoint['is_globus_connect']` is cleaner though. Should I change to that?

PR #391, opened 2018-03-24
Install cryptography by default

### LINKED ISSUE #351: Consider making globus-cli[delegate-proxy] install by default
Inspired in part by globus/globus-sdk-python#235

Since `globus-cli[delegate-proxy]` uses `cryptography`, it may be possible for it to install on all platforms easily and cleanly.
Furthermore, if `globus-sdk` starts to install `pyjwt` anyway, with `cryptography` bundled, we're not losing anything in terms of support.

One area of concern and investigation is how we want to specify the `cryptography` dependency to minimize the chance of conflict or breakage with `pyjwt`.

sirosen:
Deprecate delegate-proxy extra install target (but leave it, empty).
Make all cryptography-conditional code non-conditional because cryptography is now always-on.

@aaschaer, I must apologize for unwinding much of your hard work on making everything work nicely when cryptography is absent, but I think it's for the Greater Good.

In #351 I noted that the SDK, via pyjwt, is already installing `cryptography>1.4`, so our constraint (`cryptography>=1.8.1,<2.0.0`) is safe.
However, it turns out that our way of specifying the cryptography version is wrong!
`cryptography` doesn't use semver and guarantees compatibility: https://cryptography.io/en/latest/api-stability/

~So I'm changing this to `cryptography>=1.8.1` in this PR, with a comment about its oddity~
Update cryptography dependency to `>=2.0.0,<2.5.0`

Because this is a fairly major and potentially dangerous change, I'd like everyone to explicitly approve before merging.

Closes #351

EDIT: Changed the specified `cryptography` version. Comment below explains why.

### DIFF
--- a/globus_cli/commands/endpoint/activate.py
+++ b/globus_cli/commands/endpoint/activate.py
@@ -1,14 +1,6 @@
 import click
 import webbrowser
 
-# only have --delegate-proxy appear if cryptography is available
-try:
-    import cryptography
-    # slightly hacky way of preventing flake8 from complaining
-    cryptography_imported = bool(cryptography)
-except ImportError:
-    cryptography_imported = False
-
 
 from globus_cli.parsing import common_options, endpoint_id_arg, HiddenOption
 from globus_cli.safeio import formatted_print, FORMAT_TEXT_RAW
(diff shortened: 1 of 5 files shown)


sirosen:
Dave's reading of the `cryptography` stability promises was slightly more pessimistic than my initial one, and after chatting a bit I think he's right.
Although in all probability `>=1.8.1` would be safe for the foreseeable future, the deprecation policy stated in their docs suggests that we should use `<2.5.0`, which allows for the next two major versions (`2.3.0` and `2.4.0`). I've updated the `setup.py` with the more conservative constraint and dropped a comment on why it is that.

Assuming this merges, I'll set an annual reminder for us to check the constraint, just so we don't lag behind too far if there are no other activities (e.g. critical security issues) which drive us to update.

aaschaer:
I'm completely in favor of unwinding my code for the greater good :)

Unless there has been a compatibility uproar from using cryptography in the SDK that I missed, this looks good to me.

globus_cli/commands/endpoint/activate.py
```diff
@@ -1,14 +1,6 @@
...
 import webbrowser
 
-# only have --delegate-proxy appear if cryptography is available
-try:
-    import cryptography
-    # slightly hacky way of preventing flake8 from complaining
-    cryptography_imported = bool(cryptography)
-except ImportError:
-    cryptography_imported = False
-
```
corpulentcoffee: `globus_cli/services/transfer.py` also uses `cryptography_imported` and this detection code.

sirosen: Good catch. Will rectify. (Of course, all of that code still runs fine, just isn't necessary.)

PR #404, opened 2018-04-13
Add support for `--format=unix`

sirosen:
Pull in the awscli `--output=text` formatter, modify it very slightly to make it usable, clearly marking modifications, and setup `--format=unix` to pass `--jmespath` processed data through this transform.
(It's Apache2 licensed, and we discussed that we're comfortable with this usage.)

`--format=unix` is not guaranteed to be pretty and friendly, and with certain nesting structures can actually be quite a mess. But it is *consistent* and *good enough*.

The data is output as TSV, and we're okay with that. Fields containing tabs may expand in nasty ways -- but that's pretty rare and people shouldn't do that anyway. If you need really careful data handling, use the SDK or the `-Fjson` output.

This needs some thorough testing and documentation before it can be released, and I plan to add that to this PR before we consider merging it. However, preliminary review would help.

I've already found at least one case in which this fails
```
globus task list -Funix --jq 'DATA[*].[task_id, status, event_link]'
```
which suggests that I'm somehow misapplying the awscli text module.

### DIFF
--- a/adoc/include/format_option.adoc
+++ b/adoc/include/format_option.adoc
@@ -1,3 +1,22 @@
-*-F, --format* '[json|text]'::
+*-F, --format* '[json|text|unix]'::
 
 Set the output format for stdout. Defaults to "text".
++
+*TEXT* output is human-friendly textual output, usually in table or
+record-oriented format.
+In some cases, *TEXT* format is intentionally kept simple to support naive use
+of commands within a subshell, but it is not generally guaranteed to be
+parseable.
++
+*JSON* output will produce raw JSON data from underlying Globus services.
+This data is useful for using the CLI as part of a programmatic process, or for
+deciding on valid *--jmespath* queries.
++
+*UNIX* output will produce a line-by-line serialization of the result data
+(which would be visible with *JSON*).
+This is very suitable for use in pipelines with `grep`, `sort`, and other
+standard unix tools.
++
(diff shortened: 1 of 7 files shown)


sirosen:
I'm able to reproduce the failure with `awscli` by trying the same thing:
```
$ aws ec2 describe-instances --query 'Reservations[*].Instances[*].[VpcId, Placement]' --output text
'str' object has no attribute 'items
```

Should we be doing something to wrap the error? We could try printing the output and catch `AttributeError`. It wouldn't be very pretty, but it would work.

sirosen:
Some further thought on this after some experimentation:

The issue appears to be that the formatter can't handle arrays containing objects and non-object JSON values. I doubt we're cleverer than the awscli maintainers, so I think we should accept this constraint.

Documentation should not try to go into detail about exactly what does/does not work.
We should direct people to extract exactly the fields they want and note that `--jmespath` pads unfound values with nulls.

We also need to make the error nicer -- awscli can get away with that raw AttributeError, but I think our audience is sufficiently different.

Solutions for when people complain about this:

- Pull exact fields (best)
```
globus task list -Funix \
  --jq 'DATA[*].[task_id, status, fatal_error.code, fatal_error.description]'
```

- Use JMESPath's `to_string()` to get a JSON object encoded in the output
```
globus task list -Funix --jq 'DATA[*].[task_id, status, to_string(fatal_error)]
```

PR #406, opened 2018-04-23
catch TransferAPIError raised in BookmarkList

aaschaer:
@ranantha ran into a TransferAPIError raised by `globus bookmark list` when we try to get the display name for an endpoint that has been deleted.

I figure catching the error and displaying `[ENDPOINT DELETED]` is fairly sane behavior?

None of the other bookmark commands try to get the endpoint's display name.

### DIFF
--- a/globus_cli/commands/bookmark/list.py
+++ b/globus_cli/commands/bookmark/list.py
@@ -1,5 +1,6 @@
 import click
 
+from globus_sdk import TransferAPIError
 from globus_cli.parsing import common_options
 from globus_cli.safeio import formatted_print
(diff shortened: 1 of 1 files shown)


sirosen:
I will now name all of my endpoints `[DELETED ENDPOINT]`.
Kidding.

This behavior seems reasonable to me. If you want Truth, `-Fjson` is always your friend.
I wouldn't mind changing the text format for `bookmark list` to make this unambiguous, but in theory that could be a breaking change for someone. Certainly I don't want to do that right before GlobusWorld.

PR #407, opened 2018-04-26
Ensure that JSON string output is a *JSON* string

sirosen:
Don't make quoting of string output conditional. This is a behavioral change and therefore wrong.

Should I add a test for this as well? e.g. `globus whoami --jmespath 'Username'` and make sure it's quoted?


corpulentcoffee:
This fix works.

> Should I add a test for this as well? e.g. globus whoami --jmespath 'Username' and make sure it's quoted?

Sure, that would be a good one.

PR #409, opened 2018-05-08
Make JSON output use `sort_keys` to force ordering

sirosen:
When dumping an object to JSON output, use `sort_keys` to ensure that ordering is consistent.

This doesn't need to be a documented part of the interface for the CLI, that we use sorted JSON output, but it results in more consistent behavior. Nobody *should* be relying on the order of the output, but
this way, if they *are*, we'll be able to give them better behavior.

### DIFF
--- a/globus_cli/safeio/output_formatter.py
+++ b/globus_cli/safeio/output_formatter.py
@@ -69,7 +69,7 @@ def _jmespath_preprocess(res):
 
 def print_json_response(res):
     res = _jmespath_preprocess(res)
-    res = json.dumps(res, indent=2, separators=(',', ': '))
+    res = json.dumps(res, indent=2, separators=(',', ': '), sort_keys=True)
     safeprint(res)
 
 


jaswilli:
I don't care much either way but I will say that I'm not sure you can get away with sorting keys and also claiming to not guarantee any particular sorting behavior. Once you start doing it you're pretty much stuck with continuing to do it...

sirosen:
Yeah, let me get those too. I hadn't even thought to audit for other `json.dump{,s}` usage.
It does no harm, and sets a clean/consistent internal rule: "When you make JSON output, you always sort it."

jaswilli:
Which sort of violates our soft-rule of passing through API responses as-is.

I'm fine with this though you really like sorting.

sirosen:
> I'm not sure you can get away with sorting keys and also claiming to not guarantee any particular sorting behavior

I agree with this, so maybe I should clarify my intent here.
- I think it's reasonable for us to guarantee things which we don't make part of the documented contract we have with our users.
- But those things are still guarantees we are making, and changing/breaking those guarantees is backwards-incompatible behavior

The only reason I want to not document it is that I think it will encourage foolish/dangerous behavior.
e.g.
```
# get filename + permissions, badly
$ globus ls $ep_plus_path -Fjson | grep -A1 '"name"' | cut -d':' -f2 | tr -d '",'
```

> Which sort of violates our soft-rule of passing through API responses as-is.

Isn't this rule already violated? We're deserializing and reserializing the data we get from the APIs.
If it came to us in sorted order, we might not respect that.

We could do this in SDK-land and have response objects load JSON data with `collections.OrderedDict`, but I'm not sure if that's right either?

---

> I'm fine with this though you really like sorting.

I'm not super-committed to it. I figured it was easier to kickstart discussion with a concrete PR.
We can reject this if we think it's a bad idea.

The reason I have for doing this is that it makes output more consistent, which makes it nicer for comparison and diffing.

So if you
```
globus ls $eppluspath -Fjson > somefile.json
globus ls $eppluspath2 -Fjson > somefile2.json
diff somefile.json somefile2.json
```
I think you'll get better results if we sort. Especially if those `globus ls` commands might run on different machines or under different CLI versions.

It's a weak rationale and we shouldn't pretend otherwise. I see this as low-cost.

jaswilli:
> Isn't this rule already violated? We're deserializing and reserializing the data we get from the APIs.
If it came to us in sorted order, we might not respect that.

True. I guess a more accurate representation would be we don't really _purposefully_ manipulate API responses.

Anyway, like I said, I'm fine with this as I think it's pretty reasonable and well as low risk.

corpulentcoffee:
> Isn't this rule already violated? We're deserializing and reserializing the data we get from the APIs.
If it came to us in sorted order, we might not respect that.

Yes, this rule is already violated and I actually view this PR as improving the situation for Pythons before 3.6 (because of not having PEP 468 implemented on those older Pythons). If you try something like `globus task list` on the mainline, even though the Transfer API returns responses uses sorted keys, you'll see randomly-ordered keys on Python 2.7 and sorted keys on Python 3.6 because of the deserialization and reserialization issue.

PR #410, opened 2018-05-08
Add several examples of `--format=UNIX` to docs

sirosen:
Docs for several commands feature exciting uses of `--format=UNIX` output amongst their various examples. Although it would be nice to get these examples in place for *all* commands, this is a decent start.

### DIFF
--- a/adoc/include/jmespath_option.adoc
+++ b/adoc/include/jmespath_option.adoc
@@ -1,8 +1,10 @@
 *--jq, --jmespath* 'EXPR'::
 
-Supply a JMESPath expression to apply to json output.
-Takes precedence over any specified '--format' and forces the format to be json
-processed by this expression.
+Supply a JMESPath expression to apply to JSON output.
++
+This option is compatible with '--format=JSON' and '--format=UNIX'.
+If '--format=TEXT' is specified, output will be rendered as if '--format=JSON'
+had been specified instead.
 +
 A full specification of the JMESPath language for querying JSON structures may
 be found at https://jmespath.org/
--- a/adoc/task_list.1.adoc
+++ b/adoc/task_list.1.adoc
@@ -99,4 +99,14 @@ $ globus task list --limit 100 \
     --filter-not-label 'autolabel' --exact
 ----
 
+List active transfers in a tabular format suitable for consumption by unix
+tools:
+
+----
+$ globus task list --format unix \
+    --jmespath 'DATA[?status==`ACTIVE`].[task_id, source_endpoint_id, destination_endpoint_id, label]'
+----
+
+NOTE: 'destination_endpoint_id' will be 'None' in the case of Delete tasks.
+
 include::include/exit_status.adoc[]
--- a/adoc/task_list.1.adoc
+++ b/adoc/task_list.1.adoc
@@ -109,4 +109,24 @@ $ globus task list --format unix \
 
 NOTE: 'destination_endpoint_id' will be 'None' in the case of Delete tasks.
 
+Cancel all tasks with expired credentials:
+
+----
+# careful: do not quote this output, but instead rely on the shell's
+# word-splitting for this for-loop
+for id in $(globus task list --filter-status=INACTIVE \
+                             --format=UNIX \
+                             --jmespath='DATA[*].task_id'); do
+    globus task cancel $id
+done
+----
+
+Print out task statuses with links to their pages in the web UI:
+
+----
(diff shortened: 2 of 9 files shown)


adoc/include/jmespath_option.adoc
```diff
@@ -1,8 +1,9 @@
 *--jq, --jmespath* 'EXPR'::
 
-Supply a JMESPath expression to apply to json output.
-Takes precedence over any specified '--format' and forces the format to be json
-processed by this expression.
+Supply a JMESPath expression to apply to JSON output.
++
+If '--format' is specified as 'TEXT', this takes precedence and forces the
+format to be JSON processed by this expression.
```
corpulentcoffee: I find the "this" in "this takes precedence" potentially confusing because the most-recently mentioned thing that it might refer to is `'TEXT'`, which is actually what loses out.

I would maybe write this as "JMESPath expressions are compatible with output `--format`s of `JSON` or `UNIX`; if `TEXT` is specified, the output will be rendered as if `JSON` was specified".

adoc/task_list.1.adoc
```diff
@@ -99,4 +99,14 @@ $ globus task list --limit 100 \
...
 
+List active transfers in a tabular format suitable for consumption by unix
+tools:
+
+----
+$ globus task list --format unix \
+    --jmespath 'DATA[?status==`ACTIVE`].[task_id, source_endpoint_id, destination_endpoint_id, label]'
+----
+
+NOTE: 'destination_endpoint_id' will be 'null' in the case of Delete tasks.
```
corpulentcoffee: Well, if would be `null` for `--format=json`, but you have `--format=unix` in this example, so it'll actually be `None`.

adoc/transfer.1.adoc
```diff
@@ -178,4 +178,28 @@ godata mygodatadir -r
...
+$ task_id="$(globus transfer $source_ep $dest_ep \
+    --jmespath 'task_id' --format=UNIX \
+    --batch < my_file_batch.txt)"
+$ echo "Waiting on 'globus transfer' task '$task_id'"
+$ globus task wait "$task_id" --timeout 30
+$ if [ $? -eq 0 ]; then
+    echo "$task_id completed successfully";
+  else
+    echo "$task_id failed!";
+  fi
```
corpulentcoffee: You switch back and forth in this PR between presenting examples as if they were interactive (with leading `$`s for each command) and as if they were in a script (without the leading `$`s).

I think this example is definitely a candidate for the "as if they were in a script" treatment if not for the `cat`ing of `my_file_batch.txt` to show what is in it.

Maybe if you did want to show this one as if it were a script, you could either have a separate (labeled) AsciiDoc code block for `my_file_batch.txt` or you could use a `<<END` heredoc in the script.

corpulentcoffee:
It also occurs to me that it might be worthwhile to have an example of using `--format=unix` that isn't just assigning it to a variable so people "get" why `--format=unix` is useful.

Maybe a `for` loop?

```bash
# cancel all tasks with expired credentials
for id in $(globus task list --filter-status=INACTIVE \
                             --format=UNIX \
                             --jmespath='DATA[*].task_id'); do
  globus task cancel $id
done
```

Or maybe some silly pipeline example?

```bash
# print out my tasks' statuses with links
globus task list --format=unix --jmespath='DATA[*].[task_id, status]' |
awk '{printf "Task %s is currently %s\n", $1, $2;
      printf "View at https://www.globus.org/app/activity/%s\n\n", $1}'
```

sirosen:
Let's add these examples. I think leaving `IFS` and `awk -F` unset is fine -- otherwise, people will think that's strictly required and will ask us about those.

adoc/transfer.1.adoc
sirosen: After hemming and hawing a bit, I think I'm just going to write two separate code blocks. One that reads as `$ cat my_file_batch.txt` and a following one without the leading `$`.

I want to avoid the heredoc for people who won't know what it is.

adoc/task_list.1.adoc
sirosen: Yes, thank you. Should I document as `None`, or maybe just drop the note and people can learn by trying it out?

adoc/include/jmespath_option.adoc
sirosen: Yes, *this* can be cleaned up. *This* is confusing.

PR #417, opened 2018-06-04
fix endpoint role list's handling of group principals

aaschaer:
Ran into an error in `globus endpoint role list` when a role's principal was a group so the `username` variable wasn't being initialized.


sirosen:
I'll wait for tests to pass, but looks like a good fix.

Maybe we should generate a one-off test group which we can use to add tests for this? I'll open up a follow up issue, but won't block on this.

PR #425, opened 2018-07-16
Release v1.6.3

jaswilli:


sirosen:
I was about to open almost exactly the same thing. I'll tag and push to pypi.

PR #426, opened 2018-07-16
Update supported python versions

jaswilli:
Drop 3.3, move 3.7 to release from dev, and add 3.8-dev.


sirosen:
FWIW, I'm updating the docs with the 1.6.3 release and some minor changes, and it looks like we're already saying 3.4+ there. So this is consistent.

PR #428, opened 2018-07-24
add --no-default-directory option

aaschaer:
Adds the ability to pass an `EXPLICIT_NULL` as the `default_directory` on endpoint create or update with a `--no-default-directory` option.

Also fixes a typo Karl found.


sirosen:
I'll wait for the build to pass, but looks good.

PR #430, opened 2018-07-24
Add section to contrib doc about releasing

sirosen:
So that our release procedure is documented.

I noticed we didn't have this when getting Aaron set to do a release (because we realized he hadn't done one before).
Unlike the other parts of the contrib doc, I'm not interested in this being super-friendly -- it just need to be what someone new to the project but sufficiently knowledgeable needs to know.

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -141,13 +141,14 @@ If you are a CLI maintainer, you need to be able to do a release.
 Setup:
 
   - Make sure you have a pypi account
-  - Setup your credentials for twine (the pypi upload tool) https://github.com/pypa/twine[twine docs for detail]
-  - Make sure you have a gpg key setup for use with git
+  - Setup your credentials for twine (the pypi upload tool)
+      https://github.com/pypa/twine[twine docs for detail]
+  - Make sure you have a gpg key setup for use with git.
+      https://git-scm.com/book/en/v2/Git-Tools-Signing-Your-Work[git-scm.com guide for detail]
 
 Do it:
 
-  - `rm -r dist/` to make sure there are no old or weird source dists lying around
-  - `make upload` to build new source dists in `dist/` and upload via twine
-  - `git tag -s $VERSION` and write `v${VERSION}` in the tag. e.g. `v1.7.0`
+  - `make release` to build source dists, upload via twine, and tag the release
+  - push the newly created version tag to GitHub
   - Ensure that `docs.globus.org` gets updated with new docs/changelog
       Procedure is set in that repo.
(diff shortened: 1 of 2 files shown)


aaschaer:
Since I ran into it, maybe put something about GPG in the setup section?

CONTRIBUTING.adoc
```diff
@@ -131,3 +131,23 @@ Some rules and tips for the python and documentation in this project.
...
+
+  - Make sure you have a pypi account
+  - Setup your credentials for twine (the pypi upload tool) https://github.com/pypa/twine[twine docs for detail]
+  - Make sure you have a gpg key setup for use with git
+
+Do it:
+
+  - `rm -r dist/` to make sure there are no old or weird source dists lying around
+  - `make upload` to build new source dists in `dist/` and upload via twine
+  - `git tag -s $VERSION` and write `v${VERSION}` in the tag. e.g. `v1.7.0`
```
shifflett-globus: Would there be any appetite to just incorporate these steps into the `Makefile`?

sirosen: Some of it, at least.
Cleaning `dist/` for sure -- I'll even go so far as to do that in this PR.

`VER=$(python -c ...) git tag -s "$VER" -m "v${VER}"`? I'm less sure.

And I don't think the makefile can safely push the tag because we don't all necessarily name the repo the same thing -- `origin` vs `upstream`, etc.

sirosen:
I don't want to say more about GPG in the contrib doc than what I'm adding so far -- we could easily dive into a whole slew of issues which I don't think we really want to deal with in the repo.
If there's some minimal addition that makes sense though, let's do that.

Maybe link to [git-scm.com's guide on the topic](https://git-scm.com/book/en/v2/Git-Tools-Signing-Your-Work)?

CONTRIBUTING.adoc
shifflett-globus: > Cleaning `dist/` for sure -- I'll even go so far as to do that in this PR.

Yes, the cleaning for sure -- if having a (partially- or fully-) clean working tree is something you want to happen before people upload for a release, by all means make it automatic.

> `VER=$(python -c ...) git tag -s "$VER" -m "v${VER}"?` I'm less sure.

Yes, I agree this is less fun, but something like `<globus_cli/version.py grep '^__version__' | grep -oe '[0-9.]*'` could also give it to you, though, if you wanted to avoid starting a Python interpreter for some reason.

> And I don't think the makefile can safely push the tag because we don't all necessarily name the repo the same thing

Yes, absolutely, and I think just having an `echo` that just says "please push the new v$(VERSION) tag to the appropriate repository" will keep people from forgetting.

sirosen:
Dave and I chatted a little. I've made some further changes, including changes to the makefile.

I also renamed `make upload` to `make release` since that now seems more appropriate.

PR #436, opened 2018-08-29
Release v1.8.0

jaswilli:


sirosen:
I'm not too concerned about the testing for this because it's already been tested multiple times.

I think the changelog might be a little bit too pithy though.
Can we say a little more about "adds `globus session` commands" or something so that we can easily see that this is the version which added those?

If you agree and expand the documentation of this change (or convince me otherwise), I'm happy to merge and do the release dance.

sirosen:
We discussed that "HA support" is clear enough. Will start moving the machinery of the release.

PR #438, opened 2018-11-05
Update WebApp links

sirosen:
- all links to www.globus.org/app are rewritten as app.globus.org
- no other changes to the request URIs are made

Primarily impacts the display of Group links, but also some activation links and an example on one of the docs pages of a task overview link.

There's also a one-line addition to `make clean` because I did a `make clean; fgrep -R 'www.globus.org'` and was annoyed to see pyc files in my results.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,3 +69,4 @@ clean:
 	-rm -r dist
 	-rm -r build
 	-rm -r *.egg-info
+	find . -name '*.pyc' -delete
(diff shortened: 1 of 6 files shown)


sirosen:
Thanks for making the change for me. I was just getting around to it now, but I see it's done. I'm going to merge -- I'm treating your commit as approval for the parts you did not change.

PR #439, opened 2018-11-05
Minor fixes for flake8

sirosen:
pycodestyle updates are flagging some new, unexciting issues. Update to fix.


globus_cli/services/transfer.py
```diff
@@ -316,7 +316,7 @@ def check_completed():
 
     # Tasks start out sleepy
     if meow:
-        safeprint("""\
+        safeprint(r"""
```
shifflett-globus: Is the extra space above the cat a breaking change? (It looks better, though, FWIW.)

PR #446, opened 2019-02-21
Update dependencies and release 1.9.1

jaswilli:
Basically just a release for the sake of doing a release every so often. `cryptography` 2.5 has breaking changes but none of them seem to apply.

### DIFF
--- a/globus_cli/version.py
+++ b/globus_cli/version.py
@@ -2,7 +2,7 @@ from distutils.version import LooseVersion
 
 # single source of truth for package version,
 # see https://packaging.python.org/en/latest/single_source_version/
-__version__ = "1.9.0"
+__version__ = "1.9.1"
 
 # app name to send as part of SDK requests
 app_name = "Globus CLI v{}".format(__version__)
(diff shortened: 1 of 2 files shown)


sirosen:
Looks like we don't have any changes since the last release. I'm reading that right, yeah?

Do you want to maybe do the SDK first, so that we could update to globus-sdk v.1.7.1 here? Or don't care?

I'll merge and release if the build passes unless you say otherwise.
If we get a "flaky test" failure I might trigger a rerun.

jaswilli:
> Do you want to maybe do the SDK first, so that we could update to globus-sdk v.1.7.1 here? Or don't care?

I don't but I didn't ignore the pending changes, merely overlooked them. So yes, we may as well just do two uninteresting releases. I'll amend this when we get sdk 1.7.1 out the door.

jaswilli:
Changed to pull in SDK v1.7.1.

sirosen:
Although the Checks in GitHub aren't being updated, Travis shows that this is building. I think that Checks subsystem might not be kept as clean as we might expect. I bet it will update when the whole build is done.

I'm waiting for the output -- might not get to this until tomorrow.

sirosen:
Sorry for taking so long. I was on a brief vacation. Merging and releasing now.

Issue #449, opened 2019-02-26
Contrib standard update: black/isort + doc update


sirosen:
https://github.com/globus/globus-sdk-python/pull/325 was a lot of noise, but put us in a good state going forward.

I'd like to do the same here:
- add black and isort, apply to code
- do black and isort linting
- make things into separate Travis checks
  - also might be a good time to use Travis windows and drop appveyor
- move as much tool config into `setup.cfg` as possible
- tighten up contrib doc

However, the contrib docs mismatch in multiple ways. I think we could backport the release steps from this one into the SDK doc.

sirosen:
Closed by #454

Issue #445, opened 2019-02-12
Add "best effort" detection for `globus update` requiring permissions which the user does not have


sirosen:
If an admin installs the CLI for a user, (e.g. root-owned virtualenv install and executable symlink from `/usr/local/bin/`), `globus update` results in permission denied errors from the underlying pip command.

The better behavior in this case is for us to look at the install dir and check the ownership/permissions to see if it's user-writeable and at least try to catch this case.
I'd also add a flag like `--skip-permissions-check` to `globus update` in case we get things wrong.

I'm not really sure how feasible this is -- or how accurate it would be.
I'd look at `site.getsitepackages()`, but it isn't available inside of a virtualenv and it doesn't show us a `--user` install path.
The best information seems to come from `globus_cli.__file__`, which works under a wheel install, source install, etc. The only time it doesn't seem to work, as far as I can tell, is when you run from source without installing or do an editable source install.

Perhaps the best option is to check if the current user can write `globus_cli.__file__` and do nothing else. Worst case, `pip` throws a sensible, if long, error message.

sirosen:
The amount of jumping through hoops this entails + our current recommendation of pipx makes it unlikely we'll do anything with this. It's only ever come up once for one specific user, and nobody else has requested it to my knowledge.

If this is really needed, please reopen.

PR #454, opened 2019-03-13
Fixup contribution tools and doc to be more like globus-sdk

### LINKED ISSUE #448: Replace egg distribution with a universal wheel
Seeking greater uniformity between CLI and SDK tooling.

Also a universal wheel installs faster and is more "modern".

sirosen:
Note that one small, but significant, thing slipped into here is switching from egg bdists to wheels. So this will close #448 . It wasn't worth it to me to keep that separate, but I can pull it out into a separate PR if it's desirable.

As a whole, this addresses #449 .

- remove appveyor
- use travis matrix build with windows workers
- add `[development]` extra with dev requirements
- use tox
- separate linting from code builds
- apply isort and black linting + flake8-bugbear
- update contrib doc

I'm going to start doing a full read of everything to be sure that `black` didn't do anything weird~, but I wanted to get the PR build running too. I'll remove the "on hold" when I'm done~.

EDIT: accepting that my ability to read and parse large globs of code is a poor verification pass, the resulting code doesn't appear to have any semantic changes, even in weird corner cases with multiline strings and so forth. I think this is ready for primetime if tests pass.


sirosen:
Per #445, I think that one of the build failures we see in here is a real Windows issue. I'm going to patch the tests here to make things pass, rather than trying to incorporate a fix for that issue.

PR #458, opened 2019-03-15
Rename 'session boost' to 'session update'

sirosen:
We've been asked to change these command names and we should do so now.

To clarify for anyone reading our repo changes: much of the discussion about how/why to change these was internal, but I don't intend to restate it here.


sirosen:
Assuming that's the only fix, feel free to squash on merge to avoid the lint-failing commit going into the tree.

PR #463, opened 2019-04-29
Add `--limit` as a param on EP search

sirosen:
- limit defaults to 25
- is the limit on number of results returned
- when the limit applies (as detected by the paginator), print a warning to stderr -- use `click.style` to get yellow ANSI color on non-Win platforms when stderr is a tty

Will require an SDK release (to expose `limit_less_than_available_results`) before this can merge.

Here's an example of what the output looks like:
image

Which is, I think, what we were going for.
This inspires me to take a look at `click.style` on some of our other output, to get colors without headache on the non-Win terminals.

### DIFF
--- a/globus_cli/commands/endpoint/search.py
+++ b/globus_cli/commands/endpoint/search.py
@@ -39,8 +39,15 @@ from globus_cli.services.transfer import (
         'Username, as in "go@globusid.org"'
     ),
 )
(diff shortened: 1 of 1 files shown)


sirosen:
I've rebased this, since SDK v1.8.0 adds the needed property, and updated the labels.

PR #466, opened 2019-07-12
Update version of click (and other libs)

### LINKED ISSUE #461: RFE: Support later versions of jmespath 0.9
Hello!

I am looking at how hard it would be for Globus to be ported to FreeBSD, and I found two dependency issues that would need to be addressed, with click and jmespath.

jmespath has a hard dependency on version 0.9.2, but that version is not available in FreeBSD-land right now: FreeBSD 11 has 0.9.3 as a package, and the ports tree has version 0.9.4.

I did a quick check on the jmespath changelog, and 0.9.3 and 0.9.4 only seem to be bugfix releases.  So, I was wondering if it is possible for the jmespath dependency to be changes to `jmespath>=0.9.2,<1.0`?

There is also an upcoming issue with click: The details on why not to change the dependencies were already covered in #453, so I won't rehash it here.  As an FYI, FreeBSD 11 ships with a 6.7 version of click, but the latest version available in the ports tree is 7.0.  So, not an issue right now.

### LINKED ISSUE #465: Update to click v7.x and cleanup custom code in the process
- [ ] Look at the [click 7.0 changelog](https://github.com/pallets/click/blob/master/CHANGES.rst#version-70)
  - [ ] Identify features which can eliminate customizations of ours. In particular, these are ones I know about:
    - [custom metavars can maybe go](https://github.com/pallets/click/pull/675)
    - [case sensitive choice type can go](https://github.com/pallets/click/issues/569)
    - [custom date type can go](https://github.com/pallets/click/pull/423)
    - [broken pipe handling has been improved, check on our excepthook](https://github.com/pallets/click/pull/1106)
    - [hidden opt type can probably be removed](https://github.com/pallets/click/pull/500)
    - custom completion can potentially be replaced (numerous changes)
    - changes to automatic command naming may let us use that feature
  - [ ] Identify features and changes which could cause issues or break us
    - several changes to helptext formatting could make our help pages explode a bit, look for this
- [ ] open a branch with the update, test the results
  - make sure to use each option type and option set
  - take time with this, don't rush
- [ ] add any new tests for changes in behavior which we want to check
  - e.g. check that `cmd --help` does not include hidden opts
  - stick to non-network tests, as we want to do the same kind of change that we did in globus-sdk
- [ ] merge to master and release as a minor version update
  - no intentional breaking behavioral changes should be made
  - avoid bundling this with other updates

sirosen:
Updating to click 7.0 is the most major part of this.
Also bumps SDK, jmespath, cryptography, black, and flake8-bugbear.

closes #461, closes #465 

- Update the dependency versions
- Replace safeprint() with click.echo() now that Click handles EPIPE
- Custom excepthook will now allow non-ClickException errors through because click has a custom type of RuntimeError it uses for nonzero exits
  - click.exceptions.Exit was added to handle messy cases around non-standalone click invocations, but it means that SystemExit isn't being raised
  - click.exceptions.Exit isn't a descendant of ClickException to avoid accidental handling by users looking to catch the usage errors
  - no error handling we depend upon is impacted other than JMESPath compilation, which has been customized as a type of UsageError for this reason
- CaseInsensitiveChoice, HiddenOption, and ISOTimeType are all replaced with new click builtin behaviors
  - ISOTimeType -> click.DateTime has caused a behavioral change which I don't think is significant. Because we no longer retain the original string input, and go through `datetime.strptime(...).strftime(...)`, we'll include `00:00:00` in times given to us without an hours/minutes/seconds component when we send these times to the Transfer API
- Arguments with custom types no longer need to give explicit metavars if the type has a metavar
- Replace the custom shell completer with the built-in Click completer
  - In order to still support usage in `eval $(globus --completer)`, includes the completion shell functions directly in our source

I'm going to run this version locally to keep looking for things, but so far it seems to behave well.


globus_cli/commands/bookmark/create.py
```diff
@@ -7,11 +7,7 @@
 
 @click.command("create", help="Create a bookmark for the current user")
 @common_options
-@click.argument(
-    "endpoint_plus_path",
-    metavar=ENDPOINT_PLUS_REQPATH.metavar,
-    type=ENDPOINT_PLUS_REQPATH,
-)
```
jaswilli: Should we amend or remove this comment altogether?: https://github.com/globus/globus-cli/blob/336675ff24da64c5ee487243f39ae39fc49a7e14/globus_cli/parsing/endpoint_plus_path.py#L29

It seems like the default metavar handling has been added but this property still needs to remain because the metavar varies based on usage so the comment is no longer accurate.

globus_cli/commands/whoami.py
```diff
@@ -32,9 +31,8 @@ def whoami_command(linked_identities):
     try:
         res = client.oauth2_userinfo()
     except AuthAPIError:
-        safeprint(
-            "Unable to get user information. Please try " "logging in again.",
-            write_to_stderr=True,
+        click.echo(
+            "Unable to get user information. Please try " "logging in again.", err=True
```
jaswilli: This is weird. Probably happened when we did the `black` transition.

globus_cli/services/transfer.py
```diff
@@ -319,18 +319,18 @@ def check_completed():
...
             if meow:
-                safeprint(
+                click.echo(
                     r"""
                   _..
   /}_{\           /.-'
  ( a a )-.___...-'/
  ==._.==         ;
       \ i _..._ /,
       {_;/   {_//""",
```
jaswilli: 😻

jaswilli:
This looks good to me. If you want to address those minor/only-sort-of-related comments that's great but they could also be addressed in a follow up.

I tested this out on macOS and haven't found any issues so far.

globus_cli/commands/bookmark/create.py
sirosen: Yeah, we still need the property, but the comment needs an update. I can try to remember to do it tomorrow morning.

globus_cli/commands/whoami.py
sirosen: Almost for sure. I've caught some of these in other projects too.

globus_cli/services/transfer.py
sirosen: Hands down the best part of this product to date. I'm invested in testing it after each release.

sirosen:
I've just made those small fixes, and then I caught that the completer text was not properly escaped. It turns out not to have any impact, but it "looks weird" and it's obviously not what I meant to put in there.

I'm happy with this at this point. I'm kind of excited about ditching the custom completer. That code makes me a bit dizzy to revisit.

PR #470, opened 2019-07-26
Fix path normalization on Windows

### LINKED ISSUE #455: Windows bug: `globus delete` and `globus transfer` normalize paths in batchmode in order to join with base path
While trying to figure out why the builds for #449 were failing on Travis Windows workers, I had to do some digging around to see where/how paths are passed around.

If you look at `globus_cli.parsing.task_path`, you'll see it obviously and intentionally normalizes paths.
[The only place I found this used](https://github.com/globus/globus-cli/search?q=TaskPath&unscoped_q=TaskPath) is `globus delete` and `globus transfer` in batch mode.

On Windows, a globus CLI command running like this will normalize and then send Transfer paths with backslashes (properly escaped, if that's any comfort!).
According to [Transfer API docs](https://docs.globus.org/api/transfer/task_submit/#path_format) that's no good. Transfer will then split on `/` and find big clunking "filenames" with backslashes in them. So the Transfer will have unintended targets.

So I think we need to be a bit more Windows-conscious in how we handle these. It's nice on *nixes to normalize `/abc/def/../xyz`, but if it's wrong on Windows, then at the very least we need to turn it off on that platform.

Also, maybe change that docstring on TaskPath from

> only slightly more than string types

to

> here be dragons!

😉

sirosen:
Marking this as a work-in-progress because I don't have time to spec out the tests for this yet. I need to look at the testsuite's coverage of batch mode and see if we can consider the Travis testing to be sufficient.

batchmode transfers and deletes from a Windows system should still produce POSX-style paths. Otherwise, these will join incorrectly on Windows and then be passed to Transfer where they will be treated as POSIX paths. In order to handle this nicely without much headache, simply define our own normpath and pathjoin functions based on the posixpath implementation in the python stdlib.

This should *NOT* be thought of as "POSIX normalization", but rather "Globus Transfer normalization". While in practice those two may be the same, there are in theory subtle differences.

For example, this implementation takes the liberty of not caring about the number of leading slashes on a path, on the assumption that this can hardly even (if ever) matter to Globus Transfer.

closes #455

### DIFF
--- a/globus_cli/parsing/task_path.py
+++ b/globus_cli/parsing/task_path.py
@@ -15,20 +15,29 @@ def _normpath(path):
     Note that this does not preserve leading slashes in the same way as python's
     posixpath module -- it's not clear how Transfer would treat such paths and
     non-obvious that we need to allow such usage
+
+    Also, unlike normpath, we want to preserve trailing slashes because they may be
+    required
     """
     initial_slash = 1 if path.startswith("/") else 0
+    trailing_slash = 1 if path.endswith("/") and path != "/" else 0
     parts = path.split("/")
     new_parts = []
     for part in parts:
         if part in ("", "."):
             continue
         # either not adding a ".." OR chaining together multiple ".."s
-        if part != ".." or (new_parts and new_parts[-1] == ".."):
+        # OR working with a non-absolute path that starts with ".."
+        if (
+            part != ".."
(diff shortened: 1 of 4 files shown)


sirosen:
I've added additional testing to ensure the path normalization methods are exercised properly. Removing "WIP" from the title.

PR #471, opened 2019-07-27
Convert testsuite from nose2 to pytest

sirosen:
By converting the testsuite to pytest, we eliminate the "self" god-object which infests unittest-based testsuites. Also, we get pretty asserts and such.

I'm also changing the main run_line helper to return the CliRunner result object, not just its output, in case we ever want to directly inspect the result. Usage sites just pull the `output` property back out for now.

This nicely consolidates the notion of automatic cleanup into an `autoclean` autouse fixture which scans some fixture lists for content to delete.


sirosen:
Discussed self-merging things with Jason. We agreed that I should "within reason". I'm self-reviewing this work anyway since it's old enough that I've lost context.

PR #474, opened 2019-08-26
external checksum checksum algorithm and user

aaschaer:
Adds `--external-checksum` and `--checksum-algorithm` options to `globus transfer`

I didn't add `--checksum-algorithm` as a batch level argument as this seemed slightly less confusing, but that is a mismatch with the SDK and Transfer API, so I'd be fine changing it to conform with those.

### DIFF
--- a/adoc/transfer.1.adoc
+++ b/adoc/transfer.1.adoc
@@ -46,7 +46,7 @@ Lines are split on spaces, respecting quotes, and every line is treated as
 a file or directory to transfer.
 
 Lines are of the form
-[--recursive] SOURCE_PATH DEST_PATH
+[--recursive] [--external-checksum CKSM] SOURCE_PATH DEST_PATH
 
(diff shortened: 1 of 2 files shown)


sirosen:
Sorry it took so long to review this. The changes look good and behave well when tested.
(Do we have a way of finding out the algorithms supported by an endpoint? I tried some strings like `SHA256` on `go#ep1` but they all gave me errors about unsupported algorithms.)

> I didn't add `--checksum-algorithm` as a batch level argument as this seemed slightly less confusing, but that is a mismatch with the SDK and Transfer API, so I'd be fine changing it to conform with those.

I think this is fine/good as you've done it. If someone asks for us to add `--checksum-algorithm` to the batch in the future, we can.

---

The lint failures are about `black` formatting, so I'm going to do the merge manually and apply the formatting in that merge commit.

PR #485, opened 2020-03-06
Convert from LazyIdentityMap to sdk IdentityMap

sirosen:
Update to sdk v1.9.0 and start using IdentityMap, obsoleting LazyIdentityMap and removing it.

---

After SDK 1.9.0, I want to take some time to do several things like this to update before releasing a new CLI version.


sirosen:
Self-merging so that we can move forward with a new version to release.

PR #486, opened 2020-03-09
session_required_single_domain support

aaschaer:
This is the work I've been using to test end to end mapped collection flows on sandbox

I'm not super attached to the `--domain` option, but since the arguments to `session update` are already overloaded between identity names and identity ids I'm not sure there is a nice way to have them also potentially be domains?

### DIFF
--- a/globus_cli/commands/session/update.py
+++ b/globus_cli/commands/session/update.py
@@ -18,9 +18,15 @@ from globus_cli.services.auth import get_auth_client
 )
 @no_local_server_option
 @click.argument("identities", nargs=-1, required=False)
-@click.option("--domain", multiple=True, help="authenticate with a specific domain")
 @click.option(
-    "--all", is_flag=True, help="authenticate with every identity in your identity set"
+    "--domain", multiple=True,
+    help="authenticate with a specific domain requirement. Can be passed "
+         "multiple times if multiple domains are required. "
+         "Mutually exclusive with IDENTITIES and --all")
+@click.option(
+    "--all", is_flag=True,
+    help="authenticate with every identity in your identity set. Mutually "
+         "exclusive with IDENTITIES and --domain"
 )
 def session_update(identities, no_local_server, all, domain):
     """
--- a/globus_cli/commands/session/update.py
+++ b/globus_cli/commands/session/update.py
@@ -18,17 +18,19 @@ from globus_cli.services.auth import get_auth_client
 )
 @no_local_server_option
 @click.argument("identities", nargs=-1, required=False)
+@click.option("--domain", multiple=True, help="authenticate with a specific domain")
 @click.option(
     "--all", is_flag=True, help="authenticate with every identity in your identity set"
 )
-def session_update(identities, no_local_server, all):
+def session_update(identities, no_local_server, all, domain):
     """
     Update your current CLI auth session by authenticating
     with specific identities.
     """
+    session_params = {"session_message": "Authenticate to update your CLI session."}
 
-    if (not (identities or all)) or (identities and all):
-        raise click.UsageError("Either give one or more IDENTITIES or use --all")
+    if not (bool(identities) ^ all ^ bool(domain)):
+        raise click.UsageError("Either give IDENTITIES or use --all or --domain")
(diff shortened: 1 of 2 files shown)


sirosen:
I was experimenting with something similar. In terms of making it a single list param, `IDENTITIES_OR_DOMAINS`, we _can_ do it.

<details><summary>discriminator for ids vs names vs domains</summary>

```python
ids, names, domains = [], [], []
for param in identities_or_domains:
    try:
        uuid.UUID(param)
        ids.append(param)
    except ValueError:
        if "@" in param:
            names.append(param)
        else:
            domains.append(param)
```
</details>

However, I think it might actually be better for users the way you currently have it, with `--domain`, because it's easier to explain to users that they're mutually exclusive.

(Also, I don't think it matters too much and you already have it written this way.)

globus_cli/commands/session/update.py
```diff
@@ -18,24 +18,27 @@
 )
 @no_local_server_option
 @click.argument("identities", nargs=-1, required=False)
+@click.option("--domain", multiple=True, help="authenticate with a specific domain")
```
ml-globus: The behavior right now is that the user has to have authenticated with any of the listed domains in the current session, which I don't think this conveys.

Maybe also add a note that it can be used in conjunction with `prompt=login` to force re-auth if a matching identity is already in session?

There's also some questions now about what exactly the behavior should be in Auth -- see long thread in #testing this morning and the brief discussion during the GCS sync call earlier today.

sirosen: The CLI always sends `prompt=login`.
The helptext will need to change. It needs to state that it's a `multiple=True` opt and that it's mutex with the other options. I think we should figure out what the interface is before we try to finalize text though.

I thought the desired behavior was now clear -- I'll send email to clear up whatever is left.

globus_cli/commands/session/update.py
```diff
@@ -67,13 +70,19 @@ def session_update(identities, no_local_server, all):
                         break
                 else:
                     click.echo("No such identity {}".format(val), err=True)
-                    click.get_current_context().exit(1)
+                    click.get_current_cont / ext().exit(1)
```
sirosen, line 73: Looks like this is an accident/typo?

sirosen:
@aaschaer, as I'm touching the CLI again, I noticed that I didn't provide the feedback on this which I meant to.
I think the option as implemented is good, but I'd like the helptext to note that it's a `multiple=True` option and that it's mutually exclusive with `--all` and providing identities. (Mutex is less important to me than the `multiple=True`, since you'll discover it's exclusive if you try to use it with the others.)

I'd also like to get the testsuite passing. I'm not sure what was up at the time of the last build, but it looks like it might have been part of the click 7.1 release changing some of the error messages. So a rebase might build cleanly now (other than the typo I noted, which is making flake8 unhappy).

globus_cli/commands/session/update.py
aaschaer: Ah yeah, I'll fix that

sirosen:
This looks good to me. I'd like to test it against an endpoint with the domains requirement to exercise the new excepthook logic, but I'm not sure where to find / how to build one.
However, assuming tests pass I want to hold off on merging at least for this week. I want us to release v1.13.0 with just #490 and no other changes.

aaschaer:
I'm not sure if the auth piece required for this is even on prod yet, so holding off at least a week sounds like a good idea.

If you want to test on sandbox, any of Jason Alt's non ha mapped collections should work for testing.

sirosen:
Revisiting this. Right now, it looks like the main issue in the build is that `black` reports things as unclean.
`make lint` should run it and apply the relevant changes. I can do this myself and merge -- I'm just not 100% clear on when we want this change to go out. We should discuss at some point (perhaps in a couple of weeks?).

sirosen:
I'd like to get this cleaned up and included in the next release, but I'm going to push out v1.14.0 without it.

sirosen:
I'm working on a rework of this to use a new param type for identities which optionally allows for domain names as well.

aaschaer:
closing this in favor of #509

PR #490, opened 2020-04-28
Add endpoint set-subscription-id command

sirosen:
Adds a new command which calls the PUT subscription ID API directly.
This is useful when working with High Assurance endpoints and with GCSv5.x endpoints where the constraints around managing subscription IDs have become more complex. There are scenarios like a GCSv5 endpoint where the endpoint client is the owner/admin, and a user needs to set the endpoint to be managed under a subscription before being granted admin rights on the endpoint.

Until now, support for this method of setting the subscription ID was largely academic -- this is the new way to do it, but there was no pressing need to enable this method. However, especially with the new HA GCP setup flow and potential new versions of GCS, it's becoming more important to have this available in the CLI.

While I don't expect much usage, this is an important escape-valve, especially in case the webapp doesn't correctly decide on whether to use endpoint update or this PUT subscription API.

### DIFF
--- a/globus_cli/commands/endpoint/commands.py
+++ b/globus_cli/commands/endpoint/commands.py
@@ -9,6 +9,9 @@ from globus_cli.commands.endpoint.permission import permission_command
 from globus_cli.commands.endpoint.role import role_command
 from globus_cli.commands.endpoint.search import endpoint_search
 from globus_cli.commands.endpoint.server import server_command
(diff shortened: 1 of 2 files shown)


sirosen:
Okay, I'm going to push ahead and merge it on the strength of that review, in that case. I'll talk with both of you about doing a release in a few minutes.

PR #493, opened 2020-04-30
Add tox to development requirements?

jaswilli:
It was dropped in aa5368121868a8eeb7d08815c08ab6103234062d but I'm not sure why--can we bring it back or was there an issue?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,6 +50,7 @@ setup(
             # not just py2: py3 versions of mock don't all have the same
             # interface!
             'mock==2.0.0;python_version<"3.6"',
+            "tox<4",
         ],
     },
     entry_points={"console_scripts": ["globus = globus_cli:main"]},


sirosen:
I think I pulled out `tox` because it's not really part of the testing requirements -- it's much more like `make`.
The way I refactored the Makefile, to assume that `tox` is available, is much simpler than what we had before.

I'm fine with adding it back, but then I think the Makefile needs to revert too -- to use `.venv/bin/tox` instead of `tox` -- or else I don't see the point of re-adding it.

sirosen:
To clarify: I don't think that requiring `tox` to already be installed was key to the reduction in complexity. I just think in the course of "tear everything out and put it back together", I found it easier to just assume people had it available.

jaswilli:
Okay. I'm totally fine with needing to have tox available outside the scope of this project. I was just confused because all of the developer workflows require it (`make lint`, `make test`, `make release`).

PR #498, opened 2020-09-15
Replace `adoc/` with an automatically built `reference/` doc dir

sirosen:
I don't know when the last time was that we looked at the `adoc/` dir and tried to ensure that it's all correct, and yet that content is the source for docs site pages.
Furthermore, we often have a synchrony issue in which we release new commands or features, but don't update the `adoc/` content.

The better approach, I think, is to generate the reference documentation pages from source, and embed any documentation we want -- as strings -- in the command structure.

My first approach was to try using `click-man` for this. However, we'd have to fork it to fix several issues, and to add support for many customizations. Since `click-man` is MIT-licensed, I took a careful look at the source and used it as inspiration for a python3.6+ generator script which produces asciidoc reference pages.
`reference/_generate.py` clocks in at under 200 logical lines (250ish raw lines) and does everything we need.

In order to support this, however, it had to be possible to pass certain information to the generator script from the command source. For example, to add `adoc_examples` as an attribute of the commands, containing the `EXAMPLES` doc section, we need a new attribute on the command objects.
To this end, this replaces the standard `click.Command` with a subclass -- one which merely pulls out `adoc_examples` and several related options from kwargs before invoking `click.Command` initialization. Now our commands can store whatever extra strings we want.
`reference/_generate.py` pulls out these data to build the reference pages.

Some helptext had to be altered slightly, so that it would render correctly as asciidoc. For example
```
Batch Input
---

some text
```
does not render as a heading in the way we want, so it was changed to
```
=== Batch Input

some text
```

Although more extensive modifications were tempting, I only did this where I spotted obvious issues.

There are some additions to the header area if you look at the index page generation in `reference/_generate.py`. These are actually being applied after-the-fact today by the docs site build, and I'd like to get rid of the special casing we're doing there.

I was able to wire this into a modified build of the docs site and show that the end result renders in a sensible way:
transfer description | transfer examples
---|---
image | image

------

I mentioned in Slack that my longer-term plan is to convert all of this into a github actions build which commits changes in a separate branch every release. I think that would be *awesome*, but I'm not going to try to make that leap in a single bound. This will be wired into the doc build by means of `make reference` or `tox -e reference`, and that will be a good first-step.


sirosen:
Oh, minor note: even though I'd normally squash down these changes into fewer commits, I think given the nature of this work it's useful to see the ways in which I refined things over time. For example, the custom command class in 347e15e is a bit smaller than what I ended up with.

sirosen:
I added a GitHub workflow to this which does the doc build (on the `autodoc-build` branch only for now -- we can switch to `master`) and uploads the result as a workflow artifact.
I think the workflow is all we need to hook into this from the docs site build.

sirosen:
@aaschaer, I've removed Jason from this, so it's just a matter of whether or not you're comfortable with the direction that I'm taking things.

I'd like to try to get this integrated and included in a release which also includes the `session consent` work. That way, we should get autodoc of the new command, which I think would be great.

If you have time to do a thorough review, that's great. But if not, perhaps we should discuss? I want to get a few things like this done, and we may be comfortable with me doing it without review in some cases.

PR #499, opened 2020-09-15
Add handling for the data_access scope

sirosen:
This is the "minimum viable" version of this feature. Unfortunately, since I have limited time for this right now, I've chosen to eschew any automated tests. (I'd like to add something, but it's something of a matter of "when".)

Add a new command, `globus session consent`. Similar to `globus session update`, this is a specialized login flow which will allows users to modify their state in Auth. Specifically, in this case it will allow users to consent to additional scopes. Various names for this action are possible, e.g. "globus consent add". However, we already have the `globus session` command group for commands which manage your authentications, so this logically belongs there (whether or not the underlying implementation is treated as such notwithstanding -- it is not clear whether or not this feature is billed as part of "sessions" because it is not yet documented).

Add a new handler, like the session hook, which handles ConsentRequired errors. The hook will exit(255) if the error format is
not valid, but otherwise it looks almost identical to the session hook. The hard failure is appropriate in this case as there is no
graceful degradation we can perform if there is no required_scopes list. Other than that, it's almost a copy-paste-replace of the session hook, in this case directing users to invoke `globus session consent` with the specified scopes.

The ConsentRequired hook will single-quote the scopes for presentation to the user. Otherwise, it is easily possible to output a command which cannot run correctly as stated due to glob-expansion and other substitutions by the shell which we would expect to see on scope names.

One area that needed repair in order to handle these changes is the redirection handler for authentications. Because, in this case, we may only receive back a partial set of tokens, the handler cannot error out if the `auth.globus.org` tokens are not present. Instead, the handler must handle this possibility gracefully. In order to do so, the refactor "queues up" data to store, conditionally, and then processes that data. This flow of control should easily adapt to future storage methods or additional resource servers.

Here's a sample of this in action (against a Sandbox v5.4 Endpoint):
image

### DIFF
--- a/globus_cli/commands/session/commands.py
+++ b/globus_cli/commands/session/commands.py
@@ -1,3 +1,4 @@
+from globus_cli.commands.session.consent import session_consent
 from globus_cli.commands.session.show import session_show
 from globus_cli.commands.session.update import session_update
 from globus_cli.parsing import group
@@ -11,3 +12,4 @@ def session_command():
 # commands
 session_command.add_command(session_update)
 session_command.add_command(session_show)
+session_command.add_command(session_consent)
--- a/globus_cli/commands/session/consent.py
+++ b/globus_cli/commands/session/consent.py
@@ -1,0 +1,44 @@
+import click
+
+from globus_cli.helpers import (
+    do_link_auth_flow,
+    do_local_server_auth_flow,
+    is_remote_session,
+)
+from globus_cli.parsing import command, no_local_server_option
+
+
+@command(
+    "consent",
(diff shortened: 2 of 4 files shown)


aaschaer (approved):
Code looks good to me and I was able to get access to a non HA mapped collection just fine with this.

I'd go ahead and merge if it weren't for the Travis test failures. I don't think its related to anything in this PR, but would rather you confirm that as well @sirosen

sirosen:
Yeah, it looks like Travis itself choked. I'll rerun and if it passes, I'll merge after that.

PR #500, opened 2020-10-08
Replace travis with GH Actions

sirosen:
Remove Travis config, add a GH Actions workflow for testing.

As a workaround for test flakiness, if a test run fails, wait 60s and try again. This can result in slow failures if the testsuite is really broken by a change, but it is less likely to give us false negatives -- which we just end up re-running anyway.


aaschaer (approved):
Looks good to me!

PR #502, opened 2020-12-05
Update dependencies

jaswilli:
* Drop python 3.5, add python 3.9
* Update dev dependencies and add packages that are used for linting into the development extras install target


sirosen:
Thanks! This needed doing and I definitely forgot.

PR #503, opened 2021-01-13
Limit upgrades to <2.0 when on py2

sirosen:
First, change the `get_versions()` helper to return a 3-tuple: `upgrade_target, latest, current`

This allows for the upgrade target to differ from the actual latest version. Specifically, when running under python2, filter the versions we get back such that upgrade_target is the maximal version satisfying `< 2.0`.

Then modify the `globus version` and `globus update` commands to handle this new output.

In the case of the version printing, the logic is somewhat refactored so that we can more easily distinguish the various conditions that determine what we say. If the latest and upgrade_target versions match, the behavior is completely unchanged. But if they differ, we print a generic message about how the latest CLI version isn't supported by your runtime and you should reinstall.

In the update command, we just add a check for `current == upgrade_target` after we've checked for `current == latest`.
If it matches, we print a warning message and exit(0).

Once we release this change, we can recommend that users update to the latest version of the CLI so that `globus update` will not unintentionally update them to a version incompatible with a py2 runtime. We can work out exactly what we want to say later, after this releases.

### DIFF
--- a/globus_cli/commands/update.py
+++ b/globus_cli/commands/update.py
@@ -118,33 +118,36 @@ def update_command(yes, development, development_version):
         # latest version supported by the current platform/runtime . So exit 0 with a
         # warning
         if current == upgrade_target:
-            click.echo(
-                click.style(
-                    (
-                        "You are running the latest version ({}) supported by your "
-                        "runtime environment.\n"
-                        "However, a newer version ({}) is available.\n"
-                        "You need to uninstall and reinstall the CLI in order to "
-                        "update."
-                    ).format(upgrade_target, latest),
-                    fg="yellow",
+            if upgrade_target != latest:
+                click.echo(
+                    click.style(
+                        (
+                            "You are running the latest version ({}) supported by "
+                            "python2.\n"
+                            "However, a newer version ({}) is available.\n"
+                            "If you would like to update to the latest version, you "
+                            "need to reinstall globus-cli using python3."
+                        ).format(upgrade_target, latest),
+                        fg="yellow",
+                    )
                 )
-            )
             return
 
-        # if we're up to date (or ahead, meaning a dev version was installed)
-        # then prompt before continuing, respecting `--yes`
-        else:
-            click.echo(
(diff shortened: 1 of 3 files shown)


jaswilli:
I'm playing around with this and when I arrange things to get this:

```
$ globus update
You are already running version 1.14.0
The latest version for you is   3.0
Continue with the upgrade? [Y/n]:
```

It makes me think it might be nice to just phrase things in terms of the python version. As it reads now I'm left thinking that perhaps it's something about _me_ (my identity? etc.) that's being taken into consideration.  If it's not a lot of extra work perhaps something like:

```
$ globus update
You are running version 1.14.0 on python2
The latest version for python2 is 1.15.0
The latest version for python3 is 3.0. If you would like to install 3.0 you need to reinstall globus-cli using python3.
Continue with the upgrade to 1.15.0? [Y/n]:
```

sirosen:
If you're running python2 at this point, _isn't it really something about you?_
taps head

---

I think your suggested rephrase is better. I struggled with what to put in that text in the first place -- I didn't want to get into details in the midst of the prompt. Specifying it by version will be nicer. I'll have that done in a few.

sirosen:
I've pushed updated text for both `globus update` and `globus version`, as in
```
$ globus update
You are already running version 1.14.0
The latest version for python2 is 1.15.0
The latest version for python3 is 2.14.0.
If you would like to install version 2.14.0, you must reinstall globus-cli using python3.

Continue with the upgrade? [Y/n]: ^C


$ globus version
You are running the Globus CLI on python2

Installed version:           1.14.0
Latest version for python2:  1.15.0
Latest version for python3:  2.14.0

You should update your version of the Globus CLI.
However, your python2 runtime does not support the latest version.

You may update with the 'globus update' command, but we recommend that
you uninstall and reinstall the CLI using python3.
```

I think directly addressing py2 vs py3 (rather than being vague) is a big improvement in readability.

jaswilli:
I like it.

PR #504, opened 2021-01-14
Use status badge from GitHub Actions

jaswilli:
This switches the README status badge from Travis to GitHub Actions and changes the name of the workflow from "Test" to "build."

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,4 +1,4 @@
-name: Test
+name: build
 on:
   push:
     branches: ["master"]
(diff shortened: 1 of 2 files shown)


sirosen:
Obviously it's not important at all, but I am curious -- as I merge this -- why the name change?

jaswilli:
The name is included in the badge so I changed it so the badge stays the
same as before.

It doesn’t matter much to me. Feel free to make it whatever you want if you
have a preference.

sirosen:
I have zero preference. But I didn't know the name went into the badge; that's good to know.

PR #505, opened 2021-01-14
Update pre-commit; lint github workflow config

sirosen:
Adds the check-jsonschema hook and invokes the check-github-workflows checker. Also update the version of isort used.

Since we're doing some maintenance here, this is the first of several projects I'll do this on. I'd like to introduce this hook  to ensure that a bad workflow config can't get pushed.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,10 +1,14 @@
 repos:
 - repo: https://github.com/pre-commit/pre-commit-hooks.git
-  rev: v2.5.0
+  rev: v3.4.0
   hooks:
     - id: check-merge-conflict
     - id: trailing-whitespace
     - id: end-of-file-fixer
+- repo: https://github.com/sirosen/check-jsonschema
+  rev: 0.3.0
+  hooks:
+    - id: check-github-workflows
 - repo: https://github.com/python/black
   rev: 20.8b1
   hooks:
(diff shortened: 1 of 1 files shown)


jaswilli:
This seems to work as expected. One thing I noticed while testing is that the output is enormous because it looks like it dumps the entire schema file (example: https://gist.github.com/5e3072f0f65d9aa2edba078a68c0e6b5). I wonder if there's a way to only output the stack trace as that's where the useful information lives? Maybe there's a good reason to dump the schema?

I'm not unwilling to to merge this as-is though so if you're happy with the way it works now feel free to merge.

jaswilli (approved):

sirosen:
Yeah, improving the output is somewhere on my to-do list. I don't consider fixing it critical -- the necessary information is there -- but I don't think there's much value in showing the full schema. Since you bring it up, I'll take a look at doing that tomorrow, before merging this.

sirosen:
I had assumed that pulling this information out of the jsonschema lib was going to be a lot harder.

I did this as the output for now in v0.3.0:
```bash
$ check-jsonschema --schemafile "https://json.schemastore.org/github-workflow" .github/workflows/build.yaml
Schema validation errors were encountered.
  .github/workflows/build.yaml::jobs.test: 'runs-on' is a required property
```

If you have ideas about how to do it better, they would of course be welcome.

jaswilli:
Just surfacing the error(s) themselves seems perfect.

PR #507, opened 2021-01-19
Update dependency versions

sirosen:
sdk -> 1.10.0
jmespath -> 0.10.0
cryptography -> '<3.4'

Also update a dated comment about `mock` for python2 only.

I think this is the last thing I want to do before doing a release, maybe later today if we can. I'm taking a look at similar cleanup/dependency things in the SDK, but there's nothing in the `install_requires` for the SDK which will be changing. So I don't think we need to wait for that here.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,9 +23,9 @@ setup(
     version=version,
     packages=find_packages(exclude=["tests", "tests.*"]),
     install_requires=[
-        "globus-sdk==1.9.1",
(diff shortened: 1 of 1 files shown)


jaswilli:
>  I'm taking a look at similar cleanup/dependency things in the SDK,

There was a recent major version release of `pyjwt` which I think included a py3.6+ requirement, so we'll need to assess that and slot it into the SDK release that does the same.

PR #508, opened 2021-01-19
default disable-verify should be None

aaschaer:
While debugging some transfer behavior using the CLI I noticed that the disable_verify flag was being passed as False even when I didn't use that CLI flag.

### DIFF
--- a/globus_cli/parsing/shared_options.py
+++ b/globus_cli/parsing/shared_options.py
@@ -169,6 +169,7 @@ def endpoint_create_and_update_params(*args, **kwargs):
         )(f)
         f = click.option(
             "--disable-verify/--no-disable-verify",
+            default=None,
             is_flag=True,
             help="(Un)Set the endpoint to ignore checksum verification",
         )(f)


sirosen:
I didn't know this was an option. I take it that `disable_verify: null` is the default in Transfer when not explicitly set.

How is `disable_verify: null` different from `disable_verify: false`?
I'm wondering if we need to allow `globus endpoint update` to un-set the value, which will not be possible even with this change.

aaschaer:
Transfer won't accept a null disable_verify, and the default is False.

The issue is that we haven't overridden click's default "not passing the flag means false" behavior here like we have with the other flags so all endpoint update commands are passing along a `disable_verify: false` in their requests to Transfer

Fortunately I don't know of any use-cases where disable_verify should actually be set to True so the CLI setting it to False every update shouldn't have hurt anything - hopefully.

sirosen:
Ah! I see now, thanks for the explanation. It's been long enough that I had forgotten about the caution we need to take with boolean flags here.

As you say, it's probably been a non-issue since wanting `disable_verify: true` should be so rare, but it's still a bug. Merging.

jaswilli:
> Fortunately I don't know of any use-cases where disable_verify should actually be set to True so the CLI setting it to False every update shouldn't have hurt anything - hopefully.

As a data point, the web app doesn't expose `disable_verify` so it's highly unlikely that there are many or possibly even any endpoints at all with `disable_verify` set to true.

Issue #484, opened 2020-01-24
Check Identity ID vs Identity Name by checking for UUID, not regex


sirosen:
In [the SDK port of IdentityMap](https://github.com/globus/globus-sdk-python/pull/367), this came up as a potential issue.

Here's where the regex matching comes from:
https://github.com/globus/globus-cli/blob/50c7b4ac669b2f9f081ce9d696d602f57e3212ed/globus_cli/services/auth.py#L13

It's not certain that this is an issue -- certainly, nobody has complained thusfar -- but we should look things over and see if the same improvement can be made here.

sirosen:
Closing as this is handled in #509

sirosen:
I mixed this up with the PR, #523. However, upon rereading things, it turns out this is not exactly the same between #509 and #523 . I think I'm going to reopen.

PR #509, opened 2021-01-20
Add IdentityType param type and use it to add domain support to `session update`

sirosen:
This replaces #486 , and slightly changes the usage.
There's no `--domain` flag. Instead, you just `globus session update foo.com uchicago.edu` and that triggers the domain behavior. Mixing types, like `globus session update uchicago.edu sirosen@globus.org` results in a usage error.

`IdentityType` is a parameter type which can be flagged to accept domains or base32-encoded identity IDs.
We then use the b32-flagged variant in `get-identities` and the domain-flagged variant in `session update`.

In general, this takes any UUID-checking of "maybe identity ID, maybe username" values out of the relevant command modules and puts it in a centralized place.

Allowing the type to accept domains requires that we check for `@` being in non-UUID values. As a result, there is a very slight semantic change: a non-username, non-UUID value to `get-identities` is no longer accepted.
Previously we had
```bash
$ globus get-identities invalid  # exit(0)
NO_SUCH_IDENTITY
```
now we have
```bash
$ globus get-identities invalid  # exit(2)
Usage Error: ...
```

There's also a very slight change to the `session update` behavior. Previously, `get_identities` was called and used to resolve usernames. Now they're resolved from the `identity_set` from userinfo. That way, you get an error earlier in the process if you try to use an identity which is not in your identity set.

There's test coverage for the `session update` functionality. However, as we don't currently have tests setup to check the various excepthooks, I didn't think it was worth figuring out the wiring for all of that at the moment.


sirosen:
@aaschaer, I'll rebase in a few minutes to clean this up.
I recall that I did this after reviewing #486. I felt that we could provide a better interface by allowing the CLI to detect if an arg looks like a username or a domain.

I'll self-review as well, since I haven't touched this in months.

globus_cli/commands/session/update.py
```diff
@@ -1,25 +1,86 @@
...
+                if identity["username"] == name:
+                    identity_ids.append(identity["sub"])
+                    break
+            else:
+                click.echo("'{}' is not in your identity set".format(name), err=True)
+                click.get_current_context().exit(1)
+
+    # update session params once we have resolved usernames (if necessary)
+    if identity_ids:
+        session_params["session_required_identities"] = ",".join(identity_ids)
```
sirosen, line 64: This allows for the possibility of identity IDs which aren't in the user's identity set.

I've just checked, and attempting to authorize with an identity not in your set leads to an error. It's better to error out early with a clear message from the CLI in this case.

I think I should add such handling, which might allow for restructuring to fetch the identity set at an earlier step.

aaschaer (commented):
Code looks good minus the allowing identities not in session issue you already mentioned and I like not needing to pass `--domain`

sirosen:
> minus the allowing identities not in session issue you already mentioned

Just to clarify, I did add that in the last commit.

Are we good on this one? I'm not 100% clear; just want to make sure.

I'm going to go ahead and close #484 as a duplicate

aaschaer (approved):
I think this should be good to go now

PR #511, opened 2021-01-23
Update GitHub Actions and default branch name

sirosen:
The branch name change is master -> main . GitHub says they'll redirect downloads of tarballs after a branch rename, so the change in `update` (and the impact there) is basically nil.

This got me looking at the workflow file, and I realized that I didn't clean it up when I did the testsuite refactor. The games with retrying a flakey testsuite are no longer necessary.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,15 +1,16 @@
 name: build
 on:
   push:
-    branches: ["master"]
   pull_request:
-    branches: ["master"]
+  # build weekly at 4:00 AM UTC
+  schedule:
+    - cron: '0 4 * * 1'
 jobs:
   test:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: ["2.7", "3.5", "3.6", "3.7", "3.8"]
+        python-version: ["2.7", "3.6", "3.7", "3.8"]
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yaml
```diff
@@ -1,15 +1,16 @@
...
+  # build weekly at 4:00 AM UTC
+  schedule:
+    - cron: '0 4 * * 1'
 jobs:
   test:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: ["2.7", "3.5", "3.6", "3.7", "3.8"]
+        python-version: ["2.7", "3.6", "3.7", "3.8"]
```
sirosen, line 13: Oh, also this is "obvious". We don't support 3.5 anymore, but it was still here in the test matrix.

.github/workflows/test.yaml
```diff
@@ -1,15 +1,16 @@
 name: build
 on:
   push:
-    branches: ["master"]
   pull_request:
-    branches: ["master"]
+  # build weekly at 4:00 AM UTC
+  schedule:
+    - cron: '0 4 * * 1'
```
sirosen, line 7: We started doing this in the SDK and I see no reason not to do it here too.

PR #512, opened 2021-01-26
Add `description` to `endpoint show` text

sirosen:
And make the formatter a little bit more organized. Rather than doing tuple-unpacking in the various printers, a new FormatField class is added which can coerce a tuple to a FormatField (or passthrough any existing FormatField). This lets us keep the existing tuple-based interace but introduce fields with parameters. Namely, `description` is unusual in that text-wrapping may be desirable, so it is defined as a FormatField where `wrap_enabled=True`.

We then make a judgement while printing the data about "should wrap" to allow for non-wrapping behavior for short outputs.

Description is truncated to a maximum of 5 lines of output, newlines are respected, and the whole output is right-aligned to line up with existing fields.

The table printer needed very minor tweaks to handle FormatField instead of tuples.

When adding new tests for `endpoint show`, it turned out that our fixture data (some originally taken from docs) is out of date and doesn't have several supported fields. Simply pad out the missing fields so that tests will be in good working order.

---

I'm possibly more interested in the new tests than I am in the feature itself. I'd like to be in a state where `endpoint show` is tested and works with the fixture data.

I do _like_ the result of adding `FormatField`, however, and somewhat wish that the printing code weren't so dynamic, and that we had everything converted to a known type like this at import-time.


globus_cli/safeio/output_formatter.py
```diff
@@ -97,43 +122,57 @@ def print_unix_response(res):
...
+    indent = " " * maxlen
+    wrapper = textwrap.TextWrapper(initial_indent=indent, subsequent_indent=indent)
+    for field in fields:
+        # str in case the result is `None`
+        value = str(field(data))
+
+        # 88 char wrap based on the same rationale that `black` and `flake8-bugbear`
+        # use 88 chars (or if there's a newline)
+        # only wrap if it's enabled and detected
+        shouldwrap = field.wrap_enabled and len(value) + maxlen > 88 or "\n" in value
```
sirosen, line 136: I worried for a half-second that I had this wrong, but `or` is higher precedence than `and`. ( https://docs.python.org/3/reference/expressions.html#operator-precedence )
I can add parens if this poses a readability concern.

tests/files/api_fixtures/endpoint_operations.yaml
```diff
@@ -34,6 +34,14 @@ transfer:
...
         "organization": "My Org",
         "username": "auser",
+        "canonical_name": "auser#myserver",
+        "contact_email": null,
+        "contact_info": null,
+        "department": null,
+        "info_link": null,
+        "keywords": null,
+        "local_user_info_available": true,
+        "owner_string": "auser@globusid.org",
```
sirosen, line 44: These are the missing fields for `show`. I wasn't sure where in the documents to put them, so I put them in the same place each time, in alphabetical order.

Transfer actually sorts its output so that the order is fixed, so I might come back through here and sort these later, but I didn't want to mix that into this change.

globus_cli/commands/endpoint/show.py
```diff
@@ -22,6 +22,7 @@ def endpoint_show(endpoint_id):
     ("Display Name", "display_name"),
     ("ID", "id"),
     ("Owner", "owner_string"),
+    FormatField("Description", "description", wrap_enabled=True),
```
sirosen, line 25: Optional, but now that this is a fancy object rather than a tuple, it would be possible to write things like `FormatField("activated")` and have that mean what `("Activated", "activated")` means today.

I'd be amenable to such refinements if we like this direction. But the two-argument form will be needed at least for cases like `Owner, owner_string`.

aaschaer: I like the direction, but I'd rather we keep the two-argument form even if the display name could be inferred from the field name to keep things simple and uniform.

globus_cli/safeio/output_formatter.py
aaschaer: I generally prefer more parens over having to think about precedence, but nbd

tests/files/api_fixtures/endpoint_operations.yaml
aaschaer: This seems fine for now.
It would be nice if everything were always alphabetized Transfer side, not sure how big of a change that would be

aaschaer (approved):
Looks good to me minus maybe adding a few parens if you want to bother with that

globus_cli/commands/endpoint/show.py
sirosen: Agreed. I actually have a branch which tried doing `replace("_", " ").title()`, but it (1) doesn't save much code and (2) becomes confusing to read.

tests/files/api_fixtures/endpoint_operations.yaml
sirosen: I think Transfer does produce sorted output, but then we load-and-dump that through the python `json` module without being careful to preserve order. (I _think_ newer versions of cpython preserve order as an implementation detail, but I'm not even sure offhand.)

globus_cli/safeio/output_formatter.py
sirosen: I'll add them. Looking at this again, I don't see a reason _not_ to add the parens. The fact that I felt the need to comment on it tells us everything we need to know -- it's potentially confusing.

sirosen:
I've added the parens, so I'm going ahead with this.

PR #513, opened 2021-01-27
skip and quota support

aaschaer:
Adds --skip-source-errors and --fail-on-quota errors options to `globus transfer`
Adds --skipped-errors and a default ouput field for subtasks_skipped_errors to `globus task show`

I wrote this to depend on the SDK changes in https://github.com/globus/globus-sdk-python/pull/393
I'll see if I can write a few tests using the new framework by the time that gets merged.

### DIFF
--- a/globus_cli/commands/task/show.py
+++ b/globus_cli/commands/task/show.py
@@ -46,9 +46,24 @@ SUCCESSFULL_TRANSFER_FIELDS = [
     ("Destination Path", "destination_path"),
 ]
 
+SKIPPED_PATHS_FIELDS = [
+    ("Source Path", "source_path"),
+    ("Destination Path", "destination_path"),
+    ("Error Code", "error_code"),
+]
+
 
 def print_successful_transfers(client, task_id):
     res = client.task_successful_transfers(task_id, num_results=None)
+    formatted_print(
+        res,
+        fields=SUCCESSFULL_TRANSFER_FIELDS,
+        json_converter=iterable_response_to_dict,
+    )
+
+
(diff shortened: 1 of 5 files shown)


sirosen:
> I wrote this to depend on the SDK changes in globus/globus-sdk-python#393

From < 1 minute of looking at this, I think the tests will still all succeed, since TransferData passes through arbitrary kwargs and no tests are using the new get method.

I think at least one test which _fails_ without the SDK update is in order, to ensure that we can't ship this without the requisite update. (Not that we'd be so careless, but to codify the requirement for `globus-sdk>=X` in the tests.)

sirosen:
I'm marking this blocked until we have an SDK release with the requisite support. But I looked over everything and it looks good.

Once we have a release and the dependency version can be bumped, I'd like a test which passes with the updated SDK and fails without it (for `task show`).

aaschaer:
Added a test for --skipped-errors which is currently failing on `AttributeError: 'RetryingTransferClient' object has no attribute 'task_skipped_errors'` as expected.

Hopefully the test itself is correct other than that.

sirosen:
With the SDK update out, I'll rebase this, and probably merge. I may move that `task_id` variable into the fixture file, but otherwise don't intend to change things.

aaschaer:
Oh - I think we both pushed fixes at the same time, sorry!

sirosen:
No worries! Either fix is good.
You might be interested to see mine ( https://github.com/globus/globus-cli/commit/95bb65cc45e12763d9f0c37b987c285900fcfe12 ), but let's go with yours.

PR #514, opened 2021-01-29
Update globus-sdk-python

jaswilli:
Resolves:
https://github.com/globus/globus-cli/pull/513#issuecomment-769466743


sirosen:
I'm going to edit your comment because it looks like GitHub smart commits thinks this should close #513 . I couldn't find a way to turn off smart commits, but I think a newline will "fix" it.

PR #518, opened 2021-02-04
Add mypy to our lints, fix minor linting issues

sirosen:
Some background on motivation: I have a branch with some internal refactoring I think would be nice, and I wanted to type annotate (and, obviously, check) the new internal tooling it introduces. Having mypy linting properly in-place ahead of time seems the right way to go. Even without explicit plans to type-annotate the SDK, I'd like to introduce the same, or at least similar, config there too.

---

Introduce mypy with config to be as pedantic as we can be without causing unnecessary pain. That is, turn on all of the warnings we can which are either non-issues or easily resolved. Some settings like `check_untyped_defs` would be a huge PITA right now.

This identified three minor issues:
1. `site.USER_BASE` is actually an `Optional[str]`, so switch to `getuserbase()`
2. mypy wants a type for `version_ns` in setup.py -- `type: ignore` that one to keep complexity in setup.py down
3. `reference/_generate.py` changes the type of a variable, so do some minor name-munging to fix it

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -36,3 +36,7 @@ repos:
   hooks:
     - id: pyupgrade
       args: ["--py36-plus"]
+- repo: https://github.com/pre-commit/mirrors-mypy
+  rev: v0.800
+  hooks:
+    - id: mypy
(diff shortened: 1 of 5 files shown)


jaswilli (commented):
Do we want to add mypy to the dev requirements?

sirosen:
Oh, I guess I should.

I'm not really in the habit of depending on those, so I easily overlook them.

sirosen:
Added `mypy`. I'll make a branch with the same work for the SDK, since I'm not hearing any objection.

PR #523, opened 2021-02-26
Check identity ID with UUID, not regex

### LINKED ISSUE #484: Check Identity ID vs Identity Name by checking for UUID, not regex

sirosen:
resolves #484

### DIFF
--- a/globus_cli/services/auth.py
+++ b/globus_cli/services/auth.py
@@ -1,4 +1,4 @@
-import re
+import uuid
 
 from globus_sdk import AuthClient, RefreshTokenAuthorizer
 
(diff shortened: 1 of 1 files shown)


jaswilli:
I'm having a hard time connecting the dots. Is it okay to only check for UUIDs because the IdentityMap is resolving usernames to IDs prior to this check happening?

sirosen:
This is for get-identities ; I don't think any other commands use it. An id outputs username, and a username outputs id, so we need to accept both and detect which is which.
We've been doing it with a regex. This just switches to try-except handling around uuid parsing instead.

sirosen:
I just noticed this again and thought "didn't I do something else like this?" And I did! I'm pretty sure this is redundant with #509 . This could go in first -- they aren't mutually exclusive -- but I think #509 is the better overall approach to parsing username-or-id parameters.

jaswilli:
Yeah, I've got it now. I remembered that's what this was concerned with but the way the diff read had me confused.

I'll look at the other change before this one. I didn't want to merge that one for 1.x/2.x but I'm sure it's fine now.

sirosen:
I'm moving this forward since it's not fully redundant with #509 .

PR #524, opened 2021-02-26
Add `mutex_option_group` decorator

sirosen:
Several commands check a series of options for mutual exclusivity.
This work is always the same, in essence, and removing it to a decorator simplifies logic in some of the more complex cases (see the changes in `validate_endpoint_create_and_update_params`).

Because we're now operating in 3.6+ this introduces some type annotations, which weren't available to us in the past. We can start making strictly typed internal APIs.

In the normal cases, `mutex_option_group` will just be given the option string and will deduce the param name and correctly check for its presence. However, for unusual cases (explicit underlying param names, etc) there is a `MutexInfo` object which can be populated with the correct fields/behaviors, and could be extended further in the future if necessary.

---

I think this is a good idea, but I'm open to the idea that we shouldn't bother. I wish `click` provided this and did option inspection to consistently map option strings to param names, but I'm not willing to build that out here. I think this is good enough and still an improvement over doing these checks ad-hoc.

### DIFF
--- a/globus_cli/commands/endpoint/activate.py
+++ b/globus_cli/commands/endpoint/activate.py
@@ -7,7 +7,7 @@ from globus_cli.helpers import (
     fill_delegate_proxy_activation_requirements,
     is_remote_session,
 )
-from globus_cli.parsing import command, endpoint_id_arg
+from globus_cli.parsing import command, endpoint_id_arg, mutex_option_group
 from globus_cli.safeio import FORMAT_TEXT_RAW, formatted_print
 from globus_cli.services.transfer import activation_requirements_help_text, get_client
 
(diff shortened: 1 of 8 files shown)


aaschaer (approved):
Code looks good to me, and I like how clean the decorator looks for the simple cases.

I feel like if only `--no-managed` weren't a special case this code could be much simpler, but I suppose there's no saying that there won't be similar special cases in the future.

sirosen:
> I feel like if only `--no-managed` weren't a special case this code could be much simpler, but I suppose there's no saying that there won't be similar special cases in the future.

Yeah, I agree that there's a lot "extra" going on to make that case work smoothly.
I think a big part of the problem revolves around the way that boolean flag options `--foo/--no-foo` behave, and is maybe more general than that one weird case.

I'm looking again and not having any bright ideas about how to make it work even more smoothly. Even in that "bad" case, it's still simpler than it was before. As always, we can refine it more in the future if we think of something better.

I've done a rebase, so now I'm just waiting for the build before merging.

PR #525, opened 2021-03-12
Remove the CLI config commands

sirosen:
Pros:
- Pure negative diff with no lines added at all and tests still pass
- This won't really work if we move to a different storage mechanism (e.g. sqlite) anyway
- Config commands have unclear merits and were added on a whim
- We need to drop or change these if we want to drop configobj

Cons:
- Some user might be using it in a script and would now have to use `awk` or something


jaswilli:
Agreed, there was never really a good story about where we were going with the config command.

PR #526, opened 2021-04-02
Fix minor typos in usage statements

glentner:


sirosen:
Nice catch! Thanks for submitting a fix.

PR #527, opened 2021-06-22
Add --exclude option to transfer submit

aaschaer:
I think this is the simplest way to expose filter_rules to the cli, similar to how we only expose the name field in ls filtering.

If / when filter_rules become more complex and support more than just name and type, we could do something similar to the gcs cli and allow passing json documents for more complex use cases, but I think having --exclude <name> would still be a nice shorthand for the most common use case even then.

### DIFF
--- a/globus_cli/commands/transfer.py
+++ b/globus_cli/commands/transfer.py
@@ -193,6 +193,18 @@ fi
         "during the transfer."
     ),
 )
+@click.option(
+    "--exclude",
+    multiple=True,
+    default=None,
+    show_default=True,
+    help=(
+        "Exclude files and directories found with names that match the given "
+        "pattern in recursive transfers. Pattern may include * ? or [] for "
+        "unix style globbing. Give this option multiple times to exclude "
+        "multiple patterns."
+    ),
+)
 @click.argument(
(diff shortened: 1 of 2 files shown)


globus_cli/commands/transfer.py
```diff
@@ -398,6 +421,16 @@ def process_batch_line(dest_path, source_path, recursive, external_checksum):
             recursive=recursive,
         )
 
+    for item in transfer_data["DATA"]:
+        if item["recursive"]:
+            has_recursive_items = True
+            break
+    else:
+        has_recursive_items = False
```
sirosen, line 429: I think this is easier to read as
```python
has_recursive_items = any(item["recursive"] for item in transfer_data["DATA"])
```
as part of my general thinking that `for-else` can be confusing.

But it's not a big deal -- definitely not enough to hold up this PR.

sirosen:
Looks good and works as expected when I test it.
I agree with your call to expose `--exclude` for name filtering, and we can revisit if there's ever broader demand for CLI support for other filtering criteria.

I'm going to squash merge so that there's a single commit for me to cherry-pick onto `main`. And I can handle the release as well.

globus_cli/commands/transfer.py
aaschaer: Oh! I do agree that this is cleaner, I'll try to use that pattern in the future at the very least if we don't want to go back and use it here

PR #529, opened 2021-06-28
Bump click version to >=8.0.0.

masalim2:
Update the changed error message assertion in functional/test_jmespath_output.py

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,7 @@ setup(
     python_requires=">=3.6",
     install_requires=[
         "globus-sdk==3.0.0b2",
-        "click>=7.1.1,<8.0",
+        "click>=8.0.0,<9",
         "jmespath==0.10.0",
         "requests>=2.0.0,<3.0.0",
         # cryptography has unusual versioning and compatibility rules:
--- a/shell_completion/bash_complete.sh
+++ b/shell_completion/bash_complete.sh
@@ -1,2 +1,0 @@
-# temporary shim -- remove me after release of shell completion
-eval "$(globus --bash-completer)"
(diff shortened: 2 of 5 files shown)


masalim2:
I'm wondering if there are any challenges in bumping the `click` dependency from v7.x to 8.x.  Click 8 fixes some typing errors that arise in my project with newer versions of `mypy` (over 0.900); however, this creates a dependency conflict for projects that directly rely on both `globus-cli` and `click`.

This PR naively bumps the click version; running `make lint` and `make test` on my Macbook shows no errors.

sirosen:
Because the globus-cli is an application (rather than a library), we keep the version constraints bounded. So I think this should be `>=8.0,<9`.
That said, thank you for finding which tests break with this upgrade and applying the fix! 😄 

I'm going to review the `click` 8.0 changelog just to make sure there are no red flags before merging.

Just FYI, it will probably be a little while before we release this update: we're currently working on version 3 of the globus-sdk and intend to work on and release the next major version of the globus-cli (also version 3) after that is done.

sirosen:
I feel mostly comfortable after reading the changelog. Major things I need to test under click v8 (not covered by the testsuite):

- Completion was rewritten again. Make sure `globus --completer` and `globus --zsh-completer` are correct. (Probably need to update our included shell functions.)
- Double-check that our reference asciidoc generation script runs successfully.
- Many changes to helptext generation and options. Make sure all commands run `--help` successfully.

Those latter two could be added to automated testing, but not necessarily something to solve right now.

sirosen:
It took me (much) longer than expected to get back to this and test the results. The completion options did, indeed, need a little adjustment. But they work with the update I just applied. Assuming the build succeeds (I'd be surprised if it fails), I'll merge in a few minutes.

PR #534, opened 2021-08-13
Add requires_login decorator

aaschaer:
For https://app.clubhouse.io/globus/story/9065/add-a-first-draft-of-cli-loginmanager-component-providing-requires-login-decorator

As we discussed this doesn't fully handle gcs login logic, but it should provide the framework needed for that via the pass_manager option and the LoginManager class.

Two questions came up while I was applying the decorator to our commands, which both kinda boil down to will the user always have tokens scoped for auth in this new incremental login world
1. Many commands only need auth tokens if they are given a username they need to resolve to an id, do we want to make auth a hard requirement for those?
2. whoami and session commands introspect the auth token, but presumably they could use a token for any resource server, would we want to do something like introspect the first token found in storage?

### DIFF
--- a/src/globus_cli/helpers/auth_flows.py
+++ b/src/globus_cli/helpers/auth_flows.py
@@ -66,9 +66,9 @@ def do_local_server_auth_flow(session_params=None):
         auth_client.oauth2_start_flow(
             refresh_tokens=True, redirect_uri=redirect_uri, requested_scopes=SCOPES
         )
-        additional_params = {"prompt": "login"}
-        additional_params.update(session_params)
-        url = auth_client.oauth2_get_authorize_url(additional_params=additional_params)
+        query_params = {"prompt": "login"}
+        query_params.update(session_params)
+        url = auth_client.oauth2_get_authorize_url(query_params=query_params)
 
         # open web-browser for user to log in, get auth code
         webbrowser.open(url, new=1)
--- a/src/globus_cli/login_manager.py
+++ b/src/globus_cli/login_manager.py
@@ -47,10 +47,7 @@ def requires_login(*args: str, pass_manager: bool = False):
             manager = LoginManager()
 
             # determine the set of resource servers missing logins
-            missing_servers = set()
-            for server_name in resource_servers:
-                if not manager.has_login(server_name):
-                    missing_servers.add(server_name)
+            missing_servers = {s for s in resource_servers if not manager.has_login(s)}
 
             # if we are missing logins, assemble error text
             # text is slightly different for 1, 2, or 3+ missing servers
--- a/src/globus_cli/login_manager.py
+++ b/src/globus_cli/login_manager.py
@@ -1,0 +1,93 @@
+import functools
+
+import click
+
+from .tokenstore import token_storage_adapter
+
+
+class LoginManager:
+    def __init__(self):
+        self._token_storage = token_storage_adapter()
+
+    def has_login(self, resource_server: str):
+        """
+        Determines if the user has a refresh token for the given
+        resource server
+        """
+        tokens = self._token_storage.get_token_data(resource_server)
+        if tokens is None or "refresh_token" not in tokens:
+            return False
+
+        return True
+
+
+def requires_login(*args: str, pass_manager: bool = False):
+    """
+    Command decorator for specifying a resource server that the user must have
+    tokens for in order to run the command.
+
+    Simple usage for commands that have static resource needs: simply list all
+    needed resource servers as args:
+
+    @requries_login("auth.globus.org")
+
+    @requires_login("auth.globus.org", "transfer.api.globus.org")
+
+    Usage for commands which have dynamic resource servers depending
+    on the arguments passed to the command (e.g. commands for the GCS API)
+
+    @requies_login(pass_manager=True)
+    def command(login_manager, endpoint_id)
+
+        login_manager.<do the thing>(endpoint_id)
+
+    """
+    resource_servers = args
+
+    def inner(func):
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            manager = LoginManager()
+
+            # determine the set of resource servers missing logins
+            missing_servers = set()
+            for server_name in resource_servers:
+                if not manager.has_login(server_name):
+                    missing_servers.add(server_name)
+
+            # if we are missing logins, assemble error text
+            # text is slightly different for 1, 2, or 3+ missing servers
+            if missing_servers:
+
+                if len(missing_servers) == 1:
+                    plural_string = ""
+                    server_string = missing_servers.pop()
+
+                elif len(missing_servers) == 2:
+                    plural_string = "s"
+                    server_string = "{} and {}".format(
+                        missing_servers.pop(), missing_servers.pop()
+                    )
+
+                else:
+                    plural_string = "s"
+                    single_server = missing_servers.pop()
+                    server_string = ", ".join(missing_servers) + ", and {}".format(
+                        single_server
+                    )
+
+                raise click.ClickException(
+                    "Missing login{} for {}, please run 'globus login'".format(
+                        plural_string, server_string
+                    )
+                )
+
+            # if pass_manager is True, pass it as an additional positional arg
+            if pass_manager:
+                return func(*args, manager, **kwargs)
+            else:
+                return func(*args, **kwargs)
+
+        return wrapper
+
+    return inner
--- a/src/globus_cli/login_manager.py
+++ b/src/globus_cli/login_manager.py
@@ -15,10 +15,7 @@ class LoginManager:
         resource server
         """
         tokens = self._token_storage.get_token_data(resource_server)
-        if tokens is None or "refresh_token" not in tokens:
-            return False
-
-        return True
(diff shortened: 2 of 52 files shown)


src/globus_cli/login_manager.py
```diff
@@ -0,0 +1,93 @@
...
+    def has_login(self, resource_server: str):
+        """
+        Determines if the user has a refresh token for the given
+        resource server
+        """
+        tokens = self._token_storage.get_token_data(resource_server)
+        if tokens is None or "refresh_token" not in tokens:
+            return False
+
+        return True
```
rudyardrichter, line 21:
```suggestion
        return tokens is not None and "refresh_token" in tokens
```

src/globus_cli/login_manager.py
```diff
@@ -0,0 +1,93 @@
...
+    def inner(func):
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            manager = LoginManager()
+
+            # determine the set of resource servers missing logins
+            missing_servers = set()
+            for server_name in resource_servers:
+                if not manager.has_login(server_name):
+                    missing_servers.add(server_name)
```
rudyardrichter, line 56:
```suggestion
            missing_servers = {s for s in resource_servers if not manager.has_login(s)}
```

src/globus_cli/login_manager.py
```diff
@@ -0,0 +1,93 @@
...
+
+            # if we are missing logins, assemble error text
+            # text is slightly different for 1, 2, or 3+ missing servers
+            if missing_servers:
+
+                if len(missing_servers) == 1:
+                    plural_string = ""
+                    server_string = missing_servers.pop()
```
rudyardrichter, line 64: I might factor the formatting-focused bits out into a utils module.

tests/unit/test_login_manager.py
```diff
@@ -0,0 +1,116 @@
...
+@patch("globus_cli.tokenstore.token_storage_adapter")
+def test_requires_login_success(mock_get_adapter):
+    mock_get_adapter._instance.get_token_data = mock_get_tokens
+
+    # single server
+    @requires_login("a.globus.org")
+    def dummy_command():
+        return True
+
+    assert dummy_command()
```
rudyardrichter, line 34: Some would, perhaps rightfully, dispute the legibility of this, but I thought I would offer the suggestion since it would shorten a lot of tests.
```suggestion
    assert requires_login("a.globus.org")(lambda: True)()
```

tests/unit/test_login_manager.py
```diff
@@ -0,0 +1,116 @@
...
+
+    @requires_login(pass_manager=True)
+    def dummy_command(manager):
+
+        assert manager.has_login("a.globus.org")
+        assert not manager.has_login("c.globus.org")
+
+        return True
+
+    assert dummy_command
```
rudyardrichter, line 116: This line is asserting just the function, not a call thereof?

aaschaer: Ah! good catch!

src/globus_cli/login_manager.py
aaschaer: That's fair, it got more complicated than I intended. Not sure how much we even care about listing all the missing servers. Might just be better to say the user needs to log in again until we change the login command to accept arbitrary resource servers.

tests/unit/test_login_manager.py
aaschaer: I know they're logically equivalent, but I wanted to make sure the @ decorator usage was functioning as intended since that's how we'll be calling this in code

sirosen:
To answer the high-level questions: I think we always want to require and use an Auth token where applicable.

You're right that we could use other tokens in various cases, and the Auth token is not actually special in any way. However, we're already aware that there are odd corner cases in which your various tokens belong to different identities. We'll try to avoid such situations, but, knowing that it's possible, I think it makes sense to always use the same token.

So, the answers are

> Many commands only need auth tokens if they are given a username they need to resolve to an id, do we want to make auth a hard requirement for those?

Yes, let's require the Auth token.

> whoami and session commands introspect the auth token, but presumably they could use a token for any resource server, would we want to do something like introspect the first token found in storage?

No, let's make it an error if no Auth token can be found. The only way it could be absent while other tokens are present is some undesirable manipulation of the token DB anyway.

src/globus_cli/services/auth.py
```diff
@@ -8,6 +8,9 @@
 # what qualifies as a valid Identity Name?
 _IDENTITY_NAME_REGEX = r"^[a-zA-Z0-9]+.*@[a-zA-z0-9-]+\..*[a-zA-Z]+$"
 
+# constant for token and login management
+AUTH_RESOURCE_SERVER = "auth.globus.org"
```
sirosen, line 12: This placement for this constant is fine, but I would prefer that we define these, and the scopes for these resource servers, as part of the login manager (either in that module or in that class, either is fine). The same applies to the transfer server and scopes.

My reason is that the login manager is responsible for "handling login activity" and I expect it will at some point begin to replace parts of the `globus login` command. At that point, I think it will be convenient and more intuitive for that component to define all of the static scopes and resource servers, as well as the utilities for handling dynamic requirements.

This doesn't need to move now; it's a minor issue and we can address it when we work on more dynamic behaviors.

src/globus_cli/login_manager.py
sirosen: I think we've done something like this for output in the CLI before, FWIW.
If you're making a utility, it might be worth taking a glance around to see if we can have a helper which does this.

I think `pluralize(value: List[str]) -> Tuple[bool, str]` which returns `(is_plural, list_formatted_string)` might be good. Where to apply that `"s"` in the general case could get tricky.

On a related note, I think it's fine to print `Missing login(s) for ...` to dodge that part of the issue.

tests/unit/test_login_manager.py
sirosen: This is the sort of thing which I probably wouldn't bother to call out one way or another in most cases. Either syntax is fine and they mean the same thing.

If you had half a dozen variants of the same decorator being applied to `lambda: True` equivalents, inlining them like this would possibly make the test more readable, but that's not the case here.

I bias slightly more towards readability than brevity, so I'm going to mark this as resolved.

src/globus_cli/helpers/auth_flows.py
```diff
@@ -66,9 +66,9 @@ def do_local_server_auth_flow(session_params=None):
         auth_client.oauth2_start_flow(
             refresh_tokens=True, redirect_uri=redirect_uri, requested_scopes=SCOPES
         )
-        additional_params = {"prompt": "login"}
-        additional_params.update(session_params)
-        url = auth_client.oauth2_get_authorize_url(additional_params=additional_params)
+        query_params = {"prompt": "login"}
+        query_params.update(session_params)
+        url = auth_client.oauth2_get_authorize_url(query_params=query_params)
```
sirosen, line 71: Good catch! Sorry I didn't see and fix this when updating to 3.0b2! 😅

src/globus_cli/login_manager.py
```diff
@@ -0,0 +1,87 @@
...
+    def command(login_manager, endpoint_id)
+
+        login_manager.<do the thing>(endpoint_id)
+
+    """
+    resource_servers = args
+
+    def inner(func):
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
```
sirosen, line 46: Question: should this whole wrapper function be functionality of the LoginManager itself? i.e. `LoginManager.check_required_logins(*args)` with a body nearly identical to this function.

My thought is that in the cases for `pass_manager=True`, we will want to presumably trigger this same functionality off of the `LoginManager`.

We can save any potential refactor for later if you prefer, however.

src/globus_cli/login_manager.py
```diff
@@ -0,0 +1,87 @@
...
+    def has_login(self, resource_server: str):
+        """
+        Determines if the user has a refresh token for the given
+        resource server
+        """
+        tokens = self._token_storage.get_token_data(resource_server)
+        return tokens is not None and "refresh_token" in tokens
+
+
+def requires_login(*args: str, pass_manager: bool = False):
```
sirosen, line 21:
```suggestion
def requires_login(*resource_servers: str, pass_manager: bool = False):
```

Can we make this change (and then remove the assignment below)?
I think it will make this function much more readable. When I first saw it, I thought you were passing _these_ args to the decorated function, because you're overriding the same name (`args`).

src/globus_cli/login_manager.py
```diff
@@ -0,0 +1,87 @@
...
+    Usage for commands which have dynamic resource servers depending
+    on the arguments passed to the command (e.g. commands for the GCS API)
+
+    @requies_login(pass_manager=True)
+    def command(login_manager, endpoint_id)
+
+        login_manager.<do the thing>(endpoint_id)
+
+    """
+    resource_servers = args
```
sirosen, line 42:
```suggestion
```

sirosen:
The current bits of feedback are non-blocking suggestions (could/should rewrite this way or that, not "this is incorrect").
I'm going to move forward and merge this in its current state and we can continue to address any outstanding issues or improvements in follow-up PRs.

PR #536, opened 2021-08-17
Add string listifying helpers

sirosen:
Used by login_manager and mutex options.

I thought we had needed this formatting elsewhere in the CLI but these were the only cases I could find.

### DIFF
--- a/src/globus_cli/utils.py
+++ b/src/globus_cli/utils.py
@@ -1,0 +1,34 @@
+from typing import Dict
+
+
+def format_list_of_words(first: str, *rest: str):
+    if not rest:
+        return first
+    if len(rest) == 1:
+        return f"{first} and {rest[0]}"
+    return ", ".join([first] + list(rest[:-1])) + f", and {rest[-1]}"
+
+
+def format_plural_str(formatstr: str, pluralizable: Dict[str, str], use_plural: bool):
+    """
+    Format text with singular or plural forms of words. Use the singular forms as
+    keys in the format string.
+
+    Usage:
+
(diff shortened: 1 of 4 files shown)


src/globus_cli/utils.py
```diff
@@ -0,0 +1,34 @@
+from typing import Dict
+
+
+def format_list_of_words(word: str, *other: str):
```
rudyardrichter, line 4: `word`/`other` was a slightly confusing nomenclature to me, I think something `first`/`rest` might have been more apparent (since from the context it's already clear they're words)

src/globus_cli/utils.py
```diff
@@ -0,0 +1,34 @@
...
+    ...         len(command_list) == 1
+    ...     )
+    ... )
+    >>> print("  " + "\n  ".join(command_list))
+    """
+    argdict = {
+        singular: plural if use_plural else singular
+        for singular, plural in pluralizable.items()
+    }
+    return formatstr.format(**argdict)
```
rudyardrichter, line 34: Could "onelinerize" `argdict` into here.

src/globus_cli/utils.py
sirosen: Sure, `first/rest` works for me.

src/globus_cli/utils.py
sirosen: Yeah, I considered... But I think this is easier to read? Opinions on this probably vary.

sirosen:
Amended to reword `word,other` to `first,rest`.

PR #541, opened 2021-08-24
Reorganize login manager logic to encapsulate login flows and static requirements

sirosen:
This is based on #540 (it is part of the purpose behind that cleanup).
Marking it blocked until that merges. I'll rebase as necessary.

---

This is a "simple" reorg of the login flow logic as follows:
- requires_login, AUTH_RS, and TRANSFER_RS are attributes of the LoginManager class (classmethod and class vars)
- Dispatch between local webserver and text link login flows is always done by LoginManager.run_login_flow. Parameters cover session params and messages printed during the flow.
- LoginManager.is_logged_in_static() provides "check login" functionality for the static token requirements. It includes a token validation callout for each refresh token
- LoginManager._TEST_MODE is a bool (class var) which disables the validation call to simplify testing

The result is simpler imports -- most things that need some kind of login flow can just ask for `from globus_cli.login_manager import LoginManager` and rely on the class attributes for the relevant logic and data.

This should set the stage for us to have extra flags on `run_login_flow()` for passing new required scopes and resource servers. `globus login` would be able to take `--gcs <ID>` to expand to a `manage_collections` scope and pass this through to the `LoginManager`.

### DIFF
--- a/src/globus_cli/commands/bookmark/create.py
+++ b/src/globus_cli/commands/bookmark/create.py
@@ -1,8 +1,8 @@
 import click
 
-from globus_cli.login_manager import requires_login
+from globus_cli.login_manager import LoginManager
 from globus_cli.parsing import ENDPOINT_PLUS_REQPATH, command
-from globus_cli.services.transfer import TRANSFER_RESOURCE_SERVER, get_client
+from globus_cli.services.transfer import get_client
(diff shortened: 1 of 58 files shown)


sirosen:
As I have other work blocked on this, I'm moving it forward.

PR #543, opened 2021-08-27
Add explicit control for CLI interactivity

sirosen:
The CLI currently uses a heuristic for determining an interactive session of "is PS1 set?"

This has proven solid and accurate for many years, in spite of there not being any standard method for determining if a TTY is interactive.

This PR introduces an explicit flag which can be used to force interactive mode on or off (strtobool conversion). Set `GLOBUS_CLI_INTERACTIVE=0` and you won't get any interactive behaviors.

As the 3.x line will potentially introduce more auto-prompt behaviors, this will be useful for testing, and may prove important for some users as well.

### DIFF
--- a/src/globus_cli/termio/context.py
+++ b/src/globus_cli/termio/context.py
@@ -1,5 +1,6 @@
 import os
 import sys
+from distutils.util import strtobool
 
 import click
 
@@ -68,5 +69,11 @@ def err_is_terminal():
     return sys.stderr.isatty()
 
 
-def term_is_interactive():
+def term_is_interactive() -> bool:
+    explicit_val = os.getenv("GLOBUS_CLI_INTERACTIVE")
+    if explicit_val is not None:
+        try:
+            return strtobool(explicit_val.lower())
+        except ValueError:
+            pass
     return os.getenv("PS1") is not None
--- a/tests/unit/test_termio.py
+++ b/tests/unit/test_termio.py
@@ -1,0 +1,32 @@
+import os
+
+import pytest
+
+from globus_cli.termio import term_is_interactive
+
+
+@pytest.mark.parametrize(
+    "ps1, force_flag, expect",
+    [
+        (None, None, False),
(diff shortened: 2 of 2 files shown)


src/globus_cli/termio/context.py
```diff
@@ -68,5 +69,11 @@ def err_is_terminal():
...
 
 
-def term_is_interactive():
+def term_is_interactive() -> bool:
+    explicit_val = os.getenv("GLOBUS_CLI_INTERACTIVE")
+    if explicit_val is not None:
+        try:
+            return strtobool(explicit_val.lower())
+        except ValueError:
+            pass
```
aaschaer, line 78: I wonder if we shouldn't just silently ignore misconfigured values here? I could imagine being confused by mismatching behavior because I accidentally typed "treu" or something.

aaschaer (approved):
Left one comment, but its pretty minor so I think this is probably good to go as is

src/globus_cli/termio/context.py
sirosen: I wouldn't be opposed, as I agree that the result could be surprising. But I think this would require deeper restructuring.

We don't force this function to evaluate on every command, or at the start of a command.
So if `GLOBUS_CLI_INTERACTIVE=treu`, that would error... whenever `term_is_interactive()` gets called.
I don't like the idea that we could be halfway through something and then blow up.

If we added a step to our commands which always called some checker function, `validate_settings()`, then we could force `treu` to fail always, on all commands.

Let's write this up as an improvement, so we can look at how to attach a hook to the start of command execution (but not when `resilient_parsing` is set, so we don't break completion!).
This is the only variable which I know would be a candidate today, but we can look for `os.getenv` and `os.environ` usage to check for more.

sirosen:
I agree with your comment, but I think the Right Solution:tm: will require a bit more work. So let's do that as a later refinement of this.

PR #544, opened 2021-09-01
Add login `--gcs` opt and `collection delete`

sirosen:
This PR needs tests added, but I have tested it manually. For that reason, it's being opened as a draft.

One note for future work: we may want each `globus` command to keep a "cache" or "table" of `endpoint_id -> endpoint_document` to improve the responsiveness of GCS-related commands (which might otherwise do multiple callouts). I intend to add something like `EndpointMap` with `__getitem__` defined in terms of a `TransferClient.get_endpoint` call. But I'm explicitly avoiding that optimization for the first pass at this.

Also, the error messages are not great right now, and we'll need to refine this.

---

Update the LoginManager to support dynamic requirements in service of this new flag.

The LoginManager feature isn't really GCS-specific, but makes the login manager capable of handling newly-added, dynamic requirements. Primarily, it adds storage for dynamic requirements and replaces use of the STATIC_SCOPES constant with a property that wraps dynamic and static requirements.

The login command is extended with a `--gcs {id}` option which requests the relevant manage_collections scope from Auth. To do so, it adds the GCS to the dynamic requirements of the login manager, then proceeds with the login flow as normal.

These features are exercised in the addition of a `globus collection delete` command which operates via a GCSClient to delete a collection.


sirosen:
In the course of adding more extensive testing, I've done some work on the error messages (not perfect, but a start), so that `collection delete ${gcp_id}` will print `endpoint delete`.

PR #546, opened 2021-09-02
Change filter value handling to ignore lists of empty values

rudyardrichter:
- Use default or `""` if a list of filter values is all empty strings (`["", ""]` should translate to `""` not `","`)
- Extend `task_list` test to check that completion/request time filters are empty instead of `","`

### DIFF
--- a/src/globus_cli/commands/task/list.py
+++ b/src/globus_cli/commands/task/list.py
@@ -166,6 +166,8 @@ def task_list(
 
     def _process_filterval(prefix, value, default=None):
         if value:
+            if isinstance(value, list) and not any(value):
+                return default or ""
             if isinstance(value, str):
                 return f"{prefix}:{value}/"
             return "{}:{}/".format(prefix, ",".join(str(x) for x in value))
--- a/tests/functional/test_task_list.py
+++ b/tests/functional/test_task_list.py
@@ -1,5 +1,14 @@
+import responses
+
+
 def test_task_list_success(run_line, load_api_fixtures):
     load_api_fixtures("task_list.yaml")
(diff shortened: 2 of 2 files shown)


src/globus_cli/commands/task/list.py
```diff
@@ -166,6 +166,8 @@ def task_list(
 
     def _process_filterval(prefix, value, default=None):
         if value:
+            if isinstance(value, list) and not any(value):
+                return default or ""
```
rudyardrichter, line 170: This would also work:
```suggestion
        if value and (not isinstance(value, list) or any(value)):
```
but (of course) I think this version is moderately more readable.

tests/functional/test_task_list.py
```diff
@@ -1,5 +1,16 @@
...
     assert "SUCCEEDED" in result.output
     assert "TRANSFER" in result.output
+    # check that empty filters aren't passed through
+    filters = dict(
+        x.split(":") for x in responses.calls[0].request.params["filter"].split("/")
+    )
+    if "completion_time" in filters:
+        assert filters["completion_time"] == "", filters
+    if "request_time" in filters:
+        assert filters["request_time"] == "", filters
```
sirosen, line 16: We know in the unadorned `task list` case that nothing is passed. The fact that this is conditional is looking a little odd to me.

Can we simplify it down to this?
```suggestion
assert "completion_time" not in filters
assert "request_time" not in filters
```

If not, I just want to know why.

rudyardrichter: Yeah, this was maybe excessively test-driven, I just wanted to start with something that would fail. Fixed.

PR #547, opened 2021-09-03
Change --batch flag to take - to specify stdin

### LINKED ISSUE #222: globus transfer --batch: accept filename
Currently globus transfer --batch is defined to take the list of input files on stdin.  Rather than forcing stdin, why not define it as "--batch INPUTFILE", where INPUTFILE can be "-" to mean stdin?  That seems somewhat more flexible and easier, if my file list is in a file.

rudyardrichter:
This would change the behavior of `--batch` to resolve #222: it would require a filename argument, from which it reads source/dest arguments, and to instead use stdin to pass these one specifies the `-` value.

### DIFF
--- a/tests/functional/test_task_submit.py
+++ b/tests/functional/test_task_submit.py
@@ -1,4 +1,6 @@
 import json
+import os
+import tempfile
 
 
 def test_exclude(run_line, load_api_fixtures, go_ep1_id, go_ep2_id):
@@ -45,9 +47,26 @@ def test_exlude_recursive(run_line, load_api_fixtures, go_ep1_id, go_ep2_id):
     # batch
     batch_input = "abc /def\n"
     result = run_line(
-        "globus transfer --exclude *.txt --batch "
+        "globus transfer --exclude *.txt --batch - "
         "{}:/ {}:/".format(go_ep1_id, go_ep1_id),
         stdin=batch_input,
         assert_exit_code=2,
     )
-    assert expected_error in result.stderr
+
+    with tempfile.NamedTemporaryFile(mode="r+b", buffering=0, delete=False) as temp:
+        temp_name = temp.name
+        temp.write(bytes(batch_input, "utf-8"))
+        result = run_line(
+            [
+                "globus",
+                "transfer",
+                "--exclude",
+                "*.txt",
+                "--batch",
+                temp_name,
+                f"{go_ep1_id}:/",
+                "{go_ep1_id}:/",
+            ],
+            assert_exit_code=2,
+        )
+    os.unlink(temp_name)
--- a/tests/functional/test_task_submit.py
+++ b/tests/functional/test_task_submit.py
@@ -1,6 +1,4 @@
 import json
-import os
-import tempfile
 
 
 def test_exclude(run_line, load_api_fixtures, go_ep1_id, go_ep2_id):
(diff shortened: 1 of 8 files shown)


tests/functional/test_task_submit.py
```diff
@@ -45,7 +45,7 @@ def test_exlude_recursive(run_line, load_api_fixtures, go_ep1_id, go_ep2_id):
     # batch
     batch_input = "abc /def\n"
     result = run_line(
-        "globus transfer --exclude *.txt --batch "
+        "globus transfer --exclude *.txt --batch - "
```
sirosen, line 48: Although it's testing a click behavior, in some sense, I think we should also add a variant of this test which writes `abc /def\n` to a temp file and runs `--batch` on that file.
That way, we have an explicit test for the new behavior.

sirosen (commented):
This is great. I think `click.File` does exactly what we want, which saves us some legwork!

I'd like to see a new test which confirms the new behavior, which I noted inline.

I also think the helptext for the `transfer` and `delete` commands (the docstrings on those functions, which becomes the main helptext) needs slight adjustment. They mention `--batch` reading from stdin, which is no longer the only case.

rudyardrichter:
I'm going to have to do some more digging before this is actually mergeable, there seems to be something odd happening between click's `File` and batch values for Windows filepaths:
```
E           Exception: CliTest run_line exit_code assertion failed!
E           Line:
E           globus transfer -F json --batch C:\Users\RUNNER~1\AppData\Local\Temp\tmpels9pcwf --dry-run ddb59aef-6d04-11e5-ba46-22000b92c6ec ddb59af0-6d04-11e5-ba46-22000b92c6ec
E           exited with 2 when expecting 0
E           stdout:
E           
E           stderr:
E           Usage: globus transfer [OPTIONS] SOURCE_ENDPOINT_ID[:SOURCE_PATH]
E                                  DEST_ENDPOINT_ID[:DEST_PATH]
E           
E           Error: Invalid value for '--batch': 'C:UsersRUNNER~1AppDataLocalTemptmpels9pcwf': No such file or directory
```
I see there are other issues with `tempfile` on Windows, but this seems to be a quoting issue. I thought initially the POSIX setting for `shlex.split` could be the culprit, but 89d06d4 doesn't seem to have fixed it. I think we want to keep that change anyways?

sirosen:
> I think we want to keep that change anyways?

I don't think so. `shlex` in non-POSIX mode has some behaviors which users are likely to find surprising, like quote character preservation.
It also means that we'd be treating the same input file differently on different platforms, which means that if you move the same "batchfile" around, you could see different behavior from the same command.

I don't think that's where the error is coming from, in any case.
The testsuite has that `run_line` fixture which wraps a click `CliRunner`. There, we're doing [`shlex.split`](https://github.com/globus/globus-cli/blob/f35b947b322f992398b5be08e393c5fed7bee586/tests/conftest.py#L133) on the input. I would bet that's what's going wrong.

One easy option would be to make `run_line` take a string or list of strings:
```diff
-        args = shlex.split(line)
+        args = shlex.split(line) if isinstance(line, str) else line
```

Switching the test to pass a list of strings would skip that call entirely.

Or we could toggle POSIX mode in `run_line` based on platform. It's not user-facing (to my thoughts above about not wanting to use it in the batch processing) and we're emulating CLI parsing on that platform, so it seems appropriate.


If I get to it, I'll try this in a branch on my fork, and tweak the workflow to just run two jobs (one linux, one windows). That should let me narrow this down and figure out where those backslashes are without waiting too much for extra jobs.

rudyardrichter:
@sirosen you could take another look now; the tests work and I've cleaned up the commit history. `posix=False` caused issues in several other tests, so I went with passing a list to `run_line`.

rudyardrichter:
Ok, I take that back, the Windows test is still fragile for some reason...

rudyardrichter:
By some magic, pytest's `tmp_file` works. Thanks a ton @sirosen. I broke up the tests a bit anyhow so they're a bit more organized. Should be good to go now.

sirosen:
I bet we could read the `pytest` code for that fixture and see a lot of Windows-handling magic if we get curious.
This looks good with the restructure. I'm going ahead with the merge.

Aside: I think we should start putting together a "v3 Changes" doc to put on the docs site, like the SDK Upgrading doc. This would be part of it. I'll start something in this repo, and we can move it to the docs site and publish later.

PR #548, opened 2021-09-09
Add "experimental" static shell completers

sirosen:
This is a draft, but it's almost complete. The main thing that's missing is some level of testing.
Maybe under CI we can run `shellcheck` on the generated output, but that's pretty weak... Notes below on doing more.

---

Without removing the dynamic completion support which we get from `click`, this adds new custom completion which is built completely in the relevant shells.

In bash, this is relatively unsophisticated. The key idea is that each option can be identified by the command path + a space + the option string. This lets a bash associative array built a few (huge!) tables. A simple completer can then crawl those tables to drive the behavior.

In zsh, the completion system is much more capable, and can do the line parsing for us (!). There's one very important subtlety which is called out in inline comments, but most of this is smooth sailing.
I've compacted the result a little to reduce the sheer number of lines, but it's still quite legible.

The entire new completion system is an addition, not a replacement. For now, `globus --experimental-completer` is your way in. Just `eval $(globus --experimental-completer)` in zshrc or bashrc, and you'll have it in your shell.

Pros:
- The new completion is, especially compared with dynamic python completion, lightning fast
- Supports mutex option specs in zsh and other fancy features

Cons:
- More work to maintain (e.g. if `click` makes changes in v9.0)
- I'm still unsure how to test the results in an automated fashion (hence this being marked draft)

Other:
- Independent from changes to `click` completion (which seems to change every release)
- The entire output is readable/auditable for the zsh case, and somewhat (eeehhhhh... maybe not so readable) for the bash case

---

Testing bash completion seems somewhat doable, since we can seed the relevant variables and run in `bash`.
Testing zsh completion is... _way_ harder. I'm not sure what I can work out on this front, but I'll be thinking about this.


sirosen:
I'm going to close out this experiment. I'd like to achieve faster completions within python, not by offloading the process like this, so that we continue to benefit from `click`'s work and support.

PR #549, opened 2021-09-13
Prevent account switch on incremental auth

rudyardrichter:
- Modify `exchange_code_and_store` to issue an error if the tokens fetched don't match what the user was previously authenticated as
- Shuffle around test fixtures to support adding a test for the above, running a login flow

### DIFF
--- a/src/globus_cli/login_manager/auth_flows.py
+++ b/src/globus_cli/login_manager/auth_flows.py
@@ -5,7 +5,7 @@ import click
 from .local_server import LocalServerError, start_local_server
 from .tokenstore import internal_auth_client, token_storage_adapter
 
-_STORE_CONFIG_SUB_NAME = "auth_user_data"
+_STORE_CONFIG_USERINFO = "auth_user_data"
 
 
 def do_link_auth_flow(scopes, *, session_params=None):
@@ -97,7 +97,7 @@ def exchange_code_and_store(auth_client, auth_code):
     adapter = token_storage_adapter()
     tkn = auth_client.oauth2_exchange_code_for_tokens(auth_code)
     sub_new = tkn.decode_id_token()["sub"]
-    auth_user_data = adapter.read_config(_STORE_CONFIG_SUB_NAME)
+    auth_user_data = adapter.read_config(_STORE_CONFIG_USERINFO)
     if auth_user_data and sub_new != auth_user_data.get("sub"):
         auth_client.oauth2_revoke_token(tkn["access_token"])
         auth_client.oauth2_revoke_token(tkn["refresh_token"])
@@ -109,5 +109,5 @@ def exchange_code_and_store(auth_client, auth_code):
         )
         click.get_current_context().exit(1)
     if not auth_user_data:
-        adapter.store_config(_STORE_CONFIG_SUB_NAME, {"sub": sub_new})
+        adapter.store_config(_STORE_CONFIG_USERINFO, {"sub": sub_new})
     adapter.store(tkn)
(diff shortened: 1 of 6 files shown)


src/globus_cli/login_manager/auth_flows.py
```diff
@@ -87,7 +89,25 @@ def exchange_code_and_store(auth_client, auth_code):
...
+    as, so to secure incremental auth flows, if the new tokens don't match the previous
+    identity we revoke them and instruct the user to logout before continuing.
     """
     adapter = token_storage_adapter()
     tkn = auth_client.oauth2_exchange_code_for_tokens(auth_code)
+    sub_new = tkn.decode_id_token()["sub"]
+    auth_user_data = adapter.read_config(_STORE_CONFIG_SUB_NAME)
+    if auth_user_data and sub_new != auth_user_data.get("sub"):
+        auth_client.oauth2_revoke_token(tkn["access_token"])
+        auth_client.oauth2_revoke_token(tkn["refresh_token"])
```
sirosen, line 103: The token response will contain multiple new tokens, so we should make sure to revoke them all in this step.

I'm not sure what we should do if revocation fails (e.g. network goes down at this exact moment).
I think maybe -- and I'm not sure about this -- we want to wrap this in a try except block and print a warning if that happens?

sirosen (commented):
The testing looks good. It's got to patch/mock many parts carefully, but that's the nature of the beast.

There's one minor correctness issue and then it's good to go.

Question: do we want to call the variable for the config key `SUB`, or should we maybe call it something like `USERINFO`? We might add other fields in the future if we need to do so. (I'm considering whether or not we should make `whoami` use it.)

src/globus_cli/login_manager/auth_flows.py
rudyardrichter: Thanks. For the first part, `.by_resource_server` is a comprehensive enumeration right?
```python
for rs, tokens in tkn.by_resource_server.items():
    auth_client.oauth2_revoke_token(tokens["access_token"])
    auth_client.oauth2_revoke_token(tokens["refresh_token"])
```

> I'm not sure what we should do if revocation fails
I wouldn't be opposed to a try/except to catch such errors, log it, and just ask the user to try again. Do you think that sort of thing makes sense within `.oauth2_revoke_token` or at call sites like here?

rudyardrichter:
I had `USERINFO` originally and went to `SUB` to be more specific, but especially if you see other use for it let's go with the former.

src/globus_cli/login_manager/auth_flows.py
sirosen: Yep, `by_resource_server` will get them all. 👍

> ... just ask the user to try again ...

The trouble is that the error is happening while we're handling a usage error to begin with.
So we don't really want them to try the exact same operation again -- and even if they did, we'd get new and different tokens back.

We *could* put another blob into the token storage DB, but that's dedicating much more effort to this scenario than I think is appropriate.

It's not necessarily a problem if the CLI hard-errors in this case on failing revocation. The user would get a slightly surprising error message, but maybe that's acceptable.

Perhaps the best course of action is to use a `try-finally` construct (no `except` block) and put the error message echo in the `finally`. That way, we're guaranteed to show our desired message, even if the revocation fails. And revocation failure would show the underlying error.

sirosen: Looking at the `try-finally`, I know it's weird but I _think_ it's the right thing to do.
We can definitely revisit if we have some clear idea about improved handling in the future.

I'm going to mark this resolved.

sirosen (approved):

sirosen:
@rudyardrichter, I think it might be nice to squash-merge this particular change, but creating a merge commit would also be fine if there's work in here you want to preserve. I'm leaving it unmerged so that you can make the final judgement about squashing.

rudyardrichter:
Thanks for checking, I think you can go ahead and squash-merge—I was about to rebase-squash everything anyhow.

PR #550, opened 2021-09-15
Add group list command

aaschaer:
Adds groups service to login and group  list command to use it

One thing we might want to confirm is the fields, I'm calling `enforce_session` "High Assurance" but maybe that's incorrect. And we could potentially do some parsing of the role sub-documents to give information there

### DIFF
--- a/src/globus_cli/commands/group/list.py
+++ b/src/globus_cli/commands/group/list.py
@@ -12,12 +12,26 @@ def group_list():
 
     groups = client.get_my_groups()
 
+    def _format_session_enforcement(res):
+        if res.get("enforce_session"):
+            return "strict"
+        else:
+            return "not strict"
+
+    def _parse_roles(res):
+        roles = set()
+        for membership in res["my_memberships"]:
+            roles.add(membership["role"])
+
+        return ",".join(sorted(roles))
+
     formatted_print(
         groups,
         fields=[
             ("Group ID", "id"),
             ("Name", "name"),
             ("Type", "group_type"),
-            ("High Assurance", "enforce_session"),
+            ("Session Enforcement", _format_session_enforcement),
+            ("Roles", _parse_roles),
         ],
     )
--- a/src/globus_cli/commands/group/list.py
+++ b/src/globus_cli/commands/group/list.py
@@ -1,0 +1,23 @@
+from globus_cli.login_manager import LoginManager
+from globus_cli.parsing import command
+from globus_cli.services.groups import get_groups_client
+from globus_cli.termio import formatted_print
+
+
+@command("list", short_help="List groups you belong to")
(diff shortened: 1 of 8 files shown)


src/globus_cli/commands/group/list.py
```diff
@@ -0,0 +1,23 @@
...
+
+    groups = client.get_my_groups()
+
+    formatted_print(
+        groups,
+        fields=[
+            ("Group ID", "id"),
+            ("Name", "name"),
+            ("Type", "group_type"),
+            ("High Assurance", "enforce_session"),
```
sirosen, line 21: Although I would think `High Assurance` is a fine/good name in isolation, after looking at the webapp I think we should change this to `Session Enforcement` with values of `strict` and `not strict`. I think that's just
```python
lambda x: "strict" if x["enforce_session"] else "not strict"
```

Here's how the webapp presents this info to a group admin:
image


While I'd love for us to be consistent across services and call this `High Assurance` _everywhere_, we can't be consistent both with the webapp and with the GCP/GCS context. Given that choice, I think that using a different name for this in the CLI vs webapp will create more confusion than having a different name between Groups and Collections.

tests/functional/test_group.py
```diff
@@ -0,0 +1,17 @@
...
+    data = load_api_fixtures("groups.yaml")
+
+    group1_id = data["metadata"]["group1_id"]
+    group2_id = data["metadata"]["group2_id"]
+    group1_name = data["metadata"]["group1_name"]
+    group2_name = data["metadata"]["group2_name"]
+
+    result = run_line("globus group list")
+
+    assert group1_id in result.output
```
sirosen, line 14: I wonder... Would it be useful to have a regex matching helper for tests like this?
Something like
```python
# some "matcher fixture"
assert matcher(
    r"^Group ID:\s+(.*)$", # a regex with capturing groups
    result.output, # input string
    groups=[group1_id]  # values compared against the captured groups, in order
)
```

sirosen (commented):
> we could potentially do some parsing of the role sub-documents to give information there

I think we should do this.

Specifically, I think a field `("Roles", _comma_delimited_roles_names)` would do the trick.
Just the `"role"` field is sufficient. I don't think the average user cares very much which of their identities is associated with a specific group.

tests/functional/test_group.py
aaschaer: Hmmm, yeah that might be a good idea. I mostly looked at the output from running the command to make sure it looked good, but if something changed and messed it up this test likely wouldn't catch it

sirosen: I'll queue it up for later. It should be pretty easy to write something.

sirosen (approved):

PR #551, opened 2021-09-17
Add "Endpointish" command aliasing, associated refactors

sirosen:
`Endpointish` is an object which primarily wraps a `TransferClient.get_endpoint` call (storing the result in the `Endpointish.data` dict).

However, it also offers the following features:
- assert the type of the endpointish
- get the host endpoint ID for a collection, implicitly asserting that the type is a collection type

The errors raised on a failed assertion are a new breed of globus-cli-specific exceptions, `WrongEndpointType` errors.
The excepthook module needed a lot of rejiggering to accept this, as it had been written under the assumption that we'd mostly be using it on API errors. However, it turns out to be a good way of handling these, and decoupling the printing-and-presentation from the checking-and-erroring.
I ended up moving the excepthook out of `parsing` in the course of resolving a dependency cycle I had introduced. Although it's not essential to this change, as it turns out, I think the result is better (there's too much stuff in `globus_cli.parsing` which is not related to parsing...).

Two other little moves came in with this:
- `EndpointType` is now part of `globus_cli.endpointish` (which is the only place where it's used)
- introduce `globus_cli.constants` to hold `EXPLICIT_NULL`, which fixes the reliance of `globus_cli.services.transfer` on `globus_cli.parsing` (which is what led to the cycle)

I considered fussing around with moving more stuff into `constants.py`, and I would like to do so, but it is not related to this work.

Once `Endpointish` is established as the way to get-and-check an endpoint document, the following changes follow:
- Make `Endpointish` errors understand how to suggest a more appropriate command to use by mapping command names, which can be fetched from `click.Context.command_path`. This data is `SHOULD_USE_MAP`
- Have the excepthook for a `WrongEndpointType` error pull the `should_use` value, as well as the actual and expected types of the endpointish
- `globus endpoint delete` will now instantiate an Endpointish and check it with `assert_is_traditional_endpoint()` (We can rename later. "traditional endpoint" was the best name I could think up for now.)

New tests check the resulting behavior for `globus endpoint delete`. The existing tests for `collection delete` verify the rest of this functionality.

One unrelated change is included which I don't consider worth reverting: excepthooks previously printed to stdout, but they should "obviously" print to stderr. In the course of retooling the hooks, I fixed this.


src/globus_cli/exception_handling/registry.py
```diff
@@ -0,0 +1,44 @@
+import functools
+
+import click
+import globus_sdk
+
+from globus_cli.parsing.command_state import CommandState
+
+_REGISTERED_HOOKS = []
```
rudyardrichter, line 8: Type annotation

src/globus_cli/exception_handling/registry.py
```diff
@@ -0,0 +1,44 @@
...
+                    ctx.exit(mapping[exception.http_status])
+            ctx.exit(1)
+
+        _REGISTERED_HOOKS.append((wrapped, error_class, condition))
+        return wrapped
+
+    return inner_decorator
+
+
+def find_handler(exception):
```
rudyardrichter, line 37: Type annotation

src/globus_cli/exception_handling/registry.py
```diff
@@ -0,0 +1,44 @@
...
+
+import click
+import globus_sdk
+
+from globus_cli.parsing.command_state import CommandState
+
+_REGISTERED_HOOKS = []
+
+
+def error_handler(*, error_class=None, condition=None):
```
rudyardrichter, line 11: Annotation(s)

rudyardrichter (changes requested):

src/globus_cli/commands/collection/delete.py
```diff
@@ -13,14 +13,8 @@ def collection_delete(login_manager, *, collection_id):
...
-    endpoint_id = get_collection_endpoint_id(
-        collection_id,
-        {
-            EndpointType.GCP: "globus endpoint delete",
-            EndpointType.SHARE: "globus endpoint delete",
-            EndpointType.NON_GCSV5_ENDPOINT: "globus endpoint delete",
-        },
-    )
+    epish = Endpointish(collection_id)
+    endpoint_id = epish.get_collection_endpoint_id()
```
rudyardrichter, line 17:
```suggestion
    endpoint_id = Endpointish(collection_id).get_collection_endpoint_id()
```

src/globus_cli/commands/endpoint/delete.py
```diff
@@ -23,5 +24,8 @@ def endpoint_delete(endpoint_id):
     endpoints that are hosted on it.
     """
     client = get_client()
+    epish = Endpointish(endpoint_id, transfer_client=client)
+    epish.assert_is_traditional_endpoint()
```
rudyardrichter, line 28:
```suggestion
    Endpointish(endpoint_id, transfer_client=client).assert_is_traditional_endpoint()
```

sirosen:
@rudyardrichter, if you can make time to review 61740cc9b493e5728a1dd56331ef679f24094aca, I'd appreciate it.
The other changes are minor and as we discussed, but that one in particular involves some messy bits. Turns out that annotating those hooks wasn't quite trivial. But it's done now.🤷

sirosen:
I'm going to go ahead and merge this. It's necessary for other work.

PR #555, opened 2021-09-22
Validate environment variables before running commands

rudyardrichter:
- Add a `GlobusCommandEnvChecks` command subclass to validate env vars
- Use the above to reject invalid `GLOBUS_CLI_INTERACTIVE` values

### DIFF
--- a/src/globus_cli/parsing/commands.py
+++ b/src/globus_cli/parsing/commands.py
@@ -10,6 +10,8 @@ import sys
 
 import click
 
+from globus_cli.termio import env_interactive
+
 from .excepthook import custom_except_hook
 from .shared_options import common_options
 from .shell_completion import print_completer_option
@@ -46,6 +48,19 @@ for more details."""
                 AUTOMATIC_ACTIVATION=self.AUTOMATIC_ACTIVATION_HELPTEXT
             )
         super().__init__(*args, **kwargs)
+
+
+class GlobusCommandEnvChecks(GlobusCommand):
+    def invoke(self, ctx):
+        try:
+            env_interactive()
+        except ValueError as e:
+            click.echo(
+                f"couldn't parse GLOBUS_CLI_INTERACTIVE environment variable: {e}",
+                err=True,
+            )
+            click.get_current_context().exit(1)
+        return super().invoke(ctx)
 
 
 class GlobusCommandGroup(click.Group):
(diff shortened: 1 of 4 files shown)


src/globus_cli/parsing/commands.py
```diff
@@ -48,6 +50,19 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
 
+class GlobusCommandEnvChecks(GlobusCommand):
+    def invoke(self, ctx):
+        try:
+            env_interactive()
+        except ValueError as e:
+            click.echo(
+                f"couldn't parse GLOBUS_CLI_INTERACTIVE environment variable: {e}",
```
rudyardrichter, line 59: I can picture wanting to apply this to other things, in which case the error message might need to be more generic, or `env_interactive` might want to raise its own custom error, but this seemed an acceptable first pass.

src/globus_cli/parsing/commands.py
```diff
@@ -108,8 +123,9 @@ def command(*args, **kwargs):
     def _inner_decorator(func):
         if "help" not in kwargs:
             kwargs["help"] = func.__doc__
-
         if "cls" not in kwargs:
+            kwargs["cls"] = GlobusCommandEnvChecks
+        if kwargs.get("skip_env_checks", False) is True:
```
sirosen, line 128: I don't know that we use the `cls` argument anywhere, but this makes `cls=Foo, skip_env_checks=True` behave surprisingly.

sirosen (approved):
I think this is good. There's one oddity I noted, but it's about a hypothetical usage, not one that I think we'd ever really want.

Do you want to make any change there, or perhaps just a note in a comment? Or just leave it as is?

src/globus_cli/parsing/commands.py
sirosen: Yeah, we can refine it over time if necessary.

e.g. I almost suggested that we put the env var definition in a dedicated module. But that's not worth it for only one value.

src/globus_cli/parsing/commands.py
rudyardrichter: Fair point. I updated to (I think) a more sensible behavior, and added a note.

sirosen:
Looks good to me with the changes. I'm merging.

PR #558, opened 2021-09-24
Add 'collection update' command

sirosen:
The work is split into two commits.

First is a copy of parsing utilities in gcs-cli which will be needed for this.
These include a few nullable types, CommaDelimitedList, and the StringPrefixMapper class + the JSON-related subclass thereof.
It also gives `EXPLICIT_NULL` a repr, which is nice when unit testing things.

The second commit implements `collection update` + makes `endpoint update` redirect to it as appropriate.
Not all of the options from gcs-cli are supported yet, as some will be problematic without doing more work in the SDK to support this.
There's also some messy logic around options that cannot be used with Guest Collections. See the code surrounding `UsageError("Use of incompatible options with Guest Collection.")` in the `collection update` body. Ideally, this would be better coordinated somehow. I'm considering swapping it out for `inspect` usage on the collection class signatures, but I'm not decided on anything yet, so this is left as-is for now.

All of the new code in `src/globus_cli/parsing/param_types/` has been fairly thoroughly tested in gcs-cli at this point. Review should focus on the `collection update` implementation itself for the most part.

### DIFF
--- a/src/globus_cli/commands/collection/__init__.py
+++ b/src/globus_cli/commands/collection/__init__.py
@@ -2,6 +2,7 @@ from globus_cli.parsing import group
 
 from .delete import collection_delete
 from .show import collection_show
+from .update import collection_update
 
 
 @group("collection")
@@ -12,3 +13,4 @@ def collection_command():
 # commands
 collection_command.add_command(collection_delete)
 collection_command.add_command(collection_show)
+collection_command.add_command(collection_update)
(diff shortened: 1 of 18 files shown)


sirosen:
I've done some final review and made a change to use the results of `inspect.signature(globus_sdk.GuestCollection)` in `globus collection update` for incompatible option detection.
Assuming it passes CI, I'll merge.

PR #561, opened 2021-09-28
Add collection list command

rudyardrichter:
- Add `collection list` command
  - Support for `--include-private-policies`, `--filter`, `--mapped-collection-id`
- Add GCS `/collections` API test fixtures

### DIFF
--- a/src/globus_cli/commands/collection/list.py
+++ b/src/globus_cli/commands/collection/list.py
@@ -24,7 +24,7 @@ class ChoiceSlugified(click.Choice):
     def convert(self, value, param, ctx):
         if value is None:
             return None
-        return super().convert(value, param, ctx).replace("_", "-")
+        return super().convert(value.replace("_", "-"), param, ctx).replace("-", "_")
 
 
 @command("list", short_help="List all Collections on an Endpoint")
--- a/src/globus_cli/commands/collection/list.py
+++ b/src/globus_cli/commands/collection/list.py
@@ -1,0 +1,116 @@
+import click
+
+from globus_cli.login_manager import LoginManager
+from globus_cli.parsing import command
+from globus_cli.principal_resolver import default_identity_id_resolver
+from globus_cli.services.gcs import get_gcs_client
+from globus_cli.termio import FORMAT_TEXT_TABLE, formatted_print
+
+STANDARD_FIELDS = [
+    ("ID", "id"),
+    ("Display Name", "display_name"),
+    ("Owner", default_identity_id_resolver.field),
+    ("Collection Type", "collection_type"),
+    ("Storage Gateway ID", "storage_gateway_id"),
+]
+
+
+class ChoiceSlugified(click.Choice):
+    """
+    Allow either hyphens or underscores, e.g. both 'mapped-collections' or
+    'mapped_collections'
+    """
+
+    def convert(self, value, param, ctx):
+        if value is None:
+            return None
+        value = value.replace("_", "-")
+        return super().convert(value, param, ctx)
+
+
+def hyphen_to_underscore(ctx, param, value):
+    if not value:
+        return None
+    return [x.replace("-", "_") for x in value]
+
+
+@command("list", short_help="List all Collections on an Endpoint")
+@click.argument(
+    "endpoint_id",
+    metavar="ENDPOINT_ID",
+)
+@click.option(
+    "--filter",
+    "filters",
+    multiple=True,
+    # choices are shown with hyphens, but the command will receive them with underscores
+    type=ChoiceSlugified(
+        ["mapped-collections", "guest-collections", "managed-by-me", "created-by-me"]
+    ),
+    callback=hyphen_to_underscore,
+    help="""\
+Filter results to one of the specified categories of collections. Can be applied
+multiple times. Note that mutually exclusive filters are allowed and will find no
+results.
+
+The filters are as follows
+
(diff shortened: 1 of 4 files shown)


src/globus_cli/commands/collection/list.py
```diff
@@ -0,0 +1,115 @@
...
+    """
+    client = get_gcs_client(endpoint_id)
+    params = {}
+    if mapped_collection_id:
+        params["mapped_collection_id"] = mapped_collection_id
+    if filters:
+        params["filter"] = ",".join(filters)
+    if include_private_policies:
+        params["include"] = "private_policies"
+    res = client.get_collection_list(**params)
```
sirosen, line 114: Of these parameters, [we only have `include` as a keyword arg in the SDK](https://globus-sdk-python.readthedocs.io/en/stable/services/gcs.html#globus_sdk.GCSClient.get_collection_list).

You'll have to put the others in `query_params`, and we can add support for them in SDK v3.1.

src/globus_cli/commands/collection/list.py
```diff
@@ -0,0 +1,115 @@
...
+        if value is None:
+            return None
+        value = value.replace("_", "-")
+        return super().convert(value, param, ctx)
+
+
+def hyphen_to_underscore(ctx, param, value):
+    if not value:
+        return None
+    return [x.replace("-", "_") for x in value]
```
sirosen, line 34: Optional refactor: you could fold this into `ChoiceSlugified` and skip using a callback altogether.
Just `super().convert(value, param, ctx).replace("-", "_")` above, and this goes away.

I think I like that better (using a custom param type and no callback vs both), but I don't mind if you prefer to leave it as-is.

sirosen (changes requested):
There are some minor issues which need resolution, but this overall seems on-point. I like the custom type for handling hyphen-to-underscore munging, but think it should encompass the callback as well.

src/globus_cli/commands/collection/list.py
rudyardrichter: I kind of suspected it wouldn't be necessary to have both, thanks for the suggestion.

tests/files/api_fixtures/collection_operations.yaml
```diff
@@ -146,6 +146,43 @@ auth:
       }
 
 gcs:
+  - path: /collections
```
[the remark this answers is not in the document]
rudyardrichter: Yeah, just missed an `add`. It's kind of sparse though, if you have any other suggestions; it's a pretty basic "smoke test" as is.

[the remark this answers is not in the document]
sirosen: The biggest thing is to run the command at all, and you've got that. 😄 

I think you should also run it with each option provided (combinations if you want), and make sure the option appears in the request query string. If you look for tests which check `responses`, you'll find examples which inspect the "last request sent", and you can crib from those.

That would cover/catch the issue with the parameters which need to be in `query_params`. Plus, when we update the SDK and move those opts back "up to" the top level, it will be a measure of protection against typos and other "silly mistakes".

src/globus_cli/commands/collection/list.py
sirosen: FWIW, it also took me a few minutes for me to realize how to do this.

It's actually a case in which the plethora of options `click` gives us is somewhat paralyzing. I was reading `click.ParamType` again to remember all of the things it can -- and can't -- do!

src/globus_cli/commands/collection/list.py
sirosen: I filed two tracking stories for the follow-up on this. We'll add in the SDK, and then update here after an SDK release. All slated for CLI v3.1

rudyardrichter:
Thanks for the feedback @sirosen. I'll spend a little more time working on tests and let you know once this is updated.

src/globus_cli/commands/collection/list.py
```diff
@@ -0,0 +1,109 @@
...
+class ChoiceSlugified(click.Choice):
+    """
+    Allow either hyphens or underscores, e.g. both 'mapped-collections' or
+    'mapped_collections'
+    """
+
+    def convert(self, value, param, ctx):
+        if value is None:
+            return None
+        return super().convert(value, param, ctx).replace("_", "-")
```
sirosen, line 27: Looks like this is only doing one of the two replacements. If you want to accept `mapped_collections` as an input, match against `mapped-collections` (as show in helptext), and get `mapped_collections` as an output, I think you need
```python
super().convert(value.replace("_", "-"), param, ctx).replace("-", "_")
```

rudyardrichter: The expanded tests made a similar complaint (yay).

rudyardrichter:
So, I'm wondering about additional tests for things like, in this case, `--include-private-policies`. Looking at [this test as an example](https://github.com/rudyardrichter/globus-cli/blob/042ad2ec02eca16ec8761a09825c2df54db74ab4/tests/functional/collection/test_collection_show.py#L27-L29), is there a way to "meaningfully" test that flag if the API fixture is going to make the call return it either way (and likewise with filters)? Expanding the API fixtures stuff to handle query params to that effect seems ambitious.

sirosen (approved):
Everything looks good, and I just tested it out on a live endpoint to confirm the private policy option works.

PR #564, opened 2021-10-04
Add `globus search query` command with GET & POST

sirosen:
This is part of CLI v3.1 . It is the first bit of Globus Search support in the CLI, building on what is now possible in terms of added requirements with LoginManager.

I'll be adding other commands in the future and possibly refining this one if it seems like there are obvious improvements in advance of v3.1.

Because Globus Search returns user-defined JSON (with Globus-defined wrappers), it's not clear what non-JSON printing would be. So, for now at least, this command does not have non-JSON text output.

---

The command supports two usage modes for `GET` and `POST` query submission. Unlike the globus-search-cli version of this command, `--offset` is not supported, in anticipation of eventual use of built-in pagination.

The other major difference from the globus-search-cli commands is the fact that `query` is supplied as one command, with `-q` or `--query-document` as options which switch between GET and POST query forms.

### DIFF
--- a/src/globus_cli/commands/__init__.py
+++ b/src/globus_cli/commands/__init__.py
@@ -11,6 +11,7 @@ from globus_cli.commands.ls import ls_command
 from globus_cli.commands.mkdir import mkdir_command
 from globus_cli.commands.rename import rename_command
 from globus_cli.commands.rm import rm_command
+from globus_cli.commands.search import search_command
 from globus_cli.commands.session import session_command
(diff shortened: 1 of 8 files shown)


sirosen:
Oh, perhaps noteworthy: the index ID shown in the test fixture data exists in production and is a public index meant as an example.
e.g. Try `globus search query "6f831ac8-4c41-4812-b383-6fb04f8b9f9f" -q pepper` on this branch and you should see results.

Issue #565, opened 2021-10-06
`TypeError` raised when cancelling all tasks


kurtmckee:
Running `globus task cancel --all` raises a `TypeError` due to changes to a method signature in globus-sdk-python.

The changes were introduced in globus/globus-sdk-python@8b58b50988b17ff841dbc8a052430f5b401ebca3, which modified the `task_list()` call signature to remove implicit arbitrary keyword arguments:

```python
# Old signature
# -------------

def task_list(**params):
    pass


# New signature
# -------------

def task_list(..., query_params: dict):
    pass
```

To reproduce:

1. Run `globus login` and successfully login.
2. Run `globus task cancel --all`.

sirosen:
Yep, definitely broken! 😢 
As we discussed, you've volunteered to write up the fix. _Ideally_ that includes adding a test for this, so I'll assume you're doing that too.

I went fishing for a good example commit to point at for this, and I think [this one](https://github.com/globus/globus-cli/commit/10579f03339b1dff8d0a4ac8748b32fc6d30dc2c) fits the bill nicely. It shows a suite of new tests + a test fixture to provide the data.

You should be able to define a new file in the `api_fixtures` dir with responses under the `transfer` key, as in the commit above, and then load that up by calling the `load_api_fixtures` pytest fixture. (Sorry about the overloaded use of the word "fixture"! It never looked odd until today...)
We've got a few choices for sources of data, but I like pulling from documentation examples when I can (saves me time and validates the examples). In this case, [the Transfer Task docs](https://docs.globus.org/api/transfer/task/).

If you run into trouble, just let me know and we can get things sorted out.

kurtmckee:
Great info, Stephen! I'll work on this shortly.

PR #566, opened 2021-10-06
Fix TypeError raised when cancelling all tasks

kurtmckee:
This adds two additional functional tests, one of which demonstrates #565. It also fixes the the TypeError crash.


sirosen:
I don't have any nits to pick with this or see any issues. CI just passed, so I'm going to merge and release! 👍

PR #568, opened 2021-10-06
Add search commands

sirosen:
This adds
- `globus search query`
- `globus search ingest`
- `globus search task show`
- `globus search task list`

These are pretty much the minimum to make Search usable via the CLI.

Note that this PR is based on top of #567 , but does not set the label. It should (🤞! ) pass the CI check for changelog content, as a news fragment was added in each of the search-related commits.

---

Update: we can see 
passing check on this PR | passing check on scriv PR (with label)
---|---
image | image


sirosen:
As we discussed yesterday, I've altered things slightly and this is going to merge.

PR #571, opened 2021-10-09
Fix handling of pipx installs in 'globus update'

sirosen:
These currently fail under most cases (where PIPX_HOME matches the user's userbase dir), as `--user` is passed to pip in a non-user venv.

To resolve:
1. Remove the hidden `--development` flags from `update`, as these have not been used in practice. This simplifies and dedents code.
2. Add a hidden `--force` flag which makes testing `update` easier
3. Fix the issue by detecting when the install is under PIPX_HOME, and skipping the `--user` option in that case

---

Testing this is tricky. The best I can recommend is trying `pipx install . --force` from the repo root, followed by `globus update --force`.

### DIFF
--- a/src/globus_cli/commands/update.py
+++ b/src/globus_cli/commands/update.py
@@ -60,7 +60,7 @@ def _check_pip_installed():
 )
 @click.option("--force", is_flag=True, hidden=True)
 @click.option("--yes", is_flag=True, help='Automatically say "yes" to all prompts')
-def update_command(yes, force, development, development_version):
+def update_command(yes: bool, force: bool) -> None:
     """Update the Globus CLI to its latest version.
 
     The *globus update* command checks if a more recent version of the Globus CLI


src/globus_cli/commands/update.py
```diff
@@ -48,19 +58,11 @@ def _check_pip_installed():
...
 )
+@click.option("--force", is_flag=True, hidden=True)
 @click.option("--yes", is_flag=True, help='Automatically say "yes" to all prompts')
-# hidden options to fetch branches or tags from GitHub. One turns this mode
-# on or off, and the other is used to set a non-main target
-# --development-version implies --development
-@click.option("--development", is_flag=True, hidden=True)
-@click.option("--development-version", hidden=True, default=None)
-def update_command(yes, development, development_version):
+def update_command(yes, force, development, development_version):
```
aaschaer, line 63: Probably should remove the development and development_version arguments here as well

aaschaer (commented):
Can't speak towards the pipx functionality, but other than one place where the development arguments were cleaned out this looks good to me

src/globus_cli/commands/update.py
sirosen: Oh, yeesh! Yeah, I can't believe I missed that. And I'm somewhat surprised that the command runs happily without them. I'll fix right away.

PR #579, opened 2021-10-20
Refactor recursive ls and fix bug

### LINKED ISSUE #577: globus ls --filter incompatible with --recursive
It seems the `globus ls --filter` and `globus ls --recursive` options are mutually exclusive:

```
$ globus ls --recursive --filter='~*[Ro][Tou][Any]Complet*.[tx][mx][lt]' ${source_endpoint}:${source_path}/  
AttributeError: 'RecursiveLsResponse' object has no attribute 'first_elem'
$ globus version
Installed version:  3.1.2
Latest version:     3.1.2

You are running the latest version of the Globus CLI
```

It would be convenient if filter patterns could be specified when recursive listings are used.

sirosen:
`ls --recursive` failed with a hard error on any empty listings. The ls on `/mnt` is a good test-case for this, but it's also possible to get this situation easily by applying a filter to a recursive `ls`.

In order to fix the bug, an initial failing test was added. However, this work also refactors recursive ls to improve the maintainability of this feature. The main iteration is decomposed into smaller units and type annotations are applied.

Resolves #577

---

Did this actually improve maintainability? Well, at the very least it reduced cyclomatic complexity.
<table>
<tr>
<th>
Before
</th>
<th>
After
</th>
</tr>

<tr>
<td><pre>
$ radon cc src/globus_cli/services/transfer/recursive_ls.py
src/globus_cli/services/transfer/recursive_ls.py
    M 80:4 RecursiveLsResponse.iterable_func - C
    C 14:0 RecursiveLsResponse - B
    M 26:4 RecursiveLsResponse.__init__ - A
    M 76:4 RecursiveLsResponse.__iter__ - A
</pre></td>
<td><pre>
$ radon cc src/globus_cli/services/transfer/recursive_ls.py
src/globus_cli/services/transfer/recursive_ls.py
    M 113:4 RecursiveLsResponse._iterable_func - B
    C 53:0 RecursiveLsResponse - A
    F 35:0 _extend_q - A
    F 20:0 _client_side_limiter - A
    M 72:4 RecursiveLsResponse.__init__ - A
    M 108:4 RecursiveLsResponse.__iter__ - A
</pre></td>
</tr>
</table>

### DIFF
--- a/src/globus_cli/services/transfer/recursive_ls.py
+++ b/src/globus_cli/services/transfer/recursive_ls.py
@@ -3,9 +3,9 @@
 import logging
 import time
 from collections import deque
-from typing import Any, Deque, Dict, Iterator, List, Optional, Tuple, cast
+from typing import Any, Deque, Dict, Iterator, Optional, Tuple, cast
 
-from globus_sdk import GlobusHTTPResponse, TransferClient
+from globus_sdk import TransferClient
 
 log = logging.getLogger(__name__)
 
@@ -30,24 +30,6 @@ def _client_side_limiter() -> Iterator[None]:
             time.sleep(SLEEP_LEN)
 
         yield
-
-
-def _extend_q(q: QUEUE_T, res: GlobusHTTPResponse, relpath: str, depth: int) -> None:
-    # add data to queue, including the dir's name in the absolute and relative paths
-    # and increase the depth by one.
-    # data is reversed to maintain any "orderby" ordering
-    res_data = cast(List[Dict[str, Any]], res["DATA"])
-    q.extend(
-        [
-            (
-                res["path"] + item["name"],
-                (relpath + "/" if relpath else "") + item["name"],
-                depth + 1,
-            )
-            for item in reversed(res_data)
-            if item["type"] == "dir"
-        ]
-    )
 
 
 class RecursiveLsResponse:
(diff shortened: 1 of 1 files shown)


src/globus_cli/services/transfer/recursive_ls.py
```diff
@@ -21,63 +60,57 @@ class RecursiveLsResponse:
...
             # way of making sure that it's clear
-            self.generator = None
+            self._first_elem = None
 
-    def __iter__(self):
-        yield self.first_elem
-        yield from self.generator
+    def __iter__(self) -> Iterator[ITEM_T]:
+        if self._first_elem is not None:
+            yield self._first_elem
```
rudyardrichter, line 110: I'm wondering here what this first elem thing is accomplishing—why not just `yield from self._generator`?

rudyardrichter (commented):
One question, and like we mentioned earlier I voice my preference for `_extend_q` being inlined, but otherwise I'm in agreement with this.

src/globus_cli/services/transfer/recursive_ls.py
sirosen: The first element is grabbed from the generator eagerly so that an error will be triggered immediately if the first network call fails.
Otherwise, it's possible to have semantics like
```python
res = recursive_ls(...)  # passes
print(res)  # fails
```

It's still possible to start failing part-way through, but this ensures that once the initial `recursive_ls()` function returns, at least one call worked.

PR #583, opened 2021-11-03
Fix 'globus session consent' crashfail after login

sirosen:
At the end of the login flow, `globus session consent` would expect `id_token` in the response data in order to do the no-account-switching validation step we added in v3.0. However, the `id_token` is only present on login flows which include the `openid` scope.

To rectify:
- add `LoginManager.always_required_scopes`, an iterator which yields the openid scope string, "openid"
- `LoginManager.run_login_flow` adds `always_required_scopes` to any scope list (implicit or provided)
- `globus session consent` now passes scopes as a keyword argument, not under `session_params`
- annotate and type-check `globus session consent`

This was tested manually, and it is not clear how effective a mocked test with synthetic response data would be in preventing errors of this kind in the future.

Aside: I've converted one method typed as `Generator` to `Iterator`, as it is a simpler type to annotate. I used Iterator for the new iterator method, and it made sense to harmonize the existing method with it.

### DIFF
--- a/changelog.d/20211103_221444_sirosen_fix_session_consent.md
+++ b/changelog.d/20211103_221444_sirosen_fix_session_consent.md
@@ -1,0 +1,5 @@
+### Bugfixes
+
+* Fix a bug in `globus session consent` in which an `id_token` was expected as
+  part of the token data, but the `openid` scope was not provided to the login
+  flow
--- a/src/globus_cli/commands/session/consent.py
+++ b/src/globus_cli/commands/session/consent.py
@@ -1,3 +1,5 @@
+from typing import Tuple
+
 import click
 
 from globus_cli.login_manager import LoginManager
@@ -11,7 +13,7 @@ from globus_cli.parsing import command, no_local_server_option
 )
 @no_local_server_option
 @click.argument("SCOPES", nargs=-1, required=True)
-def session_consent(scopes, no_local_server):
+def session_consent(scopes: Tuple[str], no_local_server: bool) -> None:
     """
     Update your current CLI auth session by authenticating with a specific scope or set
     of scopes.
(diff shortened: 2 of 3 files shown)


rudyardrichter (commented):
This is quite minor, but couldn't `login_requirements` just return a list and itself add `AuthScopes.openid` to the list if not present? It seems that depends what we want `.is_logged_in()` to see? I might guess `.is_logged_in` is supposed to also see `openid` in the login requirements.

sirosen:
The trouble is that the `session consent` command is running `LoginManager.run_login_flow(..., scopes=[some, scopes])`. As a result, `LoginManager.login_requirements` is not being used to build that list of scopes.
I found adding the extra iterator was the simplest way to make the class "explain" the addition of the openid scope to any existing list of scopes.

I did consider making the "always required scopes" just be the scopes from `login_requirements`, and not bother adding a new property. However, I found that confusing, and I'm not sure it would be correct: in the `session consent` case, we're typically handling `urn:globus:auth:scopes:transfer.api.globus.org:all[*https://...]` to encode a dependency (you get to this by trying `globus ls` or another Transfer operation on a GCSv5 Mapped Collection). Passing that optional-dependent-scope string alongside the base transfer:all scope might do weird things or even fail. Adding a check that the scopes you're adding aren't in the list _or_ substrings of elements of the list is even more stuff to explain in the comments and code... and it might still miss things.

It is technically possible, even with the version of things I put together, to do odd stuff. You can `globus session consent 'openid[profile]'` and we'll happily put you through a login flow. Auth doesn't error out on the resulting scope string of `openid openid[profile]`, at least not today, but I don't know if we consider that correct or not. It's so clearly a boundary case, I could easily see that becoming invalid behavior in the future.

Since there's no use-case of which I'm aware for the `openid` scope to have dependent scopes, I'm not too worried about the potential for misuse to lead to strange behaviors or even bugs down the line. Prior to this thread, I somewhat doubt anyone would even think to run
```
globus session consent 'openid[*urn:globus:auth:scopes:transfer.api.globus.org:all]'
```

Issue #584, opened 2021-11-04
Bug in task event-list commnand


poquirion:
```
$ globus version
Installed version:  3.1.2
Latest version:     3.1.2

You are running the latest version of the Globus CLI

$ globus task event-list  --filter-non-errors   6c978c8c-faf0-11eb-ab67-d195c983855c
TypeError: task_event_list() got an unexpected keyword argument 'filter'

```

I get the same thing with the `--filter-errors` option.

sirosen:
Thanks for reporting this. It appears to be another issue with the transition to globus-sdk v3.0 (there have been a few of these, where our testing is weakest).
We'll get this resolved right away.

sirosen:
We've just released v3.1.3 which fixes this. We'll also be notifying the listhost about this, but I figured we should let you know here as well.

Again, thanks for the bug report!

PR #585, opened 2021-11-05
Fix a crash due to incorrect function parameter names

kurtmckee:
This appears to resolve the crash and match the query parameter signature expected by the API:

https://docs.globus.org/api/transfer/task/#get_event_list

[sc-11551]


sirosen:
Looks good. This also closes #584.

I want this command tested and type checked, but I can work on that today. Hopefully I can get that in and do a release this afternoon.

PR #589, opened 2021-11-30
Consolidate client creation into LoginManager

sirosen:
@aaschaer, this is very similar to what we worked on together. As I feared, the naive version of it introduced a dependency cycle (because CustomTransferClient used globus_cli.termio). I've worked through that issue, and some others.

I ended up removing `ClientGetter` as a class and put the various methods directly on the `LoginManager`, which I think you'll like. It ended up being simpler than trying to do this on a separate dedicated object.

---

This adds the following methods to LoginManager:
  - get_transfer_client
  - get_auth_client
  - get_groups_client,
  - get_search_client
  - get_gcs_client

Excepting the GCS client, none of these methods take any arguments.
The GCS client requires *either* the collection ID of a collection on the GCS endpoint or the endpoint ID.

These replace the client getters in `globus_cli.services`.
For the most part, the clients should be now obtained by taking a login manager passed to a command and calling these methods.

For this reason, this changeset also removes the `pass_manager=t/f` argument from `LoginManager.requires_login` and the manager is always passed.

Additional changes and tweaks were needed to adjust for these updates:
1. globus_cli.login_maanger now relies on globus_cli.services to provide custom client classes. To avoid dependency cycles,
 `CustomTransferClient.task_wait_with_io` has been moved to a helper function in globus_cli.commands
2. The login manager is annotated in the arguments to every command now, meaning that mypy is now checking command method bodies. Some annotations were needed.
3. In support of (2), introduce `globus_cli.types` as originally introduced in #587
4. The recursive LS response type is now visible from the transfer subpackage for use in annotations
5. Several helper functions are now custom client methods in order to simplify our internal interfaces


aaschaer (approved):
I do like having the client getting methods consolidated in one class, and it doesn't seem to bloat the LoginManager too much.

Looks good to me, and having helped work on the first pass made this a pretty easy review despite the large number of files with changes.

PR #590, opened 2021-12-02
client identities

aaschaer:
Work for allowing client identities to use the CLI by setting GLOBUS_CLI_CLIENT_ID and GLOBUS_CLI_CLIENT_SECRET environment variables.

One thing I haven't fully tested is switching identities without running logout. Once the old client's tokens have expired all should be well, but it might be possible to be in a bad state before then. Not sure how much we care about that, but we could potentially namespace by client id in the db if we wanted to avoid the issue altogether?

### DIFF
--- a/changelog.d/20211213_135655_aaschaer_client_login.md
+++ b/changelog.d/20211213_135655_aaschaer_client_login.md
@@ -1,5 +1,10 @@
 ### Enhancements
 
-* Add support for client identities to use the Globus CLI
+* Add support for client credentials for authentication in the Globus CLI
   by setting GLOBUS_CLI_CLIENT_ID and GLOBUS_CLI_CLIENT_SECRET
   environment variables (:pr:`NUMBER`)
+** Both variables must be set to enable this behavior
+** Tokens generated with client credentials are cached in the current user's home
+   directory, but isolated from any user credentials
+** With client credentials, `globus login` is invalid, but `globus logout` can be used
+   to revoke any cached tokens
--- a/changelog.d/20211213_135655_aaschaer_client_login.md
+++ b/changelog.d/20211213_135655_aaschaer_client_login.md
@@ -1,0 +1,5 @@
+### Enhancements
+
+* Add support for client identities to use the Globus CLI
+  by setting GLOBUS_CLI_CLIENT_ID and GLOBUS_CLI_CLIENT_SECRET
+  environment variables (:pr:`NUMBER`)
--- a/src/globus_cli/login_manager/manager.py
+++ b/src/globus_cli/login_manager/manager.py
@@ -293,6 +293,8 @@ class LoginManager:
             collection_id=collection_id, endpoint_id=endpoint_id
         )
 
+        # client identities need to have this scope added as a requirement
+        # so that they correctly request it when building authorizers
         self.add_requirement(
             gcs_id, scopes=[GCSEndpointScopeBuilder(gcs_id).manage_collections]
         )
--- a/src/globus_cli/login_manager/manager.py
+++ b/src/globus_cli/login_manager/manager.py
@@ -4,7 +4,13 @@ from typing import Dict, Iterator, List, Optional, Tuple
 
 import click
 import globus_sdk
-from globus_sdk.scopes import AuthScopes, GroupsScopes, SearchScopes, TransferScopes
+from globus_sdk.scopes import (
+    AuthScopes,
+    GCSEndpointScopeBuilder,
+    GroupsScopes,
+    SearchScopes,
+    TransferScopes,
+)
 
 from globus_cli.endpointish import Endpointish, EndpointType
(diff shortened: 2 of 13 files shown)


changelog.d/20211213_135655_aaschaer_client_login.md
```diff
@@ -0,0 +1,5 @@
+### Enhancements
+
+* Add support for client identities to use the Globus CLI
+  by setting GLOBUS_CLI_CLIENT_ID and GLOBUS_CLI_CLIENT_SECRET
+  environment variables (:pr:`NUMBER`)
```
sirosen, line 5:
```suggestion
* Add support for client credentials for authentication in the Globus CLI
  by setting GLOBUS_CLI_CLIENT_ID and GLOBUS_CLI_CLIENT_SECRET
  environment variables (:pr:`NUMBER`)
** Both variables must be set to enable this behavior
** Tokens generated with client credentials are cached in the current user's home
   directory, but isolated from any user credentials
** With client credentials, `globus login` is invalid, but `globus logout` can be used
   to revoke any cached tokens
```

I think the changelog is a good place to explain usage, since it's where a lot of people will be looking to see the new feature. Above is a suggested first-pass at fleshing out the text a bit more. We can iterate further on this, but I think we want a bit more text to explain.

We'll also want a page on the docs site, probably. But adding it here is still important.

src/globus_cli/login_manager/tokenstore.py
```diff
@@ -116,6 +122,9 @@ def internal_auth_client():
         client_secret = client_data["client_secret"]
     else:
         # register a new instance client with auth
+        if is_client_login():
+            raise ValueError("client logins shouldn't create internal auth clients")
```
sirosen, line 126: What about moving this check to the start of the method?

I see that there's handling for this in the `delete_templated_client` method, and it might be simpler if we assert that this method can _never_ run when `is_client_login() == True`.

src/globus_cli/login_manager/tokenstore.py
```diff
@@ -140,7 +149,15 @@ def delete_templated_client():
...
-    ac = internal_auth_client()
+    try:
+        ac = internal_auth_client()
+
+    # client logins shouldn't have templated clients to delete,
+    # but just in-case there was one catch errors here and return
+    # rather than skip cleanup entirely
+    except ValueError:
+        if is_client_login():
+            return
```
sirosen, line 160: Right now, this check would discard any `ValueError` and fallthrough if `is_client_login() == False`.
We want to fix that, but I'm also looking at the overall "flow" and wondering if it might be better to leave this method unmodified.

Instead of making `delete_templated_client()` work when `is_client_login() == True`, what about letting it fail, and updating `globus logout` to ensure that we don't end up calling it?
Otherwise, when reading the source for logout, we see a method call which "looks wrong" (because there shouldn't be a client to delete), and it's not immediately obvious that it's a no-op.

src/globus_cli/login_manager/manager.py
```diff
@@ -246,6 +293,9 @@ def get_gcs_client(
             collection_id=collection_id, endpoint_id=endpoint_id
         )
 
+        self.add_requirement(
+            gcs_id, scopes=[GCSEndpointScopeBuilder(gcs_id).manage_collections]
+        )
```
sirosen, line 298: This change doesn't seem obviously related to the other content of this PR, but I expect that it's needed to get the accurate scope passed to the client creds token callout.

Can you add a comment which explains? I don't think this impacts the non-client-creds case, but it would be good to have internal documentation about how things behave.

sirosen (commented):
I want to be extra-careful and test things thoroughly before approving and merging.
However, this all looks to me like it's going to work.

I've left a few comments on things which I recommend changing.

I think it would also be nice to have two additional tests:
- One which runs any real command (e.g. get-identities) with a client login setup
- One test for the retry handler on the "happy path" where the client creds callout works and the subsequent call succeeds

These will likely require new fixture data.

I'm really happy to see the retry handler was possible to write, and doesn't seem to have been too complex or difficult. But I think we need a test for it.

src/globus_cli/login_manager/manager.py
aaschaer: Added a comment explaining

sirosen (commented):
I've come back to this and finished my testing. I think this is good to go.

I'd like us to discuss it briefly at our meeting tomorrow prior to merging, however.
If we put this in without adding the two tests I want to add, they might never get added.
That's not catastrophic -- I'd rather release this change with those tests as a TODO item than hold this feature back for a long time -- but if they're within reach, let's get them included.

sirosen (approved):
Adding a test for the retry handling would be very valuable, but we don't need to block on it. On the strength of manual testing, I'm going to move ahead with this.

PR #592, opened 2021-12-03
Update dependency versions, build on py3.10

sirosen:
Update versions of dev tools and dependencies. This touches pre-commit config, tox config, and setup.py

Add python 3.10 to the build matrix and classifiers.

I also tried to make the github workflows more consistent about how quoting is done.


sirosen:
I just had to amend to revert some of the quotation mark changes. It seems that single-quotes are valid inside the github expression syntax in some contexts where double-quotes are not.

rudyardrichter (approved):

PR #593, opened 2021-12-04
Add `globus group member add` command

rudyardrichter:
- Add `member` command group
- Add `add` command on member
- Add tests and API fixture for member add


src/globus_cli/commands/group/member/add.py
```diff
@@ -0,0 +1,45 @@
...
+
+ADD_USER_FIELDS: FIELD_LIST_T = [
+    ("Group ID", _format_group_id),
+    ("Added User ID", _format_added_user_id),
+    ("Added User Username", _format_added_user_username),
+]
+
+
+@command("add", short_help="Add one or more members to a group")
+@click.argument("group", type=str)
```
sirosen, line 29: Because it's a positional argument (and therefore has no help string), I think this should be `group_id`.
It's a small thing, but I think it will be better for users, since it clarifies that you need the ID, not the name or some other field here.

I would also type it as a `click.UUID` so that non-UUID inputs fail.

src/globus_cli/commands/group/member/add.py
```diff
@@ -0,0 +1,45 @@
...
+@click.argument("user", type=IdentityType())
+@LoginManager.requires_login(LoginManager.GROUPS_RS)
+def member_add(group: str, user: ParsedIdentity, login_manager):
+    """Add one or more members to a group"""
```
sirosen, line 33: Let's expand the helptext for this command a little bit to explain that `USER` can be a username or identity ID.
Unlike `group` -> `group_id`, I don't think there's a subtle indicator we can do here. I'd just explain in the help, similar to what we have for `globus session update`: https://github.com/globus/globus-cli/blob/8f9267e66159ed5816c3b2db1af7731c6665190f/src/globus_cli/commands/session/update.py#L89

src/globus_cli/commands/group/member/add.py
```diff
@@ -0,0 +1,45 @@
...
+def _format_added_user_id(x: dict):
+    return x["add"][0]["identity_id"]
+
+
+def _format_added_user_username(x: dict):
+    return x["add"][0]["username"]
+
+
+def _format_group_id(x: dict):
+    return x["add"][0]["group_id"]
```
sirosen, line 18: Unfortunately, there's an issue with unpacking the response like this.
Groups responds with an empty array if the user is already in the group, i.e. `{"add": []}`.

We're going to need to make some change to handle it. I think these helpers can stay unmodified, and I'll provide a suggestion down below, where `formatted_print` is called.

src/globus_cli/commands/group/member/add.py
```diff
@@ -0,0 +1,45 @@
+import click
+
+from globus_cli.login_manager import LoginManager
+from globus_cli.parsing import IdentityType, ParsedIdentity, command
+from globus_cli.termio import FORMAT_TEXT_RECORD, formatted_print
+from globus_cli.types import FIELD_LIST_T
+
+
+def _format_added_user_id(x: dict):
+    return x["add"][0]["identity_id"]
```
sirosen, line 10: When I first saw this, it reminded me of a feature we have in the output formatter to unpack results. You can pass `response_key=...` to `formatted_print`. It's then used to index into the response data before text printing:
https://github.com/globus/globus-cli/blob/8f9267e66159ed5816c3b2db1af7731c6665190f/src/globus_cli/termio/output_formatter.py#L258

Unfortunately, `response_key` only accepts a string right now, as you can see. We could make it accept a callable, and then these helpers would be replaced with this usage:
```python
ADD_USER_FIELDS: FIELD_LIST_T = [
    ("Group ID", "group_id"),
    ("Added User ID", "identity_id"),
    ("Added User Username", "username"),
]

...

formatted_print(
    ...,
    response_key=lambda x: x["add"][0]
)
```

If you're interested in doing that here, to support this command, I'd be onboard with it. You basically need to change the use of `response_key` to check `callable(response_key)` and do little else.

src/globus_cli/commands/group/member/add.py
```diff
@@ -0,0 +1,45 @@
...
+    identity_id = auth_client.maybe_lookup_identity_id(user.value)
+    if not identity_id:
+        raise click.UsageError(f"Couldn't determine identity from user value: {user}")
+    actions = {"add": [{"identity_id": identity_id}]}
+    response = groups_client.batch_membership_action(group, actions)
+    formatted_print(
+        response,
+        text_format=FORMAT_TEXT_RECORD,
+        fields=ADD_USER_FIELDS,
+    )
```
sirosen, line 45: I believe that this `formatted_print` call is the correct place to handle the case where the user is already a member of the group.

Here's an example of the full response data:
```json
{
  "add": [],
  "errors": {
    "add": [
      {
        "code": "ALREADY_ACTIVE",
        "detail": "The identity is already an active member of the group.",
        "identity_id": "c699d42e-d274-11e5-bf75-1fc5bf53bb24"
      }
    ]
  }
}
```

We may want more sophisticated behavior in the future, but for now I think the correct approach is for the CLI to treat this type of response as an error condition. We could raise a `ValueError`, or a custom error class -- either would be acceptable. That will get us a relatively clean `exit(1)` plus a message.

It's okay to try to unpack the `"detail"` and `"code"` fields, but please assume if you do so that the fields may not be present. I don't know if it's possible in fact for a response to come back which doesn't populate `"add"` in the main response body and also doesn't populate `"errors"`, but it's at least possible in theory.

sirosen (changes requested):
I'm overall pretty well satisfied that this is what we want.

I have some feedback which is minor cleanup, but unfortunately I did find one buggy condition when adding a user to a group twice (or otherwise to a group in which they are already a member). That needs resolution, and I've provided a suggested way of handling it.

src/globus_cli/commands/group/member/add.py
rudyardrichter: Thanks, I was thinking this might have been flaky and this would need to be discussed. Thanks for such detailed suggestions.

sirosen:
I see a fresh request for review, but this doesn't look quite done yet (?). The `response_key` change in `formatted_print` looks good, but it's not being used in the command yet.

I'd also like to see the "member already in group" case tested. I think there's an error in the nested `get()` calls to extract `detail` from `errors`, and a test would quickly confirm this and lead to resolution.

rudyardrichter:
Ah, sorry. I misunderstood what is actually going on in `formatted_print`. Actually, wouldn't this update to `formatted_print` then overlap with what a callable in the fields argument is already doing?

sirosen:
> Actually, wouldn't this update to `formatted_print` then overlap with what a callable in the fields argument is already doing?

There is a good amount of overlap, but I don't think that's necessarily a problem.

One allows us to use an arbitrary callable compute a string from a complex structure. The other lets us use an arbitrary callable to get a dict from a payload, in order to pass to all of the fields.

Allowing a callable for `response_key` allows all of the fields to be simple strings. It's just a matter of where you want to do that dict and list unpacking step: in each field, or before printing?

rudyardrichter:
Yeah, I'm on the same page now. It does make the fields constant nicer. The latest version here should be in line with your original feedback now, I think.

sirosen (approved):

PR #595, opened 2022-01-04
Add `group member remove` command

rudyardrichter:
Add `group member remove` (basically the counterpart of `group member add`) and a couple test cases.

Also fix the docs on `member add` which say "one or more members", because it must be exactly one. It seems like it would be nice to have `IdentityType`, or a variant thereof, handle that case. It doesn't seem like that would be too complicated, so if that makes sense I can add that here (or maybe, add that to `member add` and then update this).

### DIFF
--- a/changelog.d/20220103_164410_rudyardrichter_add_group_member_remove.md
+++ b/changelog.d/20220103_164410_rudyardrichter_add_group_member_remove.md
@@ -1,0 +1,3 @@
+### Enhancements
+
+* Add a `globus group member remove` command to remove a user from a group
--- a/src/globus_cli/commands/group/member/__init__.py
+++ b/src/globus_cli/commands/group/member/__init__.py
@@ -1,4 +1,5 @@
 from globus_cli.commands.group.member.add import member_add
+from globus_cli.commands.group.member.remove import member_remove
 from globus_cli.parsing import group
 
 
@@ -8,3 +9,4 @@ def group_member() -> None:
 
 
 group_member.add_command(member_add)
+group_member.add_command(member_remove)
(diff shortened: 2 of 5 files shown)


sirosen:
First off, the changes look good. I want to try using it a bit -- and trying to suss out any odd cases -- before approving and merging, but I don't foresee any problems.

> It seems like it would be nice to have `IdentityType`, or a variant thereof, handle that case. It doesn't seem like that would be too complicated, so if that makes sense I can add that here (or maybe, add that to `member add` and then update this).

I'm not sure that I'm following what "that case" is correctly. It sounds like you're referring to handling multiple usernames/identity-ids. I think we can rely on `multiple=True` to take care of this, so we shouldn't need to change our param types.

Regardless of whether or not I've understood, let's leave out any changes to `IdentityType` for now.
I want to get this and #590 tested and merged, and do a release. So I'd like to avoid anything which could stretch out the timeline for that.

rudyardrichter:
If I understand correctly, with this case as an example, since the user is an argument allowing multiples is via `nargs=-1`, which doesn't work out of the box because `IdentityType` gets all the arguments passed together as an iterable or something. Either way, agreed on leaving that out for now.

sirosen:
> allowing multiples is via `nargs=-1`

Sorry, yes, my mistake. `multiple=True` is for multi-use options.
I think that with `nargs=-1`, the param type gets passed each argument individually, but then the function implementing the command gets an iterable. We can (read: should) dig into this more later.

sirosen (approved):
I've just done some testing and wasn't able to break it in any new or exciting ways.

Excellent work!

PR #596, opened 2022-01-07
Add 'cli-profile-list' command + minor login fix

sirosen:
With the addition of client creds, I believe that it's important to have this command included in the same release.
For now, I've tried not to overthink it, and just list profiles and client creds in a single command with a separator in the output.

This is the last change I want to get included before we do the next CLI release.

---

Testing this work uncovered a minor issue with login, in which client creds do not trigger a specific message and error. This could be confusing. Rectify with a custom usage error.

To implement `globus cli-profile-list`, directly check the token storage database to list the namespaces which are in use. Filter to those created by the CLI, and then parse them (slash-delimited) to identify them by type (user vs client) and environment. Listing is done by table-printing the results.

Tests need to carefully inject data into the test token storage's in-memory DB to simulate the presence of additional profiles. Once done, several cases are tested in simple fashion, ensuring that expected profile names are or are not present in the output.

The `--all` option, which expands the listing to all environments, is kept hidden for now, on the assumption that it would confuse most users. If it is frequently used or the fact that it is hidden leads to confusion, it can be made public later.

### DIFF
--- a/src/globus_cli/commands/cli_profile_list.py
+++ b/src/globus_cli/commands/cli_profile_list.py
@@ -1,0 +1,97 @@
+import os
+from typing import Any, Dict, Iterator, List, Optional, Tuple
+
+import click
+import globus_sdk.tokenstorage
+
+from globus_cli.login_manager import token_storage_adapter
+from globus_cli.parsing import command
+from globus_cli.termio import FORMAT_TEXT_TABLE, formatted_print
+from globus_cli.types import FIELD_LIST_T
+
+
+# TODO: upstream this into the SDK as a method of the SQLiteStorageAdapter
+def _iter_namespaces(adapter: globus_sdk.tokenstorage.SQLiteAdapter) -> Iterator[str]:
+    conn = adapter._connection
+
+    cursor = conn.execute("SELECT DISTINCT namespace FROM token_storage;")
+    for row in cursor:
+        yield row[0]
+
+
+def _profilestr_to_datadict(s: str) -> Optional[Dict[str, Any]]:
+    if s.count("/") < 1:
+        return None
+    if s.count("/") < 2:
+        category, env = s.split("/")
+        if category == "clientprofile":  # should not be possible
+            return None
+        return {"client": False, "env": env, "profile": None, "default": True}
+    else:
+        category, env, profile = s.split("/", 2)
+        return {
+            "client": category == "clientprofile",
+            "env": env,
+            "profile": profile,
+            "default": False,
+        }
+
+
+def _parse_and_filter_profiles(
+    all: bool,
+) -> Tuple[List[Dict[str, Any]], List[Dict[str, Any]]]:
+    namespaces = list(_iter_namespaces(token_storage_adapter()))
+    globus_env = os.getenv("GLOBUS_SDK_ENVIRONMENT", "production")
+
+    client_profiles = []
+    user_profiles = []
+    for n in namespaces:
+        data = _profilestr_to_datadict(n)
+        if not data:  # skip any parse failures
+            continue
+        if (
+            data["env"] != globus_env and not all
+        ):  # unless --all was passed, skip other envs
+            continue
+        if data["client"]:
+            client_profiles.append(data)
+        else:
+            if all or data["profile"] is not None:
+                user_profiles.append(data)
+
+    return (client_profiles, user_profiles)
+
+
+@command(
+    "cli-profile-list",
+    disable_options=["format", "map_http_status"],
+)
+@click.option("--all", is_flag=True, hidden=True)
+def cli_profile_list(*, all: bool) -> None:
+    """
+    List all CLI profiles which have been used
+
+    These are the values for GLOBUS_PROFILE which have been recorded, as well as
+    GLOBUS_CLI_CLIENT_ID values which have been used.
+    """
+
+    client_profiles, user_profiles = _parse_and_filter_profiles(all)
+
+    if user_profiles:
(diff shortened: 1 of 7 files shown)


src/globus_cli/commands/cli_profile_list.py
```diff
@@ -0,0 +1,97 @@
...
+        yield row[0]
+
+
+def _profilestr_to_datadict(s: str) -> Optional[Dict[str, Any]]:
+    if s.count("/") < 1:
+        return None
+    if s.count("/") < 2:
+        category, env = s.split("/")
+        if category == "clientprofile":  # should not be possible
+            return None
```
aaschaer, line 28: Should we print a warning or something if we do somehow hit a "shouldn't be possible" case?

src/globus_cli/commands/cli_profile_list.py
```diff
@@ -0,0 +1,97 @@
...
+        if not data:  # skip any parse failures
+            continue
+        if (
+            data["env"] != globus_env and not all
+        ):  # unless --all was passed, skip other envs
+            continue
+        if data["client"]:
+            client_profiles.append(data)
+        else:
+            if all or data["profile"] is not None:
```
aaschaer, line 59: I feel like we should include the default userprofile for the current environment even without all set. It feels odd/potentially confusing not having that profile shown even when its the active profile,

aaschaer (commented):
First of all I agree this is a useful command to have! And that `_iter_namespaces` should be backported into the SDK at some point.

Two things that might be interesting to add to output:
An "Active" or "In Use" header, showing which of the profiles is currently active? Might be redundant with "whoami" though?
A "Logged in" header, showing which of the userprofiles have tokens (always True for client profiles)?

src/globus_cli/commands/cli_profile_list.py
sirosen: I was avoiding doing so because the only way I can imagine these cases existing is someone tinkering with their DB directly. I think that at least by default we should ignore anything unrecognized.
Maybe with `--all` we could provide a list of unrecognized namespaces?

src/globus_cli/commands/cli_profile_list.py
sirosen: In my first draft, I had a warning about the default profile not being listed (but that looked more confusing, so I cut it).

Let me look again at including the default profile again. With an "active/current" marker, it might work out nicer.

sirosen:
> An "Active" or "In Use" header, showing which of the profiles is currently active? Might be redundant with "whoami" though?

I think this would be nice. Perhaps without it having a header, imagine
```
   | foo | bar
--------------
   | abc | def
-> | xyz | pqr
   | pqr | ghi
```

> A "Logged in" header, showing which of the userprofiles have tokens (always True for client profiles)?

I think I'll pass on trying to do this right now. I'm currently selecting from the token storage table, so the code will only list profiles with logins.
We could take this idea into the SDK, however. We could check the token storage and config tables, and provide results of the form `(namespace: str, found_in_token_storage: bool, found_in_config_storage: bool)`.

If we start getting back that kind of info, it would be a good time to improve the output in this command. Does that sound like a good plan?

sirosen:
I've made changes to include a current profile marker, and to always list the default profile (+ include the `is_default` column).

Output from my testing:
```
$ globus cli-profile-list
   GLOBUS_PROFILE | is_default
   -------------- | ----------
-> NULL           | True
   test-profile   | False

GLOBUS_CLI_CLIENT_ID
------------------------------------
7477f190-cecf-417f-80e3-ec246f992de8

$ globus cli-profile-list --all
   GLOBUS_PROFILE | is_default | GLOBUS_SDK_ENVIRONMENT
   -------------- | ---------- | ----------------------
-> NULL           | True       | production
   test-profile   | False      | production
   NULL           | True       | sandbox
   test-prof      | False      | sandbox
   NULL           | True       | test

GLOBUS_CLI_CLIENT_ID                 | GLOBUS_SDK_ENVIRONMENT
------------------------------------ | ----------------------
7477f190-cecf-417f-80e3-ec246f992de8 | production

$ . ./cc_creds.txt
$ globus cli-profile-list
GLOBUS_PROFILE | is_default
-------------- | ----------
NULL           | True
test-profile   | False

   GLOBUS_CLI_CLIENT_ID
   ------------------------------------
-> 7477f190-cecf-417f-80e3-ec246f992de8
```

If you look at the changes in 3458fc2 , you'll see several modifications to table printing.
This is to accommodate a fieldname of `""` as meaning that the column is untitled and should not have a header separator. (If we find there are issues with this, we can always add an explicit `Field(untitled=True)` or similar usage.)

src/globus_cli/commands/cli_profile_list.py
aaschaer: Probably fine to leave this as is then

sirosen: If we find that we want this at some point, we can always add an `if all: warnecho(...)` here. But yeah, let's leave it out for now.

aaschaer:
I like the arrow! I wonder if it might look better on the right side so the two tables would be aligned?
Not a big deal, so feel free to merge if you like it better on the left

aaschaer (approved):

sirosen:
We could move the arrow around, but I like it on the left.
Maybe we could do something to indent the other table. e.g. Make the arrow detected as `"-> "` or `"   "`.

I'm not going to fiddle with this right now, but I'd happily look at a PR if you want to tinker with it.

PR #597, opened 2022-01-24
Add Globus Search index role commands

sirosen:
Add commands for role create, list, delete, and tests for these commands with appropriate test fixture data as needed.

Like Transfer ACL commands, these commands support some interactions with Globus Auth. Response rendering is done with Auth to get usernames as appropriate for text output, and usernames are accepted as inputs.

The URN usages here may fit in a new param type in the future, but none is added yet, as this is the first potential use-case. Instead, existing helpers are used as much as is possible.


src/globus_cli/commands/search/index/role/create.py
```diff
@@ -0,0 +1,85 @@
...
+    login_manager: LoginManager,
+):
+    """
+    Create a role (requires admin or owner)
+
+    PRINCIPAL is expected to be an identity or group ID, a principal URN, or a username.
+
+    Example usage:
+       globus-search index role create admin "$index_id" "globus@globus.org"
+       globus-search index role create writer "$index_id" "$group_id" --type group
```
aaschaer, line 44: I think the arguments need to have "$index_id" before the ROLE_NAME, unless this is the intended order and the argument decorators are swapped?

aaschaer (commented):
Code looks good to me, minus one doc issue, and basic manual testing looks good on my end.

I'd prefer if the commands were calling dedicated SDK helpers instead of post etc, but don't think that needs to be a blocker necessarily.

src/globus_cli/commands/search/index/role/create.py
sirosen: No, your first guess correct. It was meant to be `"$index_id" admin`, and likewise for `writer`.
Thanks for the catch; I'll fix this.

sirosen:
> I'd prefer if the commands were calling dedicated SDK helpers instead of post etc, but don't think that needs to be a blocker necessarily.

I agree. But it would block this on at least one more SDK release, so I was going to forgo the helpers for now.

Let's refrain from merging this for a little bit -- I'll see how quickly I can get some simple helpers for these added.

aaschaer (approved):

PR #600, opened 2022-01-25
Use pypa's `build` to build dists

sirosen:
I was in a hurry to add `build` for `globus-sdk` as it was the best path to fixing the bad build which I had released.
This PR is the same change, but for the CLI. We can therefore take a bit more time to review and confirm that this is what we want.

---

`build` replaces use of `setup.py sdist bdist_wheel`. It has several advantages over the older approach of using `setuptools` and `wheel` directly:
- it builds in an isolated environment (by default), so that it can't be confused by local files when building a wheel
- it can support alternative build backends via pyproject.toml (if we add these), per [PEP517](https://www.python.org/dev/peps/pep-0517/)
- it's the more modern (and therefore documented) way of doing a build

A pyproject.toml is not needed for this change. `build` touts itself as an implementation of the standard set out in PEP517, and the spec states:

>  If the pyproject.toml file is absent, or the build-backend key is  missing, the source tree is not using this specification, and tools should revert to the legacy behaviour of running setup.py

This is our case, so `build` works as expected and desired.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -60,7 +60,7 @@ commands_pre = rm -rf dist/
 # check that twine validating package data works
 commands =
     python -m build
-   twine check dist/*
+    twine check dist/*
 
 [testenv:prepare-release]
 skip_install = true
--- a/tox.ini
+++ b/tox.ini
@@ -52,13 +52,15 @@ commands = python ./reference/_generate.py
 
 [testenv:twine-check]
 skip_install = true
-deps = twine
-       wheel
+deps =
+    build
+    twine
 whitelist_externals = rm
 commands_pre = rm -rf dist/
 # check that twine validating package data works
-commands = python setup.py sdist bdist_wheel
-           twine check dist/*
+commands =
+    python -m build
+   twine check dist/*
 
 [testenv:prepare-release]
 skip_install = true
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -52,13 +52,15 @@ commands = python ./reference/_generate.py
...
+    build
+    twine
 whitelist_externals = rm
 commands_pre = rm -rf dist/
 # check that twine validating package data works
-commands = python setup.py sdist bdist_wheel
-           twine check dist/*
+commands =
+    python -m build
+   twine check dist/*
```
rudyardrichter, line 63:
```suggestion
    twine check dist/*
```

rudyardrichter:
This is new to me, but `build` looks promising and useful here. I have to admit I'm not familiar at all with PEP 517, or `build`, so without more research I'm kind of yes-manning.

tox.ini
sirosen: Good catch; thank you!

sirosen:
> I have to admit I'm not familiar at all with PEP 517

You are, however, familiar with `poetry`, which gives you a good basis for understanding what PEP 517 is all about. I'll try to give you the 30 second summary:

Imagine you wanted to distribute a poetry package, with its original `pyproject.toml` file, and specify that the package should be installed with `poetry`. Likewise for `flit` or some of the other `setuptools` replacements.

Prior to the PEP, `pip` really just ran `setup.py` on install, so there wasn't any room for someone to add in a new mechanism. Now, the whole package-on-one-end, install-on-the-other-end system is pluggable, and `pyproject.toml` is the mechanism.

PR #606, opened 2022-02-23
group set-policies command

aaschaer:
Making a draft PR so we can discuss some design questions that came up while working on this:

* `group set-policy` or `group set-policies` ?

* should all policy fields be visible on a normal `group show`, or should there be something like an `--include-policies` flag?

* should `group_visibility` and `group_members_visibility` be flags rather than enums?

* there is currently no way to unset `signup_fields`, what would be a good UI for doing that?

### DIFF
--- a/changelog.d/20220222_140127_aaschaer_groups.md
+++ b/changelog.d/20220222_140127_aaschaer_groups.md
@@ -1,0 +1,27 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section or sections which match your change. Use "Other" for all
+changes which do not match a different section.
+
+Fill in one or more bullet points with details of your change.
+
+Make sure you add the new file in `changelog.d/` to your pull request!
+-->
+
+<!--
+### Bugfixes
+
+* A bullet item for the Bugfixes category.
+
+-->
+### Enhancements
+
+* Add `globus groups set-policies` command for managing a groups policies
+
+<!--
+### Other
+
+* A bullet item for the Other category.
+
+-->
--- a/changelog.d/20220222_140127_aaschaer_groups.md
+++ b/changelog.d/20220222_140127_aaschaer_groups.md
@@ -17,7 +17,7 @@ Make sure you add the new file in `changelog.d/` to your pull request!
 -->
 ### Enhancements
 
-* Add `globus groups set-policies` command for managing a groups policies
+* Add `globus groups set-policies` command for managing a group's policies
 
 <!--
 ### Other
--- a/src/globus_cli/commands/group/_common.py
+++ b/src/globus_cli/commands/group/_common.py
@@ -28,6 +28,18 @@ def parse_visibility(res):
     return res["policies"]["group_visibility"]
 
 
+def parse_members_visibility(res):
+    return res["policies"]["group_members_visibility"]
+
+
+def parse_join_requests(res):
+    return res["policies"]["join_requests"]
+
+
+def parse_signup_fields(res):
+    return ",".join(sorted({f for f in res["policies"]["signup_fields"]}))
+
+
 def group_create_and_update_params(
     f: Optional[Callable] = None, *, create: bool = False
 ) -> Callable:
--- a/src/globus_cli/commands/group/_common.py
+++ b/src/globus_cli/commands/group/_common.py
@@ -37,7 +37,7 @@ def parse_join_requests(res):
 
 
 def parse_signup_fields(res):
-    return ",".join(sorted({f for f in res["policies"]["signup_fields"]}))
+    return ",".join(sorted(f for f in res["policies"]["signup_fields"]))
 
 
 def group_create_and_update_params(
--- a/src/globus_cli/commands/group/set_policies.py
+++ b/src/globus_cli/commands/group/set_policies.py
@@ -1,6 +1,6 @@
-import click
+from typing import Optional
 
-from typing import Any
+import click
 
 from globus_cli.login_manager import LoginManager
 from globus_cli.parsing import command
@@ -8,16 +8,34 @@ from globus_cli.termio import formatted_print
 
 from ._common import group_id_arg
 
+SIGNUP_FIELDS = [
+    "address",
+    "address1",
+    "address2",
+    "city",
+    "country",
+    "current_project_name",
+    "department",
+    "field_of_science",
+    "institution",
+    "phone",
+    "state",
+    "zip",
+]
+
 
 @click.option(
     "--high-assurance/--not-high-assurance",
-    help="Flag if the group will enforce high assurance access policies or not."
+    default=None,
+    help="Flag if the group will enforce high assurance access policies or not.",
 )
 @click.option(
     "--authentication-timeout",
     type=int,
-    help=("Time in seconds before a user must reauthenticate to access a "
-          "high assurance group")
+    help=(
+        "Time in seconds before a user must re-authenticate to access a "
+        "high assurance group"
+    ),
 )
 @click.option(
     "--visibility",
@@ -29,7 +47,7 @@ from ._common import group_id_arg
     ),
 )
 @click.option(
-    "--member-visibility",
+    "--members-visibility",
     type=click.Choice(("members", "managers"), case_sensitive=False),
     help=(
         "Determine who can see members of the group. "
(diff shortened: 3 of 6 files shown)


src/globus_cli/commands/group/set_policies.py
```diff
@@ -0,0 +1,103 @@
...
+from globus_cli.login_manager import LoginManager
+from globus_cli.parsing import command
+from globus_cli.termio import formatted_print
+
+from ._common import group_id_arg
+
+
+@click.option(
+    "--high-assurance/--not-high-assurance",
+    help="Flag if the group will enforce high assurance access policies or not.",
```
sirosen, line 14: I think the behavior of this flag is that it is binary, `True/False`.
We actually want ternary, `True/False/None` (so that nothing is modified if it is not set). I'm hoping we have an example around of doing this -- or that I've forgotten and the `click` automatically gives us the `True/False/None` behavior.

src/globus_cli/commands/group/set_policies.py
```diff
@@ -0,0 +1,103 @@
...
+            "field_of_science",
+        ),
+        case_sensitive=False,
+    ),
+    multiple=True,
+    help=(
+        "Field required from users to apply for group membership. "
+        "Pass this option multiple times to require multiple fields. "
+        "Any existing required fields will be overwritten if given."
+    ),
```
sirosen, line 70: The behavior of this flag in particular is making me wonder if we want to pursue a suite of commands for managing a group's policies.

Do we actually want
```
globus groups policies set  # current implementation
globus groups policies signup-field add
globus groups policies signup-field remove
```
?

The group `policies` object _probably_ won't grow in complexity, but if it does, and there are more fields like `signup-field` which benefit from special-handling, it might be nice to have a dedicated namespace for such commands.

If we think that is the direction we want to take, the _only_ change I'd suggest today is to rename to `policies set`. Fancy `signup-field add/remove` support can be put on the roadmap.

src/globus_cli/commands/group/set_policies.py
```diff
@@ -0,0 +1,103 @@
...
+        "is_high_assurance": "high_assurance",
+        "authentication_assurance_timeout": "authentication_timeout",
+        "group_visibility": "visibility",
+        "group_members_visibility": "member_visibility",
+        "join_requests": "join_requests",
+        "signup_fields": "signup_field",
+    }
+
+    # assemble data using existing values for any field not given
+    print(kwargs)
```
sirosen, line 93:
```suggestion
```
Just a remnant. :-)

src/globus_cli/commands/group/set_policies.py
```diff
@@ -0,0 +1,103 @@
...
+
+    # map of groups API field names to CLI option names
+    field_option_map = {
+        "is_high_assurance": "high_assurance",
+        "authentication_assurance_timeout": "authentication_timeout",
+        "group_visibility": "visibility",
+        "group_members_visibility": "member_visibility",
+        "join_requests": "join_requests",
+        "signup_fields": "signup_field",
+    }
```
sirosen, line 90: Given that you have to do this anyway, I wonder if it might be clearer not to use `**kwargs` at all. Instead, data assembly could be
```python
data = {
  "is_high_assurance": high_assurance,
  ...,
  "group_members_visibility": member_visibility if member_visibility != () else None,
  ...
}
data = {k: v for k, v in data.items() if v is not None}
data = {
    k: (data[k] if k in data else existing_data[k])
    for k, v in existing_data.items()
}
```

The main thing I'm looking at is the mix between this "remapping" dict and the somewhat subtle check below of
```python
kwargs.get(option_name) not in (None, ())
```

This makes sense if you remember that a multiple option becomes the empty tuple when omitted. But this is a bit obscure for someone who isn't used to the library.

sirosen (commented):
I did a quick pass in advance of our taking time to talk about this. I haven't gotten a chance to try it out yet; just my initial thoughts after a read-through.

src/globus_cli/commands/group/set_policies.py
aaschaer: Ah right, I need to add a default=None, thank you!

src/globus_cli/commands/group/set_policies.py
aaschaer: whoops, good catch

src/globus_cli/commands/group/set_policies.py
aaschaer: Yeah, this would also be cleaner if we need to add more logic to convert click options into api fields

changelog.d/20220222_140127_aaschaer_groups.md
```diff
@@ -0,0 +1,27 @@
...
+
+<!--
+### Bugfixes
+
+* A bullet item for the Bugfixes category.
+
+-->
+### Enhancements
+
+* Add `globus groups set-policies` command for managing a groups policies
```
rudyardrichter, line 20:
```suggestion
* Add `globus groups set-policies` command for managing a group's policies
```

src/globus_cli/commands/group/_common.py
```diff
@@ -28,6 +28,18 @@ def parse_visibility(res):
...
+def parse_members_visibility(res):
+    return res["policies"]["group_members_visibility"]
+
+
+def parse_join_requests(res):
+    return res["policies"]["join_requests"]
+
+
+def parse_signup_fields(res):
+    return ",".join(sorted({f for f in res["policies"]["signup_fields"]}))
```
rudyardrichter, line 40:
```suggestion
    return ",".join(sorted(f for f in res["policies"]["signup_fields"]))
```

src/globus_cli/commands/group/set_policies.py
```diff
@@ -0,0 +1,109 @@
...
+    "phone",
+    "state",
+    "zip",
+]
+
+
+@click.option(
+    "--high-assurance/--not-high-assurance",
+    default=None,
+    help="Flag if the group will enforce high assurance access policies or not.",
```
rudyardrichter, line 30: I think I might find this wording a little clearer?
```suggestion
    help="Whether the group should enforce high assurance access policies or not.",
```

tests/functional/groups/test_group_set_policies.py
```diff
@@ -0,0 +1,78 @@
...
+def test_group_set_policies(run_line, option_data):
+    fixtures = load_response_set("get_and_set_group_policies")
+    group_id = fixtures.metadata["group_id"]
+
+    option, arg, field_name, expected_value = option_data
+    arg_str = f" {arg}" if arg else ""
+    result = run_line(f"globus group set-policies {option}{arg_str} {group_id}")
+    assert "Group policies updated successfully" in result.output
+
+    # TODO: expose get_last_request in globus_sdk._testing ?
```
sirosen, line 68: 👍 I've been thinking the same thing. Let's add it. The SDK testsuite already has a "get_last_request" helper which returns `None` if `responses.calls == []`.

tests/functional/groups/test_group_set_policies.py
```diff
@@ -0,0 +1,78 @@
...
+        ),
+    ],
+)
+def test_group_set_policies(run_line, option_data):
+    fixtures = load_response_set("get_and_set_group_policies")
+    group_id = fixtures.metadata["group_id"]
+
+    option, arg, field_name, expected_value = option_data
+    arg_str = f" {arg}" if arg else ""
+    result = run_line(f"globus group set-policies {option}{arg_str} {group_id}")
```
sirosen, line 65: I'm going to suggest a rewrite which is optional, but which I think makes this test easier to read.

`parametrize` can handle tuple unpacking into named arguments, and if we use `run_line(<list>)`, then we can take the options as an array of args instead of doing the `if arg` "slamming" to assemble a string.

```suggestion
@pytest.mark.parametrize(
    "add_args, field_name, expected_value",
    [
        (["--high-assurance"], "is_high_assurance", True),
        (["--no-high-assurance"], "is_high_assurance", False),
        (["--authentication-timeout", "100"], "authentication_assurance_timeout", 100),
        (["--visibility", "authenticated"], "group_visibility", "authenticated"),
        (["--members-visibility", "members"], "group_members_visibility", "members"),
        (["--join-requests"], "join_requests", True),
        (["--no-join-requests"], "join_requests", False),
        (["--signup-fields", "address"], "signup_fields", ["address"]),
        (
            ["--signup-fields", "address1,address2"],
            "signup_fields",
            ["address1", "address2"],
        ),
    ],
)
def test_group_set_policies(run_line, add_args, field_name, expected_value):
    fixtures = load_response_set("get_and_set_group_policies")
    group_id = fixtures.metadata["group_id"]

    result = run_line(["globus", "group", "set-policies", group_id] + add_args)
```

src/globus_cli/commands/group/set_policies.py
```diff
@@ -0,0 +1,109 @@
...
+    "city",
+    "country",
+    "current_project_name",
+    "department",
+    "field_of_science",
+    "institution",
+    "phone",
+    "state",
+    "zip",
+]
```
sirosen, line 24: I think we can remove this, as there's now `MEMBERSHIP_FIELDS` instead, for the same purpose.

```suggestion
```

tests/functional/groups/test_group_set_policies.py
```diff
@@ -0,0 +1,78 @@
...
+    assert "Group policies updated successfully" in result.output
+
+    # TODO: expose get_last_request in globus_sdk._testing ?
+    last_req = responses.calls[-1].request
+    body = json.loads(last_req.body)
+
+    # confirm expected put body values
+    get_response = fixtures.lookup("get_group_policies")
```
sirosen, line 73: I hadn't foreseen this kind of usage of `lookup()`. Cool! 😄

sirosen (commented):
There's one change which I think we should make (removing `SIGNUP_FIELDS`), and I have a suggestion for the test which I consider "nice but optional".

Overall this looks good though! As soon as `SIGNUP_FIELDS` is gone we can merge.

src/globus_cli/commands/group/set_policies.py
aaschaer: Oh whoops, good catch

tests/functional/groups/test_group_set_policies.py
aaschaer: I like this more, thank you

PR #609, opened 2022-03-04
Add 'group member list' command

sirosen:
Basic implementation of this command, with a `--fields` argument for including membership fields.

I've expanded `CommaDelimitedList` to take a choices argument and a `convert_values` callback, so that we can lowercase things by passing `str.lower`. These let us support `--fields` more nicely.

I chose to format the metavar for a comma-delimited list when there are choices in an imperfect way -- very open to ideas about that.

### DIFF
--- a/src/globus_cli/parsing/param_types/comma_delimited.py
+++ b/src/globus_cli/parsing/param_types/comma_delimited.py
@@ -1,14 +1,29 @@
+from typing import Callable, Iterable, Optional
+
 import click
 
 
 class CommaDelimitedList(click.ParamType):
+    def __init__(
+        self,
+        *,
+        convert_values: Optional[Callable[[str], str]] = None,
+        choices: Optional[Iterable[str]] = None,
+    ):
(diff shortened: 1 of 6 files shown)


src/globus_cli/parsing/param_types/comma_delimited.py
```diff
@@ -18,7 +33,18 @@ def convert(self, value, param, ctx):
...
+        resolved = value.split(",") if value else []
+
+        if self.convert_values is not None:
+            resolved = [self.convert_values(x) for x in resolved]
+
+        if self.choices is not None:
+            bad_values = [x for x in resolved if x not in self.choices]
+            if bad_values:
+                self.fail(
+                    f"the values {bad_values} were not valid choices",
```
aaschaer, line 45: Might be nice to have the choices in the error output?
But also could make the error wayy too long, so nbd

aaschaer (approved):
Left a minor comment, but this is looks good to go

PR #610, opened 2022-03-04
Convert the whole testsuite to '_testing' response sets

sirosen:
There are a few things we knew about this changeover just from having seen it in the SDK. The nicer interface, etc.

However, there was an unexpected benefit in that the testsuite now runs _significantly_ faster.
With `tox` overhead, it looks like a bit better than a 60% speedup. But it's actually even better than that:
command | python_version | old_time | new_time | speedup %
---|---|---|---|---
`tox` | all | 3m19s | 1m05s | 67.3%
`pytest` | 3.10 | 32.4s | 7.7s | 75.6%
`pytest` | 3.7 | 30.3s | 7.7s | 74.5%

(All times just from a just a few samples, not meant to be serious benchmarking data.)

I haven't done a thorough analysis. But if I had to guess why, my assumption would be that having all of the fixtures loaded from disk into memory only once is the thing that does it. Note that in `main`, every call to `load_api_fixtures` loads from disk. But `load_response_set` goes to the registry, which is purely in memory. There's one pass over the files to load them at the start of tests, and then it's done.

I'd like to compare again if we ever convert our fixture data from YAML to python, so that the importer is responsible for reading data and it goes to the registry without passing through a YAML parser. I bet it will be even faster.


rudyardrichter (approved):
Cool that the existing framework mapped pretty directly into this!

PR #611, opened 2022-03-08
Add `group invite [accept|decline]` commands

sirosen:
`invite [accept|decline]` commands for accepting or rejecting group invitations. They take the same form, either accepting an identity argument or operating over all invitations to a group.

Because the new tests feature rather extensive mock data in a fixture, the group tests are here also being split into multiple files.

---

A bit of background/meta about this PR: I realized that the way the webapp behaves makes it possible to invite a client ID to a group (this is easy). However, there's no way for the client ID to accept that invitation. With CLIENT_ID/CLIENT_SECRET support now in the CLI, if we add `invite accept|decline`, then a new potential user-facing workflow opens up.

### DIFF
--- a/src/globus_cli/commands/group/invite/_common.py
+++ b/src/globus_cli/commands/group/invite/_common.py
@@ -1,0 +1,51 @@
+import sys
+import uuid
+from typing import Dict, List, Optional
+
+import click
+import globus_sdk
+
+from globus_cli.parsing import ParsedIdentity
+from globus_cli.services.auth import CustomAuthClient
+
+if sys.version_info >= (3, 8):
+    from typing import Literal
+else:
+    from typing_extensions import Literal
+
+
+def get_invite_formatter(case: Literal["accept", "decline"]):
+    action_word = "Accepted" if case == "accept" else "Declined"
+
+    def formatter(data):
+        values = [f"{x['identity_id']} ({x['username']})" for x in data[case]]
+        if len(values) == 1:
+            click.echo(f"{action_word} invitation as {values[0]}")
+        else:
+            click.echo(f"{action_word} invitation as")
+            for v in values:
+                click.echo(f"  {v}")
+
+    return formatter
+
+
+def build_invite_actions(
+    auth_client: CustomAuthClient,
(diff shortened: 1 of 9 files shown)


src/globus_cli/commands/group/invite/_common.py
```diff
@@ -0,0 +1,51 @@
...
+from globus_cli.parsing import ParsedIdentity
+from globus_cli.services.auth import CustomAuthClient
+
+if sys.version_info >= (3, 8):
+    from typing import Literal
+else:
+    from typing_extensions import Literal
+
+
+def get_invite_formatter(case: Literal["accept", "decline"]):
```
rudyardrichter, line 17:
```suggestion
def get_invite_formatter(case: Literal["accept", "decline"]) -> Callable:
```

src/globus_cli/commands/group/invite/decline.py
```diff
@@ -0,0 +1,48 @@
...
+    actions = build_invite_actions(
+        auth_client, groups_client, "decline", group_id, identity
+    )
+    response = groups_client.batch_membership_action(group_id, actions)
+
+    # if this failed to return at least one accepted user, figure out an error to show
+    if not response.get("decline", None):
+        try:
+            raise ValueError(response["errors"]["decline"][0]["detail"])
+        except LookupError:
```
rudyardrichter, line 45: I never knew this was a thing :exploding_head:

sirosen: I'm still making the adjustment. It was added in py3.0, but the muscle-memory for `(KeyError, IndexError)` is still very strong. 😄

src/globus_cli/commands/group/invite/_common.py
sirosen: Yeah, perhaps I shouldn't be so lazy about annotations if I'm going to include them. I'll have to add the imports as well, and I'll make it `Callable[[GlobusHTTPRepsonse], None]`.

rudyardrichter (approved):

PR #612, opened 2022-03-09
Add 'group join' and 'group leave' commands

sirosen:


src/globus_cli/commands/group/join.py
```diff
@@ -0,0 +1,72 @@
...
+                f"Couldn't determine identity from value: {identity}"
+            )
+    else:
+        userinfo = auth_client.oauth2_userinfo()
+        identity_id = userinfo["sub"]
+
+    response = groups_client.batch_membership_action(
+        group_id, {action: [{"identity_id": identity_id}]}
+    )
+    # if this failed to return at least one 'leave', figure out an error to show
```
aaschaer, line 60: think the leave in this comment should be a join?

aaschaer (approved):
Looks good other than a small comment typo

src/globus_cli/commands/group/join.py
sirosen: Yep, thanks for the catch! I've amended.

PR #617, opened 2022-03-11
Move collection list args to "top level"

rudyardrichter:
Now that the SDK supports these arguments on `GCSClient.get_collection_list`, we can pass them directly: `mapped_collection_id`, `filter`.

### DIFF
--- a/src/globus_cli/commands/collection/list.py
+++ b/src/globus_cli/commands/collection/list.py
@@ -94,14 +94,13 @@ def collection_list(
     List the Collections on a given Globus Connect Server v5 Endpoint
     """
     gcs_client = login_manager.get_gcs_client(endpoint_id=endpoint_id)
-
     params = {}
+    if mapped_collection_id:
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -48,7 +48,7 @@ def read_readme():
     package_dir={"": "src"},
     python_requires=">=3.6",
     install_requires=[
-        "globus-sdk==3.3.1",
+        "globus-sdk==3.5.0",
```
sirosen, line 51: I'm surprised that this is not considered a conflict. I'm not 100% certain what `git blame` would show on `setup.py` after a merge.

sirosen (commented):
Can you rebase to the current `main`? I'm pretty sure that there would be no issue merging, but the `setup.py` change is odd and I'd rather see a clean diff before merge.

The change itself -- the part that really matters -- is good to go.

rudyardrichter:
Thought I rebased but it was missing some recent commits, let's try that again.

PR #621, opened 2022-03-16
Note that - can be used for query document filename

rudyardrichter:
Since `--query-document` is using `click.File`, we get the `-` behavior for free too (yay!), so add a note here to explain that.

I also noticed another site with a similar note has an extra space.


sirosen (approved):
👍 for both changes. Good catch on the extra space one!

PR #626, opened 2022-04-08
Add 'globus api' command

sirosen:
Right now, this is just a "first cut", and needs a bit more work. However, I was able to use it to do a few Transfer calls and it seems okay.

This is lacking in testing, there's no detailed documentation, and several parameters need more docs.
In no particular order, these are the TODO items which I still have on this:
- [x] add `--body-file`, make it mutex with `--body`
- [x] update SDK upstream to fix annotations to allow `data: str` for request bodies
- [x] add tests for content-type detection
- [x] add a `--header` option with `multiple=True`
- [x] set `app_name` on clients in the api commands to `"globus-cli-raw"` + CLI version info
- [x] add some basic tests

---

Update: This is now ready for review.

It add `globus api [transfer|auth|groups|search] [HEAD|GET|PUT|POST|DELETE] PATH`, with options for setting headers and query params with some basic parsing.

Some specifics on `Content-Type`:
Because many Globus APIs expect or require a content-type when reading JSON data, it seemed worthwhile to add some special handling for it. Namely, there is some simple Content-Type detection which will treat json-parseable data as `Content-Type: application/json`, and this can be controlled or fine-tuned with `--content-type`. It is applied to the headers _before_ `--header` values are parsed and handled, so that `--header "Content-Type: application/octet-stream"` works as desired/expected and overrides the autodetect behavior.

### DIFF
--- a/src/globus_cli/commands/api.py
+++ b/src/globus_cli/commands/api.py
@@ -25,15 +25,17 @@ class QueryParamType(click.ParamType):
 
 class HeaderParamType(click.ParamType):
     def get_metavar(self, param):
-        return '"Key: Value"'
+        return "Key:Value"
 
     def convert(self, value, param, ctx):
         value = super().convert(value, param, ctx)
         if value is None:
             return None
-        if ": " not in value:
+        if ":" not in value:
             self.fail("invalid header param", param=param, ctx=ctx)
-        left, right = value.split(": ", 1)
+        left, right = value.split(":", 1)
+        if right.startswith(" "):
+            right = right[1:]
         return (left, right)
 
 
--- a/src/globus_cli/commands/api.py
+++ b/src/globus_cli/commands/api.py
@@ -198,6 +198,7 @@ sends a 'GET' request to '{_get_url(service_name)}foo/bar'
         help="If the server responds with a redirect (a 3xx response with a Location "
         "header), follow the redirect. By default, redirects are not followed.",
     )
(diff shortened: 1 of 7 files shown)


src/globus_cli/commands/api.py
```diff
@@ -0,0 +1,217 @@
...
+        if body_file:
+            body = body_file.read()
+
+        headers_d = {}
+        if content_type != "none":
+            detected_content_type = detect_content_type(content_type, body)
+            if detected_content_type is not None:
+                headers_d["Content-Type"] = detected_content_type
+        for header_name, header_value in header:
+            headers_d[header_name] = header_value
```
sirosen, line 202: TODO: add a comment here that order of option evaluation (content-type before headers) matters

src/globus_cli/commands/api.py
```diff
@@ -0,0 +1,217 @@
...
+    ):
+        client = _get_client(login_manager, service_name)
+        client.app_name = version.app_name + " raw-api-command"
+
+        query_params_d = {}
+        for param_name, param_value in query_param:
+            query_params_d[param_name] = param_value
+
+        if body_file:
+            body = body_file.read()
```
sirosen, line 194: TODO: add a comment that `body is None` is okay and desired if neither option is given

sirosen:
I've just pushed an intentionally-failing commit which I'll squash/amend later. It serves to explain some of the work I'm about to open in a PR against globus-sdk.

PR #637, opened 2022-05-13
Add behavior like no-args-is-help to commands

sirosen:
Unlike the builtin no_args_is_help option, we do not want this behavior to trigger on commands which take no arguments (all args are optional).

In our custom command class, override the behavior of `parse_args` to check for MissingParameter errors combined with an empty arg list. Such a situation indicates that a command requires arguments (e.g. `globus transfer`) and none were given. In this situation, print the MissingParameter error message, followed by the full helptext for the current command, then exit(2) .

If there were arguments and a MissingParameter error is encountered, just reraise and trust click to handle it as normal.

---

@rudyardrichter, we discussed the variant of this in timer-cli. I considered that and almost lifted it into here. I didn't really like handling other errors beyond MissingParameter (easy to tweak), so I cut it down. But then I had to construct artificial Groups for my testing to ensure there would be a parent context (because that approach catches the errors outside of `make_context`). That "felt wrong" and manually touching usage parts with the help formatter was fiddly, so I started looking at [the no_args_is_help implementation](https://github.com/pallets/click/blob/c96545f6f4ba0eab99de6ec8b4ceb77c9bdb2528/src/click/core.py#L1370-L1372). That looks like a good place to hook in, so this is just wrapping the `click` behavior with a little bit of extra handling.

### DIFF
--- a/changelog.d/20220513_160219_sirosen_improve_cmd_help.md
+++ b/changelog.d/20220513_160219_sirosen_improve_cmd_help.md
@@ -1,0 +1,6 @@
+### Enhancements
+
+* Commands which have required arguments will print their helptext if invoked
+  with no arguments. They still `exit(2)` (usage error). This only applies to
+  the case of a command with required arguments being called with no arguments
+  at all.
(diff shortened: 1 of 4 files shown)


rudyardrichter (approved):
Looks great. Glad to see this added.

PR #638, opened 2022-05-13
Implement partial lazy importing in globus-cli

sirosen:
This relates very closely to https://github.com/globus/globus-sdk-python/pull/560 . Ideally, both improvements are combined when testing, resulting in the most significant possible performance improvement for commands like `globus --help` and tab completion.

---

Defer imports of the following stdlib modules until they are needed:

- webbrowser
- http.client
- http.server
- distutils
- shlex
- inspect
- pathlib
- configparser

And defer imports of the following 3rd party modules:

- jmespath
- cryptography

In concert with partial lazy importing in globus-sdk, this results in a dramatic improvement in the responsiveness of many commands. `globus --help` times fall as much as 80% , and completion becomes significantly more responsive without a deeper and more complex refactor (e.g. lazily loaded subcommands).

### DIFF
--- a/src/globus_cli/commands/endpoint/activate.py
+++ b/src/globus_cli/commands/endpoint/activate.py
@@ -1,4 +1,3 @@
-import webbrowser
 from typing import Dict, Optional, Union
 
 import click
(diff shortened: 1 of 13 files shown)


sirosen:
I'm closing this because I think #650 (which includes this) is a better "single step" to take. It delivers the whole speed improvement at once.

PR #640, opened 2022-05-19
Add support for 'localsdk' tox factor + env var

sirosen:
For testing against a local copy of the SDK repo.

Usage is like so:

    GLOBUS_SDK_PATH=../globus-sdk tox -e py39-localsdk

I refrained from adding a make target or anything else. You can also run a matrix as in

    tox -e 'py{37,38,39,310}-localsdk'


aaschaer (approved):
Amazing, thank you!

PR #643, opened 2022-05-23
Increase output width of all commands

rudyardrichter:
Use the `context_settings` argument to `click.Command` to adjust max content width to 80% of the terminal size.

### DIFF
--- a/src/globus_cli/parsing/commands.py
+++ b/src/globus_cli/parsing/commands.py
@@ -8,6 +8,7 @@ and all other components will be hidden internals.
 
 import logging
 import sys
+from shutil import get_terminal_size
 from typing import List
 
 import click
@@ -51,6 +52,15 @@ class GlobusCommand(click.Command):
             kwargs["help"] = helptext.format(
                 AUTOMATIC_ACTIVATION=self.AUTOMATIC_ACTIVATION_HELPTEXT
             )
+        if "context_settings" not in kwargs:
+            kwargs["context_settings"] = {}
+        if "max_content_width" not in kwargs["context_settings"]:
+            try:
+                cols = get_terminal_size(fallback=(80, 20)).columns
+                content_width = cols if cols < 100 else int(0.8 * cols)
+                kwargs["context_settings"]["max_content_width"] = content_width
+            except OSError:
+                pass
         super().__init__(*args, **kwargs)
 
     def invoke(self, ctx):
--- a/changelog.d/20220523_140701_rudyardrichter.md
+++ b/changelog.d/20220523_140701_rudyardrichter.md
@@ -1,0 +1,3 @@
+### Other
+
+* Increase the maximum width of help output to 80% of the terminal size.


src/globus_cli/parsing/commands.py
```diff
@@ -51,6 +52,13 @@ def __init__(self, *args, **kwargs):
             kwargs["help"] = helptext.format(
                 AUTOMATIC_ACTIVATION=self.AUTOMATIC_ACTIVATION_HELPTEXT
             )
+        if "context_settings" not in kwargs:
+            kwargs["context_settings"] = {}
+        if "max_content_width" not in kwargs["context_settings"]:
+            term_size = get_terminal_size(fallback=(80, 20))
+            kwargs["context_settings"]["max_content_width"] = int(
+                0.8 * term_size.columns
+            )
```
sirosen, line 61: One thing which surprised me with this change is that `globus transfer -h | cat` became _narrower_.
I'm wondering if we should "widen out" if the terminal width is small, which will also cover the fallback case.

So, something like (and don't mind exactly how I write this please 😉 ):
```python
cols = get_terminal_size(fallback=...).columns
if cols < 100: cols
else: cols * 0.8
```

Does that seem nice?

sirosen:
I dropped the helptext from `globus transfer -h` before and after into files for a side-by-side (copy-paste, to avoid shutil shenanigans):
```
$ wc -l narrow.txt
136 narrow.txt
$ wc -l wide.txt
82 wide.txt
```

I think that's a huge improvement, and it's equally striking with commands that have shorter helptext.

I have a change to the `--jmespath` helptext I plan to suggest as a follow-up, to make that option help fit on one line in more cases.

<details><summary>example helptext with shorter jmespath help</summary>

Compare this:
```
$ globus whoami -h
Usage: globus whoami [OPTIONS]

  Display information for the currently logged-in user.

Options:
  --linked-identities            Also show identities linked to the currently logged-in primary identity.
  -v, --verbose                  Control level of output
  -h, --help                     Show this message and exit.
  -F, --format [unix|json|text]  Output format for stdout. Defaults to text
  --jmespath, --jq TEXT          A JMESPath expression to apply to json output. Forces the format to be json processed by this expression
```

vs main:
```
$ globus whoami -h
Usage: globus whoami [OPTIONS]

  Display information for the currently logged-in user.

Options:
  --linked-identities            Also show identities linked to the currently
                                 logged-in primary identity.
  -v, --verbose                  Control level of output
  -h, --help                     Show this message and exit.
  -F, --format [unix|json|text]  Output format for stdout. Defaults to text
  --jmespath, --jq TEXT          A JMESPath expression to apply to json
                                 output. Forces the format to be json
                                 processed by this expression
```
</details>

I would like to have the text widen to full width on very narrow terminals -- since I think that's an easy improvement -- and then we can merge.

src/globus_cli/parsing/commands.py
rudyardrichter: Oh, right. The fallback will take over for the pipe case, so we don't want to again take 80% of that.

I also wonder whether about cases where `get_terminal_size` could fail; I figure it doesn't help to catch those.

sirosen: Just looking at the [py3.10 source for get_terminal_size](https://github.com/python/cpython/blob/251104f12d0cebb997f26f8ca3dac48202c7456a/Lib/shutil.py#L1378) I think this is already an "error handling variant" of `os.get_terminal_size(sys.stdout.fileno())`.

I would assume that it works as expected and we can handle any bug reports if someone finds a context where it does not. It doesn't look like there have been changes there since py3.5, so I bet it's pretty solid.

sirosen (approved):
I see the change I asked for is in. Although a somewhat low-key improvement, I'm quite excited for this!

PR #647, opened 2022-05-27
Eliminate a warning about _custom_text_formatter

rudyardrichter:
Pyright, which I'm using as an LSP extension, issues the warning here: `Object of type "None" cannot be called`. (I'm not certain if this corresponds to some stricter mypy setting?) In any case it's easy to eliminate & document the logic.


sirosen:
Yeah, I'm going to merge this straightaway and then we can think about how to handle `pyright`.

I've experimented with `pyright` on our repos in the past. IMO, it's the only non-`mypy` checker which we might want to make part of our linting. It's used on typeshed to validate the stubs, and the devs are very much included in typing discussions with the mypy maintainers. So it's much more realistic to support than, for example, `pyre`. And LSP integration gives us a motivation to keep it happy.

PR #648, opened 2022-05-27
Add `FORMAT_TEXT_RECORD_LIST` for `formatted_print`

rudyardrichter:
Add new output format that looks like:
```
Job ID: 1234...
X: Y
foo: bar

Job ID: 8976...
X: Z
foo: baz
```
...as well as a simple test.


rudyardrichter:
(force-push to add the `assert "" in output.splitlines()` statement to the test)

tests/unit/test_termio.py
```diff
@@ -30,3 +36,20 @@ def test_term_interactive(ps1, force_flag, expect, monkeypatch):
...
+
+
+def test_format_record_list(capsys):
+    data = [
+        {"bird": "Killdeer", "wingspan": 46},
+        {"bird": "Franklin's Gull", "wingspan": 91},
+    ]
+    fields = [("Bird", "bird"), ("Wingspan", "wingspan")]
+    with click.Context(click.Command("fake-command")) as _:
+        formatted_print(data, text_format=FORMAT_TEXT_RECORD_LIST, fields=fields)
```
sirosen, line 48: I'm perfectly happy with this, if you want to keep it, but I'll point out some other tests we have which define a click command _inline_ in order to check behavior:
https://github.com/globus/globus-cli/blob/7db15fbb09d3bb3f6e6aa22f04520936af0b99cf/tests/unit/test_param_types.py#L24

If you prefer that, feel free to switch.

Otherwise, I wonder if `with click.Command("foo").make_context():` reads a bit better.

tests/unit/test_termio.py
```diff
@@ -30,3 +36,20 @@ def test_term_interactive(ps1, force_flag, expect, monkeypatch):
         monkeypatch.delitem(os.environ, "GLOBUS_CLI_INTERACTIVE", raising=False)
 
     assert term_is_interactive() == expect
+
+
+def test_format_record_list(capsys):
+    data = [
+        {"bird": "Killdeer", "wingspan": 46},
+        {"bird": "Franklin's Gull", "wingspan": 91},
+    ]
```
sirosen, line 45: I'm disappointed that you chose not to include the scientific names, flavor text, or ability text. 😜

src/globus_cli/termio/output_formatter.py
```diff
@@ -282,6 +284,17 @@ def _print_as_text():
         elif text_format == FORMAT_TEXT_RECORD:
             _assert_fields()
             colon_formatted_print(data, fields)
+        elif text_format == FORMAT_TEXT_RECORD_LIST:
+            _assert_fields()
+            if not isinstance(data, list):
+                raise ValueError("only lists can be output in text record list format")
```
sirosen, line 290: The requirement for `list` is the only thing that makes me hesitate here. If we were to pass a paginated call, it would be cool to be able to stream output.

Maybe it's better to check `collections.abc.Iterable`?

We can leave this as-is until we have a use-case for a non-list iterable.

sirosen (approved):
I'm happy with this (Approved™️), but I also have some thoughts on refinements I shared. Changes optional.

tests/unit/test_termio.py
rudyardrichter: Yeah, I felt like this was a little more clear since we're not concerned with any command setup, just what gets printed.

tests/unit/test_termio.py
rudyardrichter: You know what they say about programmers and laziness...and it seems like the reference still got across :wink:

src/globus_cli/termio/output_formatter.py
rudyardrichter: Agh, yeah. The thing that makes me hesitate with Iterable (or trying `iter()`) is that it allows dicts, which I think is against the intent here. I guess `isinstance(Iterable) and not isinstance(dict)`?

rudyardrichter: What do you think of this rendition?

tests/unit/test_termio.py
sirosen: I wasn't sure until I checked that those numbers are the wingspans on the cards. I'm glad to have it in our test data! Look out for Tardigrades and Deimos Down in my PRs.

tests/unit/test_termio.py
sirosen: Sounds good to me. Marking resolved.

src/globus_cli/termio/output_formatter.py
sirosen: To be straight with you, I prefer the original.
When it's done via `iter(x)` with the try-catch, it's a lot of cognitive/reading overhead for this. Is `isinstance(x, collections.abc.Iterable)` not viable? Have I forgotten that being missing in py3.6 or something?

There are other mapping types and mapping-likes, `GlobusHTTPResponse` in particular, which can get passed. If we're not going to accept an arbitrary iterable, I'd probably just stick with `list` for today and we can always revisit and expand it later.

rudyardrichter: I vaguely recall reading some encouragement to use the try-iter version since it was more reliable than `Iterable`, but that's probably outdated. I think I'm operating on superstition here.

I'm realizing also `collections.abc.Sequence` is very close to what we want, but that would require also adding a `__len__` to `IterableResponse`. I'm good to just assert list here for now and revisit.

sirosen: FWIW, you're not alone. I've definitely done things in the past because it's what we had to do for py2.6 compatibility, and I never "re-learned" how to do it. With the rate of changes in typing, I suspect that for some time we'll be (re)learning how we're "supposed" to express iterable, comparable, mapping, frozen, truthy/falsey, and a wide array of other concepts that loosely fit `typing.Protocol`.

PR #653, opened 2022-06-08
Remove support for python3.6

sirosen:
- Remove from tox and CI
- Update python_requires in setup.py
- Adjust several builds to use py3.7 as our new "minimum version"

---

The plan here is that we drop 3.6, we do a release with this and the other minor tweaks that are in `main`, and we assess impact. That puts us in a position to start working with 3.7+ features, but without doing any major refactoring that we'd have to unwind if we get (unexpected) blowback from dropping this EOL version.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -27,7 +27,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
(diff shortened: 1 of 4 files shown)


rudyardrichter (approved):
Looks like this covers all the changes needed!

PR #656, opened 2022-06-15
add user-credential commands and storage-gateway list

aaschaer:
These are the commands I used for live API testing SDK support, and are not in a state ready to merge. Opening a draft PR anyways since I think it would be good to discuss a few usage things:

1. Are create commands useful without support for connector specific policy docs? Do we need `globus user-credential create posix`? Or is allowing passing arbitrary policies to `globus user-credential create` useful for supporting future connectors the CLI doesn't know about yet?

2. Is passing an endpoint_id the best way to discover the gcs manager in this case? Or should we ask users to instead pass a mapped_collection_id? That would let us automatically discover the storage gateway, which would make this more end user friendly, but listing all credentials on an endpoint might still be useful?

3. Somewhat related to 2, should these commands be top level commands, or should they live under endpoint and/or collection?


sirosen:
Sorry it took so long to get you some feedback on this.

> * Are create commands useful without support for connector specific policy docs? Do we need `globus user-credential create posix`? Or is allowing passing arbitrary policies to `globus user-credential create` useful for supporting future connectors the CLI doesn't know about yet?

I think both modes of use are potentially helpful for users, but they aren't exclusive. I would focus on the 80% case for this: `globus user-credential create posix`. Once we have that, it is possible to add others alongside it. Including `globus user-credential create json ...` for a passthrough. (Other words for `json` in that context: `raw`, `any`, `from-document`, etc.)

> * Is passing an endpoint_id the best way to discover the gcs manager in this case? Or should we ask users to instead pass a mapped_collection_id? That would let us automatically discover the storage gateway, which would make this more end user friendly, but listing all credentials on an endpoint might still be useful?

Again, this seems like a case in which both modes are useful. What about exposing it as a filtering option?
```
globus user-credential list $ENDPOINT_ID --filter-mapped-collection $COLLECTION_ID
```

It may be verbose, but it's unambiguous, it supports both desirable usages, and it doesn't look costly to maintain.

> * Somewhat related to 2, should these commands be top level commands, or should they live under endpoint and/or collection?

I don't like making this a top-level command. It serves a relatively small niche use-case and is part of the GCSv5 ecosystem of commands.

I think that as a user, it would be _ideal_ if I could use
```
1. globus collection user-credential list $COLLECTiON_ID  # alias for 3
2. globus endpoint user-credential list $ENDPOINT_ID
3. globus endpoint user-credential list $ENDPOINT_ID --filter-mapped-collection $COLLECTION_ID
```

However, I don't think that's worth doing in an initial effort because (1) is just a convenience alias. I would be inclined to implement it as a command under `endpoint` for now, and we can add the `collection` alias in the future if we ever feel the need.

sirosen:
Our conclusion from discussion:

- `globus endpoint user-credential` is the namespace
- these are the list commands:
```
globus endpoint user-credential list $ENDPOINT_ID
globus endpoint user-credential list $ENDPOINT_ID --filter-mapped-collection $COLLECTION_ID
```
- `globus endpoint user-credential create posix` is one creation command (included)
- `globus endpoint user-credential create from-json` is another creation command (included)
- all other create commands left as future work
- `globus endpoint user-credential update from-json` is included, BUT hidden (`update` should be hidden)

src/globus_cli/commands/endpoint/user_credential/update/from_json.py
```diff
@@ -0,0 +1,23 @@
...
+    *,
+    login_manager: LoginManager,
+    endpoint_id,
+    user_credential_json,
+):
+    """
+    Update a User Credential on an endpoint with a JSON document
+    """
+    gcs_client = login_manager.get_gcs_client(endpoint_id=endpoint_id)
+    res = gcs_client.create_user_credential(user_credential_json)
```
sirosen, line 22:
```suggestion
    res = gcs_client.update_user_credential(user_credential_json)
```

aaschaer:
This should (finally) be ready for review!

src/globus_cli/commands/endpoint/user_credential/create/posix.py
```diff
@@ -0,0 +1,37 @@
...
+@user_credential_create_and_update_params(create=True)
+@LoginManager.requires_login(LoginManager.TRANSFER_RS, LoginManager.AUTH_RS)
+def posix(
+    *,
+    login_manager: LoginManager,
+    endpoint_id,
+    storage_gateway,
+    globus_identity,
+    local_username,
+    **kwargs,  # values from user_credential_create_and_update_params
```
sirosen, line 21: How much would it take to eliminate `**kwargs` usage here?
I've been trying to remove/avoid `**kwargs` for our commands because it doesn't play nicely with type checking.

In this case, I think we have `display_name: str | None` and nothing else to add.

src/globus_cli/commands/endpoint/user_credential/create/s3.py
```diff
@@ -0,0 +1,50 @@
...
+def s3(
+    *,
+    login_manager: LoginManager,
+    endpoint_id,
+    storage_gateway,
+    globus_identity,
+    local_username,
+    s3_key_id,
+    s3_secret_key,
+    **kwargs,  # values from user_credential_create_and_update_params
```
sirosen, line 26: As in the other case of `**kwargs` usage, I think this is just `display_name: str | None`.

src/globus_cli/commands/endpoint/user_credential/create/posix.py
sirosen: Just to note: if you take that suggestion and use `str | None`, make sure to add `from __future__ import annotations`.

sirosen (changes requested):
I've gone over things pretty carefully and I don't see anything that's really an issue. It looks good to go!

There is one nitpick that I am going to hold this on -- I'd like us to avoid `**kwargs` where possible because it results in the SDK annotations not being useful to check our usages.

src/globus_cli/commands/endpoint/user_credential/create/posix.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+@command("posix", short_help="Create a User Credential for a POSIX storage gateway")
+@endpoint_id_arg
+@user_credential_create_and_update_params(create=True)
+@LoginManager.requires_login(LoginManager.TRANSFER_RS, LoginManager.AUTH_RS)
+def posix(
+    *,
+    login_manager: LoginManager,
+    endpoint_id: str,
```
sirosen, line 19:
```suggestion
    endpoint_id: uuid.UUID,
```

Since `endpoint_id_arg` passes a UUID object, we should tweak this (and you'll need the appropriate import).

Right now this is causing a mypy failure because `get_gcs_client` expects a UUID value.

PR #659, opened 2022-06-30
Add timer create, update, delete & tests

rudyardrichter:
- Labeling this no-news since the commands are hidden.
- Increment SDK requirement to 3.10.0.
- Add create, delete, and update commands for timer.
  - As discussed/planned these commands are hidden for now.

### DIFF
--- a/src/globus_cli/commands/timer/_common.py
+++ b/src/globus_cli/commands/timer/_common.py
@@ -7,7 +7,6 @@ from typing import Any, Dict, Generator, List, Optional, Union
 from urllib.parse import urlparse
 
 import click
-from globus_sdk import GlobusError, TransferClient
 
 # List of datetime formats accepted as input. (`%z` means timezone.)
 DATETIME_FORMATS = [
@@ -113,10 +112,7 @@ def parse_timedelta(s: str) -> datetime.timedelta:
     matches = timedelta_regex.match(s)
     if not matches:
         raise ValueError(f"couldn't parse timedelta from string: {s}")
-    groups = {k: int(v) for k, v in matches.groupdict(0).items()}
-    # timedelta accepts kwargs for units up through days, have to convert weeks
-    groups["days"] += groups.pop("weeks", 0) * 7
-    return datetime.timedelta(**groups)
+    return datetime.timedelta(**{k: int(v) for k, v in matches.groupdict(0).items()})
 
 
 def read_csv(
@@ -145,43 +141,3 @@ def read_csv(
         reader = DictReader(decomment(f), fieldnames=fieldnames)
         for row_dict in reader:
             yield {k: transform_val(k, v) for k, v in row_dict.items()}
-
-
-def require_activated_endpoints(
-    transfer_client: TransferClient,
-    endpoints: List[str],
-    reactivate_if_expires_in: int = 86400,
-):
-    not_activated = []
-    for endpoint in endpoints:
-        try:
-            if not transfer_client.get_endpoint(endpoint).get("activated"):
-                not_activated.append(endpoint)
-        except GlobusError as e:
-            err_msg = "couldn't get information for endpoint {endpoint}"
-            code = getattr(e, "code", None)
-            msg = getattr(e, "message", None)
-            if code and msg:
-                err_msg += f": {code} {msg}"
-            click.echo(err_msg, err=True)
-            sys.exit(1)
-    still_not_activated = []
-    for endpoint in not_activated:
-        response = transfer_client.endpoint_autoactivate(
-            endpoint, if_expires_in=reactivate_if_expires_in
-        )
-        if response.get("code") == "AutoActivationFailed":
-            still_not_activated.append(endpoint)
-    if still_not_activated:
-        show_endpoints = ", ".join(still_not_activated)
-        click.echo(
-            f"Error: requested endpoint is not activated: {show_endpoints}\n"
-            "Open in the web app to activate:",
-            err=True,
-        )
-        for endpoint in still_not_activated:
-            click.echo(
-                f"    https://app.globus.org/file-manager?origin_id={endpoint}",
-                err=True,
-            )
-        sys.exit(1)
--- a/src/globus_cli/commands/timer/_common.py
+++ b/src/globus_cli/commands/timer/_common.py
@@ -1,7 +1,13 @@
 import datetime
 import re
-from typing import Any, Dict, Optional
+import sys
+from csv import DictReader
+from distutils.util import strtobool
+from typing import Any, Dict, Generator, List, Optional, Union
 from urllib.parse import urlparse
+
+import click
+from globus_sdk import GlobusError, TransferClient
 
 # List of datetime formats accepted as input. (`%z` means timezone.)
 DATETIME_FORMATS = [
@@ -72,6 +78,17 @@ JOB_FORMAT_FIELDS = [
     ("N. Timer Errors", lambda data: data["n_errors"]),
 ]
 
+DELETED_JOB_FORMAT_FIELDS = [
+    ("Job ID", "job_id"),
+    ("Name", "name"),
+    ("Type", _get_action_type),
+    ("Submitted At", lambda data: isoformat_to_local(data["submitted_at"])),
+    ("Start", lambda data: isoformat_to_local(data["start"])),
+    ("Interval", _get_interval),
+    ("Stop After Date", _get_stop_date),
+    ("Stop After N. Runs", _get_stop_n_runs),
+]
+
 START_HELP = """
 Start time for the job. Defaults to current time. (The example above shows the allowed
 formats using Python's datetime formatters; see:
@@ -79,7 +96,7 @@ https://docs.python.org/3/library/datetime.html #strftime-and-strptime-format-co
 """
 
 timedelta_regex = re.compile(
-    r"""\
+    r"""
     ^
     ((?P<weeks>\d+)w)?
     ((?P<days>\d+)d)?
@@ -90,3 +107,81 @@ timedelta_regex = re.compile(
     """,
     flags=re.VERBOSE,
 )
+
+
+def parse_timedelta(s: str) -> datetime.timedelta:
+    matches = timedelta_regex.match(s)
+    if not matches:
+        raise ValueError(f"couldn't parse timedelta from string: {s}")
+    groups = {k: int(v) for k, v in matches.groupdict(0).items()}
+    # timedelta accepts kwargs for units up through days, have to convert weeks
+    groups["days"] += groups.pop("weeks", 0) * 7
+    return datetime.timedelta(**groups)
+
+
+def read_csv(
+    file_name: str,
+    fieldnames: List[str],
+    comment_char: str = "#",
+) -> Generator[Dict[str, Union[str, bool]], None, None]:
+    def decomment(f):
+        for row in f:
+            if not row.startswith(comment_char):
+                yield row
+
+    def transform_val(k: str, v: str) -> Union[str, bool]:
+        v = v.strip()
+        if k == "recursive":
+            try:
+                return bool(strtobool(v))
+            except ValueError:
+                # "invalid truth value"
+                click.echo(f"In file {file_name}: couldn't parse {v} as a boolean")
+                sys.exit(1)
+        else:
+            return v
+
+    with open(file_name) as f:
+        reader = DictReader(decomment(f), fieldnames=fieldnames)
+        for row_dict in reader:
+            yield {k: transform_val(k, v) for k, v in row_dict.items()}
+
+
+def require_activated_endpoints(
+    transfer_client: TransferClient,
+    endpoints: List[str],
+    reactivate_if_expires_in: int = 86400,
+):
+    not_activated = []
+    for endpoint in endpoints:
+        try:
+            if not transfer_client.get_endpoint(endpoint).get("activated"):
+                not_activated.append(endpoint)
+        except GlobusError as e:
+            err_msg = "couldn't get information for endpoint {endpoint}"
+            code = getattr(e, "code", None)
+            msg = getattr(e, "message", None)
+            if code and msg:
+                err_msg += f": {code} {msg}"
+            click.echo(err_msg, err=True)
+            sys.exit(1)
+    still_not_activated = []
+    for endpoint in not_activated:
+        response = transfer_client.endpoint_autoactivate(
+            endpoint, if_expires_in=reactivate_if_expires_in
+        )
+        if response.get("code") == "AutoActivationFailed":
+            still_not_activated.append(endpoint)
+    if still_not_activated:
+        show_endpoints = ", ".join(still_not_activated)
+        click.echo(
+            f"Error: requested endpoint is not activated: {show_endpoints}\n"
+            "Open in the web app to activate:",
+            err=True,
+        )
+        for endpoint in still_not_activated:
+            click.echo(
+                f"    https://app.globus.org/file-manager?origin_id={endpoint}",
+                err=True,
+            )
+        sys.exit(1)
--- a/src/globus_cli/commands/timer/create.py
+++ b/src/globus_cli/commands/timer/create.py
@@ -1,9 +1,11 @@
 import datetime
+from enum import Enum
 from typing import List, Optional, Tuple, cast
 
 import click
 from globus_sdk import TimerJob, TransferData
 
+from globus_cli.commands.transfer import autoactivate
 from globus_cli.login_manager import LoginManager
 from globus_cli.parsing import command, group, mutex_option_group
 from globus_cli.termio import FORMAT_TEXT_RECORD, formatted_print
@@ -14,7 +16,6 @@ from ._common import (
     START_HELP,
     parse_timedelta,
     read_csv,
-    require_activated_endpoints,
 )
 
 INTERVAL_HELP = """
@@ -23,6 +24,13 @@ INTERVAL_HELP = """
     '24h', '1d 12h', '2w', etc. Must be in order: hours -> minutes -> seconds. You
     should either use quotes ('1d 2h') or write without spaces (1d2h).
 """
+
+
+class SyncLevel(Enum):
+    exists = 0
+    size = 1
+    mtime = 2
+    checksum = 3
 
 
 @group("create", short_help="Submit a Timer job", hidden=True)
@@ -85,15 +93,11 @@ def create_command():
 @click.option(
     "--sync-level",
     required=False,
-    type=int,
+    default=None,
+    type=click.Choice(("exists", "size", "mtime", "checksum"), case_sensitive=False),
     help=(
-        "Specify that only new or modified files should be transferred. The behavior"
-        " depends on the value of this parameter, which must be a value 0–3, as"
-        " defined in the transfer API: 0. Copy files that do not exist at the"
-        " destination. 1. Copy files if the size of the destination does not match the"
-        " size of the source. 2. Copy files if the timestamp of the destination is"
-        " older than the timestamp of the source. 3. Copy files if checksums of the"
-        " source and destination do not match."
+        "Specify that only new or modified files should be transferred, depending on"
+        " which setting is provided"
     ),
 )
 @click.option(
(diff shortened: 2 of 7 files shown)


src/globus_cli/commands/timer/_common.py
```diff
@@ -90,3 +107,81 @@ def isoformat_to_local(
...
+    def decomment(f):
+        for row in f:
+            if not row.startswith(comment_char):
+                yield row
+
+    def transform_val(k: str, v: str) -> Union[str, bool]:
+        v = v.strip()
+        if k == "recursive":
+            try:
+                return bool(strtobool(v))
```
sirosen, line 136: If we keep this (rather than following the suggestion to use the `--batch` processing below), then I'd like to see us not use `distutils` because it's deprecated. I can read this section more carefully if we go down this route.

src/globus_cli/commands/timer/_common.py
```diff
@@ -90,3 +107,81 @@ def isoformat_to_local(
...
 )
+
+
+def parse_timedelta(s: str) -> datetime.timedelta:
+    matches = timedelta_regex.match(s)
+    if not matches:
+        raise ValueError(f"couldn't parse timedelta from string: {s}")
+    groups = {k: int(v) for k, v in matches.groupdict(0).items()}
+    # timedelta accepts kwargs for units up through days, have to convert weeks
+    groups["days"] += groups.pop("weeks", 0) * 7
```
sirosen, line 118: This might be evidence of an older python version?
3.7 has `weeks=...` support:
https://github.com/python/cpython/blob/239b2d9b4703ea5301dab698fe922de997973476/Lib/datetime.py#L474

Overall though, I really like how this regex + parse function reads. It's very legible.

src/globus_cli/commands/timer/_common.py
```diff
@@ -90,3 +107,81 @@ def isoformat_to_local(
...
+
+def require_activated_endpoints(
+    transfer_client: TransferClient,
+    endpoints: List[str],
+    reactivate_if_expires_in: int = 86400,
+):
+    not_activated = []
+    for endpoint in endpoints:
+        try:
+            if not transfer_client.get_endpoint(endpoint).get("activated"):
```
sirosen, line 158: This function, and this line in particular, is the main thing I want to revisit in this PR.

There are dedicated APIs and [SDK support](https://github.com/globus/globus-sdk-python/blob/56085ab4aaf6b6d263313799f94e4388f77ed68e/src/globus_sdk/services/transfer/response/activation.py#L7) for handling activation. I'm looking at how [`globus endpoint is-activated`](https://github.com/globus/globus-cli/blob/b16ef4c18fcfbfa1c76136209ba76b3223523480/src/globus_cli/commands/endpoint/is_activated.py#L100) is implemented, but I also think we can get away with the `endpoint_autoactivate` call as our check.

I think we should take the same approach that other CLI commands take:
- attempt autoactivate (with `if_expires_in=...`)
- if `AutoActivationFailed` is returned, treat that as the endpoint not being activated (for most other commands, that's [an error](https://github.com/globus/globus-cli/blob/b16ef4c18fcfbfa1c76136209ba76b3223523480/src/globus_cli/services/transfer/activation.py#L91))

While we're tinkering with this, other changes worth considering...
- Do we need special handling for the activation call failing? I would tend to say "No" until proven otherwise. The Transfer API should give a good error with an informative message, and the CLI already has top-level handling for these errors.
- Rather than suggesting opening the webapp as the only resolution, what about `globus endpoint activate`? Or perhaps code can be shared with [`globus endpoint activate --web`](https://github.com/globus/globus-cli/blob/b16ef4c18fcfbfa1c76136209ba76b3223523480/src/globus_cli/commands/endpoint/activate.py#L280)?

src/globus_cli/commands/timer/create.py
```diff
@@ -0,0 +1,288 @@
...
+@click.option(
+    "--stop-after-runs",
+    required=False,
+    type=int,
+    help=("Stop running the transfer after this number of runs have happened"),
+)
+@click.option(
+    "--sync-level",
+    required=False,
+    type=int,
```
sirosen, line 88: Rather than using an `int` and requiring that users learn the Transfer enum, can we share code with `globus transfer` for this?
https://github.com/globus/globus-cli/blob/b16ef4c18fcfbfa1c76136209ba76b3223523480/src/globus_cli/commands/transfer.py#L122

src/globus_cli/commands/timer/create.py
```diff
@@ -0,0 +1,288 @@
...
+    required=True,
+    type=str,
+    help="ID for the source transfer endpoint",
+)
+@click.option(
+    "--dest-endpoint",
+    required=True,
+    type=str,
+    help="ID for the destination transfer endpoint",
+)
```
sirosen, line 63: I would like the src and dest to be ENDPOINT_PLUS_OPTPATH items, so that they have the same colon-delimited syntax used throughout the rest of the CLI.
For an example, see: https://github.com/globus/globus-cli/blob/b16ef4c18fcfbfa1c76136209ba76b3223523480/src/globus_cli/commands/transfer.py#L212-L217

This ties in with a note I have below on `--item`.

Additionally, I would strongly prefer that these arguments be positionals, rather than `required=True` options.
We only use required optionals where necessary (or if there's been a mistake), but positional arguments more clearly describe themselves to a user e.g. in usage and synopsis.

src/globus_cli/commands/timer/create.py
```diff
@@ -0,0 +1,288 @@
...
+    required=False,
+    type=(str, str, bool),
+    multiple=True,
+    help=(
+        "Used to specify the transfer items; provide as many of this option as files"
+        " to transfer. The format for this option is `--item SRC DST RECURSIVE`, where"
+        " RECURSIVE specifies, if this item is a directory, to transfer the entire"
+        " directory. For example: `--item ~/file1.txt ~/new_file1.txt false`"
+    ),
+)
```
sirosen, line 135: Is this option necessary?

I don't mean to pretend that the interface we offer in `globus endpoint transfer` is perfect, but it seems to be working well enough. We therefore need strong justification for introducing a new and different interface for an activity which is very similar and takes the same inputs.

src/globus_cli/commands/timer/create.py
```diff
@@ -0,0 +1,288 @@
...
+    required=False,
+    type=click.Choice(["ignore", "keep", "copy"], case_sensitive=False),
+    default="ignore",
+)
+@click.option(
+    "--items-file",
+    required=False,
+    type=str,
+    help="file containing table of items to transfer",
+)
```
sirosen, line 162: If we accept my argument above about `--item`, and internal consistency of the CLI, then this should become `--batch`.

src/globus_cli/commands/timer/create.py
```diff
@@ -0,0 +1,288 @@
...
+    is_flag=True,
+    default=False,
+    help="Cancel the submitted transfer tasks if QUOTA_EXCEEDED errors are hit.",
+)
+@click.option(
+    "--recursive-symlinks",
+    required=False,
+    type=click.Choice(["ignore", "keep", "copy"], case_sensitive=False),
+    default="ignore",
+)
```
sirosen, line 156: Although this option is supported by Transfer, it doesn't do anything (except on a very small number of special endpoints). We should therefore not expose it here -- it has caused confusion in the past, so the docs now have a special note about it.

(And yes, we might have it in other places where it needs to be removed.)

src/globus_cli/commands/timer/create.py
```diff
@@ -0,0 +1,288 @@
...
+@click.option(
+    "--start",
+    required=False,
+    type=click.DateTime(formats=DATETIME_FORMATS),
+    help=START_HELP,
+)
+@click.option(
+    "--interval",
+    required=False,
+    type=str,
```
sirosen, line 49: I'm not sure I want to pursue this right this minute, but it seems like instead of a `str`, we could do the parsing at `click` parse-time, and have a custom type here to do the work.

src/globus_cli/commands/timer/create.py
```diff
@@ -0,0 +1,288 @@
...
+    default=True,
+    help="Send notification email on task failure",
+)
+@click.option(
+    "--notify-on-inactive",
+    is_flag=True,
+    default=True,
+    help="Send notification email on task becoming inactive",
+)
+@mutex_option_group("--action-file", "--action-body")
```
sirosen, line 181: I think these are not included here, and the mutex opt group should be removed?

src/globus_cli/commands/timer/create.py
```diff
@@ -0,0 +1,288 @@
...
+    Interval at which the job should run. Use 'w', 'd', 'h', 'm', and 's' as suffixes
+    to specify weeks, days, hours, minutes, and seconds. Examples: '1h 30m', '500s',
+    '24h', '1d 12h', '2w', etc. Must be in order: hours -> minutes -> seconds. You
+    should either use quotes ('1d 2h') or write without spaces (1d2h).
+"""
+
+
+@group("create", short_help="Submit a Timer job", hidden=True)
+def create_command():
+    pass
```
sirosen, line 30: Not strictly necessary at the moment, but at some point I'd like to move this to make `create/` a subpackage and `create/transfer.py` the implementation.

Very very early in CLI development we tried some other structures for things, but this is the one which we found was intelligible to people because the layout (on disk) mirrors the usages.

src/globus_cli/commands/timer/delete.py
```diff
@@ -0,0 +1,35 @@
...
+        deleted = timer_client.delete_job(job_id)
+        if first:
+            first = False
+        else:
+            click.echo()
+        formatted_print(
+            deleted,
+            text_format=FORMAT_TEXT_RECORD_LIST,
+            fields=DELETED_JOB_FORMAT_FIELDS,
+        )
```
sirosen, line 35: If run with `-Fjson`, this will potentially produce invalid JSON as its output.

If it's very important to support bulk-deletes, there are options for this (including adding API support), but for now, why not just go with `JOB_ID` (singular)?

src/globus_cli/commands/timer/update.py
```diff
@@ -0,0 +1,56 @@
...
+    submitted—only the name, interval, and stop conditions.
+    """
+    timer_client = login_manager.get_timer_client()
+    data: Dict[str, Any] = {}
+    if name:
+        data["name"] = name
+    if stop_after_date:
+        data["stop_after"] = {"date": stop_after_date.isoformat()}
+    if stop_after_runs:
+        data["stop_after"] = {**data.get("stop_after", {}), "n_runs": stop_after_runs}
```
sirosen, line 51: This feels awkward, but I get it... How do you feel about `defaultdict(dict)`? Better, worse, or about the same?

Another option, to the tune of "super verbose, but super clear":

```suggestion
    if stop_after_date or stop_after_runs:
        data["stop_after"] = {}
        if stop_after_date:
            data["stop_after"]["date"] = stop_after_date.isoformat()
        if stop_after_runs:
            data["stop_after"]["n_runs"] = stop_after_runs
```

src/globus_cli/commands/timer/update.py
```diff
@@ -0,0 +1,56 @@
...
+    Start time and parameters such as transfer options cannot be changed once a job is
+    submitted—only the name, interval, and stop conditions.
+    """
+    timer_client = login_manager.get_timer_client()
+    data: Dict[str, Any] = {}
+    if name:
+        data["name"] = name
+    if stop_after_date:
+        data["stop_after"] = {"date": stop_after_date.isoformat()}
+    if stop_after_runs:
```
sirosen, line 50: What happens with `--stop-after-runs 0`?

sirosen (changes requested):
My overall feeling right now is that we can take steps to make the interface from `globus timer create transfer` both more uniform on the surface with `globus transfer` and to make the code share more between the two commands.

The tests look good -- but if we're going to keep the CSV behavior it needs to be tested. And for the most part the commands look good to me as well, although I've dropped some inline notes on bits I think we should tweak.

But I have noted a couple of things which I think fit more into the category of "blocking changes", mostly on the transfer command.

src/globus_cli/commands/timer/create.py
rudyardrichter: Yeah, I think I'm in agreement with both of these? I think this slightly predated the CLI feature of reading from files here so to accomplish the csv-file interface it looked different originally. The only thing is that the timer interface is unable to care about more than the `recursive` setting. Do you think it would make sense to still allow writing just a boolean in the file instead of the flag?

src/globus_cli/commands/timer/create.py
rudyardrichter: Excellent!

src/globus_cli/commands/timer/delete.py
rudyardrichter: Yeah, this was a workaround for `-F json` because it doesn't work with the list type. An envelope of some sort would make more sense in that respect but wouldn't reflect anything on the API side.

We can go with singular deletes, but it seems odd to have the `nargs="+"` behavior be obnoxious to support.

src/globus_cli/commands/timer/update.py
rudyardrichter: The suggestion is much clearer at the expense of 2 lines, I'll definitely go with that.

src/globus_cli/commands/timer/update.py
rudyardrichter: That's a fun question. I think we could be more aggressive about errors on the API side; as it is the job would simply go immediately to the "ended" state.

src/globus_cli/commands/timer/_common.py
rudyardrichter: Indeed, a relic of 3.6 support.

src/globus_cli/commands/timer/create.py
rudyardrichter: Yeah, it would be nice, though likewise I think maybe that's a good target for a future refactor.

rudyardrichter:
So it seems like the moral of the story is to homogenize this with the existing transfer things, which makes plenty of sense. (In retrospect I could definitely have gone further in that direction than the current timer-cli-copy-paste-y-ness, so apologies @sirosen for asking for a review involving much more extensive feedback than was probably necessary.) I'll go ahead with those changes, and copy as much as I can from what's already in Transfer. It should be mostly smooth but I might come back with questions where anything actually diverges.

src/globus_cli/commands/timer/create.py
sirosen: > The only thing is that the timer interface is unable to care about more than the `recursive` setting. Do you think it would make sense to still allow writing just a boolean in the file instead of the flag?

I'm not sure I understand fully. Let's suppose I have a TransferData built with three items:
```
tdata = TransferData(...)
tdata.add_item("foo", "foo", recursive=True)
tdata.add_item("bar", "bar", recursive=True)
tdata.add_item("baz", "baz", recursive=False)  # default, but explicit for this example
```

Does the Timer service not support this mode of use? That seems unlikely to me based on the CSV reader, but maybe I'm mistaken.

If your concern is about `--recursive` vs `--batch` and how those behave, there's mutex behavior in `globus transfer` for `--recursive` and `--batch` with a specialized error message:
https://github.com/globus/globus-cli/blob/b16ef4c18fcfbfa1c76136209ba76b3223523480/src/globus_cli/commands/transfer.py#L321-L326

I think the mutex_option_group decorator could support this in the future, if we give it the ability to customize the error message used.

src/globus_cli/commands/timer/update.py
sirosen: In that case, I think we want to update to `if stop_after_runs is not None`?
It's a "weird" behavior, but maybe someone will have a use or need for it.

rudyardrichter: I really can't imagine a reason we _should_ support it but this is also fine.

src/globus_cli/commands/timer/create.py
rudyardrichter: No it does, I'm trying to say that (due to the AP interface) Timer at most can use those three values for each line (path, path, recursive)—any other flag should apply to all of them. So `--recursive` is the only flag that should be allowed in the file. I guess it makes sense, I just find the csv-ish interface more intuitive than putting the flags in the file.

sirosen: I don't mean to defend the "inner command" method of parsing as perfect. It's odd at best.

My worry with CSV data is that it won't be easy to use when there are many options, mostly used infrequently. We may one day see data like
```
foo,bar
bar,baz,,,auto,,true,,md5
```
because it's a format that may accept the optionality of the rightmost columns, but doesn't have a per-row way of mapping keys to values.

Obviously, that example is extreme, but I like situations in which order doesn't matter. `--recursive a b` and `a b --recursive` are the same.

rudyardrichter: I agree with your example, and I'm not saying we should look to apply this in general. Specifically for the case with timer where you have exactly 3 values I would lean slightly towards continuing to support the CSV mode there. Do you feel like if the CSV behavior exists for Timer but not elsewhere that it would be prohibitively weird?

sirosen: Let me separate this into two separate points, because I think your last message is very-much on-target.

> Specifically for the case with timer where you have exactly 3 values I would lean slightly towards continuing to support the CSV mode there.

Yes, and I agree that with exactly 3 known columns, CSV is a nice format. However...

> Do you feel like if the CSV behavior exists for Timer but not elsewhere that it would be prohibitively weird?

Maybe sub "weird" for "creates surprising results", but yes. 😄 

Let's suppose that `globus transfer $src $dest --batch foo.txt` works today.
I'd like for `globus timer create transfer $src $dest --batch foo.txt --interval 10m ...` to also work, and vice versa. [1]

That means that because `globus transfer --batch` has its existing line-parsing behavior, we should support it in `globus timer create transfer`.

The "out" for this, if we much prefer the CSV usage, is to support both. And if we're going to do the work for that, I think we need `globus transfer` to support the same. If we implement both parsing styles, we need an option or usage to set the behavior. We could have `--batchmode [auto|txt|csv]` where `auto` has fancy behavior.

But all in all, this seems like it's going down a path of a new and separate feature to support CSV as a format for transfer batches. I'm not sure I'm convinced that the (IMO, mild) niceness of CSV for the simple case is worth it vs the cost of supporting it in both commands?

----

[1] I'm aware that this could be messed up by `--external-checksum`, but IMO that's a reason to push adding that to the AP and making it work, not to change how the CLI tries to make these commands appear related.

rudyardrichter: > I'd like for globus timer create transfer $src $dest --batch foo.txt --interval 10m ... to also work

This is kind of what I'm getting at, that for this `foo.txt` is limited to the src, dst, recursive triplet. This would kind of be similar to [1], that the AP should just do it the, really, _better_ way. But as it is the batch behavior is distinct between the two. So I'm trying to get to what seems the least confusing given that present limitation.

src/globus_cli/commands/timer/update.py
sirosen: I agree that there's little to no reason for a user to do this.
However, I've also learned (usually the hard way) that deciding that something doesn't need to be possible is an open invitation for a ticket asking us to change things to make it possible.

Given that we don't have a reason for forbidding it, I'd tweak to `is not None`.

src/globus_cli/commands/timer/create.py
sirosen: What I would consider ideal is that -- and let's set aside the details of how the parsing happens -- both `globus transfer` and `globus timer create transfer` use the same parsing methodology and same parse function, potentially parametrized for their different cases.

I'd like to see the `globus transfer` case implemented as:
```python
if batch:
    tdata = magical_parse_batch_input(batch)
else: ...
```
and the `globus timer create transfer` case implemented as:
```python
if batch:
    tdata = magical_parse_batch_input(batch, external_checksum_supported=False)
else:
    ...
```

That way, we don't have two different overall parse behaviors, and it's possible to formulate a single input which is valid for both. Even if we have an unfortunate situation like `--external-checksum` in which some inputs are valid for one but not the other, it makes it possible for this to work some of the time.

In principle, I would be fine with `magical_parse_batch_input` doing CSV, shlexed line-reads, newline-delimited JSON, any other "reasonable" format, or even detection based on the first line of input (imagine trying different parses until one works). In practice, I don't know that we want to bother about having multiple ways of parsing those inputs today.

sirosen:
I've kept this open until it seemed like we had extracted as much value as possible from this work, but I think it's now time to close it out.

PR #662, opened 2022-07-13
Apply "future annotations" imports and rewrites to all relevant modules

sirosen:
End goal: use `from __future__ import annotations` and newer import syntaxes in most/all modules where relevant.

1. Add a script to find modules which import `Union`, `Optional`, `List`, `Dict` from typing and don't import `annotations` from `__future__`
2. Run that script and add `from __future__ import annotations` to flagged modules
3. Let pyupgrade/black converge
4. Use [`unimport`](https://github.com/hakancelikdev/unimport) to quickly purge old imports which are obsoleted
5. Fix the only regression introduced by these rewrites (an incorrect removal by `unimport`, reported upstream), make sure linting/testing passes
6. Remove the `check_future_annotations.py` script


sirosen:
I rebased to make sure that the workflows were freshened up, but intentionally did not apply the no-news label. I've just applied it, and you can now see that CI is green!

PR #667, opened 2022-08-03
ignore empty lists of session_required_identities

aaschaer:
Will do some manual testing to make sure this works as expected with the Transfer errors.

### DIFF
--- a/changelog.d/20220803_153655_aaschaer_session_hook_fix.md
+++ b/changelog.d/20220803_153655_aaschaer_session_hook_fix.md
@@ -1,0 +1,27 @@
+<!--
+A new scriv changelog fragment.
+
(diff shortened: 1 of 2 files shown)


aaschaer:
Live testing looks good, think this is ready to go

aaschaer:
However, auth does not accept Transfer's dict as is, so maybe this should just be considered a Transfer bug?

sirosen (approved):
I'm going to approve+merge.

> auth does not accept Transfer's dict as is

Can you let me know what you mean by this? The fix looks good to me -- it reverts the change in v3.0.0 -- so I'm just dotting Is and crossing Ts. An email or DM is fine if you don't want to try to explain in this PR.

PR #677, opened 2022-09-14
Bugfix to `transfer --dry-run` text output

sirosen:
When `external_checksum` is omitted, the SDK now omits it from the payload instead of filling with `None`. This causes a failure in the CLI text formatting code which expects that any string key it's given is present.

To fix:
- add a test which runs `--dry-run` with text formatted output
- switch from a string key to a lambda which calls `x.get(...)`

This effectively emulates the previous behavior, in which the value was `None` as well.


aaschaer (approved):
Looks good to me!

PR #678, opened 2022-09-16
Add `globus gcp create` commands for creating mapped and guest collections

sirosen:
This introduces `globus gcp create mapped` and `globus gcp create guest` to create a GCP Endpoint a.k.a. Mapped Collection or a Shared Endpoint on GCP a.k.a. Guest Collection, respectively.
It is the first in a long series of steps towards replacing `globus endpoint create` with more GCSv5-compatible commands.

These commands needed _some_ but not all of the options which were previously shared between `globus endpoint create|update`, which necessitated a bit of work on shared option refactoring in the parsing subpackage. Specifically, a new decorator has been introduced with usage as follows:
```python
@endpointish_create_and_update_params("create", "collection")
def gcp_create_guest(...): ...

@endpointish_create_and_update_params("create", "collection")
def gcp_create_mapped(...): ...

@endpointish_create_and_update_params("create")
def endpoint_create(...): ...

@endpointish_create_and_update_params("update")
def endpoint_update(...): ...
```

The first parameter to `endpointish_create_and_update_params` is checked with `typing.Literal["create", "update"]`, so a variety of potential typos are rendered impossible. (Read: `typing.Literal` makes it safer and easier to define string-based internal APIs.)
Because the Literal usage is type-checked with `assert_never()`, `typing_extensions` is added to requirements data to support python < 3.11.

The `globus endpoint create` command is now documented as deprecated, but no runtime warning or other change is made.
After checking with our PM team, I've confirmed some of the helptext and changelog details, so we're okay to ship this text in a release without worrying about the phrasing.

The command implementations themselves look very similar to two of the branching paths of `globus endpoint create`. For `create mapped`, a single endpoint creation call is performed, and for `create guest`, the host is checked for activation, and then the shared endpoint creation call is performed. In both cases, the text output fields are similar to `globus endpoint create` outputs: `Message, Collection ID` and `Message, Collection ID, Setup Key`.

One related subtle change is included here to `globus endpoint create`. As part of the parameter refactor, `endpoint_create_and_update_params` was split into two decorators, `endpoint_create_params` and `endpoint_update_params`. Since `globus endpoint create --no-default-directory` doesn't make sense as a usage, `--no-default-directory` is only applied in the `endpoint_update_params` (along with the associated mutex info).

New tests are modeled fairly closely on the existing `globus endpoint create` tests, and a test-case for `--no-default-directory` needed to switch from `create` to `update`.

### DIFF
--- a/src/globus_cli/commands/endpoint/create.py
+++ b/src/globus_cli/commands/endpoint/create.py
@@ -84,7 +84,10 @@ def endpoint_create(
     **kwargs: Any,
 ) -> None:
     """
-    Create a new endpoint.
+    Create a new endpoint. (deprecated)
+
+    This command is deprecated. Either `globus gcp create` or the Globus Connect Server
+    CLI should be used instead for most cases.
 
     Requires a display name and exactly one of --personal, --server, or --shared to make
     a Globus Connect Personal, Globus Connect Server, or Shared endpoint respectively.
--- a/src/globus_cli/commands/endpoint/create.py
+++ b/src/globus_cli/commands/endpoint/create.py
@@ -6,15 +6,13 @@ from globus_cli.login_manager import LoginManager
 from globus_cli.parsing import (
     ENDPOINT_PLUS_REQPATH,
     command,
+    endpointish_create_and_update_params,
     mutex_option_group,
     one_use_option,
 )
 from globus_cli.termio import FORMAT_TEXT_RECORD, formatted_print
 
-from ._common import (
-    endpoint_create_and_update_params,
-    validate_endpoint_create_and_update_params,
-)
+from ._common import endpoint_create_params, validate_endpoint_create_and_update_params
 
 COMMON_FIELDS = [("Message", "message"), ("Endpoint ID", "id")]
 
@@ -47,7 +45,8 @@ $ globus endpoint create --shared host_ep:~/ my_shared_endpoint
 ----
 """,
 )
-@endpoint_create_and_update_params(create=True)
+@endpointish_create_and_update_params("create")
+@endpoint_create_params
 @one_use_option(
     "--personal",
     is_flag=True,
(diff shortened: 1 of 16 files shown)


src/globus_cli/commands/endpoint/create.py
```diff
@@ -84,7 +83,10 @@ def endpoint_create(
     **kwargs: Any,
 ) -> None:
     """
-    Create a new endpoint.
+    Create a new endpoint. (deprecated)
+
+    This command is deprecated. Either `globus gcp create` or the Globus Connect Server
+    CLI should be used instead for most cases.
```
aaschaer, line 89: Do we also want to add a deprecation warning to the command's output, or would that be too noisy?

src/globus_cli/parsing/shared_options/endpoint_create_and_update.py
```diff
@@ -0,0 +1,75 @@
...
+    f: C, name: str, *, display_name: Literal["argument", "option", "null"] = "null"
+) -> C:
+    if display_name == "argument":
+        f = click.argument("DISPLAY_NAME")(f)
+    elif display_name == "option":
+        f = click.option("--display-name", help=f"Name for the {name}")(f)
+    elif display_name == "null":
+        pass
+    else:
+        assert_never()
```
aaschaer, line 30: first time seeing assert_never for type checking impossible conditions like this, I like it

aaschaer (approved):
Looks good to me, left one comment asking if we might want to more loudly deprecate the old command, but this would be fine as is if not

src/globus_cli/commands/endpoint/create.py
sirosen: Not _yet_, but we will in the future. I discussed this with Brigitte, and she agreed with my decision to use the current (subtle) doc note.

src/globus_cli/parsing/shared_options/endpoint_create_and_update.py
sirosen: It's very good for enums, Literals, and the new match-case statements. Expect to see more of it from me in the future. :-)

PR #683, opened 2022-10-11
Update to mypy==0.982 and add update script

sirosen:
1. Copy the SDK version bump script and use it to write a mypy update script which tweaks tox.ini and .pre-commit-config.yaml
2. Apply the script update (latest mypy version is now in use)
3. Adjust one annotation to pass under the new mypy

Regarding (3), the failure stems from an annotation being more specific than is really needed. An extra type variable is being used and bound as a parameter to another type variable -- probably this should be supported by mypy, but perhaps there's a reason that it should not. In any case, the typevar is not really needed since the entire value in question is being passed through a `cast` call after the usage site. Therefore, the wayward type variable can be safely replaced with Any.

Regarding (1), this may be an unnecessary script being added. However, it's a nice and low-key way of exploring the potential space for an auto-update process which ensures the mypy version stays consistent. Next step: should we hook this script in to generate automatic PRs which bump `mypy` in both locations?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -29,7 +29,7 @@ repos:
     - id: pyupgrade
       args: ["--py36-plus"]
 - repo: https://github.com/pre-commit/mirrors-mypy
-  rev: v0.971
+  rev: v0.982
   hooks:
     - id: mypy
       additional_dependencies:
--- a/scripts/update_mypy.py
+++ b/scripts/update_mypy.py
@@ -1,0 +1,77 @@
+#!/usr/bin/env python
+from __future__ import annotations
+
+import json
+import pathlib
+import re
+import urllib.request
+
+REPO_ROOT = pathlib.Path(__file__).parent.parent
+
+
+class Abort(RuntimeError):
+    pass
+
+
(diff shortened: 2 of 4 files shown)


derek-globus (approved):
On (3), I don't think it was because we were being too specific.
I think it because RT was being used as a generic context ("return the same type as this decorator is wrapping") but wasn't actually doing that since python couldn't handle the out-of-scope typevar -> typevar reference.

I'd bet that if you instead dropped C & replaced all references with explicit `Callable[..., RT]` type hints it'd be happy with the generic reference.

That being said, the type hinting here is probably more verbose than the actual value we're getting out of it so I'm happy with the change as is if mypy is.

kurtmckee (approved):

sirosen:
> On (3), I don't think it was because we were being too specific.

Yes, "too specific" is not a very precise phrasing of this. However, the previous annotation was definitely accurate, at least from a theoretical perspective. It's just not useful to even try to describe `RT` on this context, since it is not used in a meaningful way.

I believe that this bit of type hinting dates from when I was still learning how to describe decorators correctly. The updated content better matches what I currently consider the best approach.

PR #687, opened 2022-10-13
Add `globus api flows` command

sirosen:


kurtmckee (approved):
Awesome! I've tried to use this before and was hoping this would get added!

sirosen:
Well, for the very near term you can `pipx install` from `main`. We'll ship a release soon, I'm sure, to put this and the like into everyone's hands.

sirosen:
Also, I probably would have included it in today's release if I had remembered/realized it was missing.

PR #688, opened 2022-10-17
Timer create data_access handling part 1

sirosen:
Enable detection of endpoints which *will* require use of the data_access scope, and reject them when used with the `globus timer create transfer` command.

This checks for non-HA Mapped Collections and rejects them with a usage error.


sirosen:
> Do we want to disallow / warn for usage against HA collections?

I think that kind of behavior is out of scope for the CLI. If the Timer service gets updated such that it can be used with HA collections, the CLI becomes "wrong". That makes me firmly against disallowing usage and weakly against warning about usages.

PR #690, opened 2022-10-19
Convert to 'import typing as t' style

sirosen:
And apply flake8-typing-as-t

### DIFF
--- a/src/globus_cli/commands/group/invite/_common.py
+++ b/src/globus_cli/commands/group/invite/_common.py
@@ -1,27 +1,27 @@
 from __future__ import annotations
 
 import sys
+import typing as t
 import uuid
-from typing import TYPE_CHECKING, Callable
 
 import click
 
 from globus_cli.parsing import ParsedIdentity
 
-if TYPE_CHECKING:
+if t.TYPE_CHECKING:
     import globus_sdk
 
     from globus_cli.services.auth import CustomAuthClient
 
 if sys.version_info >= (3, 8):
-    from typing import Literal
+    from typing import Literal  # noqa: TYT03
 else:
     from typing_extensions import Literal
 
 
 def get_invite_formatter(
     case: Literal["accept", "decline"]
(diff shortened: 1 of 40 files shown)


src/globus_cli/commands/group/invite/_common.py
```diff
@@ -1,27 +1,27 @@
...
 
-if TYPE_CHECKING:
+if t.TYPE_CHECKING:
     import globus_sdk
 
     from globus_cli.services.auth import CustomAuthClient
 
 if sys.version_info >= (3, 8):
-    from typing import Literal
+    from typing import Literal  # noqa: TYT03
```
kurtmckee, line 17: I'm assuming that `TYT03` is specific to the `flake8-typing-as-t` checks. Would it be an option to add a note about what this is specifically addressing? I'm hoping that there's a concise descriptor of the error that's being disabled by this. For example,

```python
    from typing import Literal  # noqa: TYT03 # ignore 'from typing import ...'
```

kurtmckee (approved):
I think I saw about 4 places where "TYT03" was introduced. If there's a concise way to describe the error that `noqa: TYT03` is disabling, I would sure like to have that inline. :grinning:

src/globus_cli/commands/group/invite/_common.py
sirosen: I actually want to do more work on `flake8-typing-as-t` in my off hours to eliminate these comments.
That's the rule for `from typing import ...` usages, but if I can do the static analysis to see that we're in an `if sys.version_info ...` check or the else-branch thereof, it should be possible to make the plugin ignore these usages.

kurtmckee: Fair enough. You've got my approval on the PR either way! :+1:

sirosen: Lest I forget, I put it into an issue for myself
https://github.com/sirosen/flake8-typing-as-t/issues/1

PR #697, opened 2022-11-02
Add `globus flows delete` to delete flow by ID

sirosen:
In addition to the core change here, a new command, some of the parsing code is slightly rearranged. `endpoint_id_arg` and `collection_id_arg` helpers are moved into a dedicated module where they are joined by `flow_id_arg`. The goal of this adjustment is not to make major changes, but to better make room for `flow_id_arg`.

The test case relies on globus_sdk._testing data as well as an explicit mock.

### DIFF
--- a/src/globus_cli/commands/flows/delete.py
+++ b/src/globus_cli/commands/flows/delete.py
@@ -1,0 +1,35 @@
+from __future__ import annotations
+
+import uuid
+
+from globus_cli.login_manager import LoginManager
+from globus_cli.parsing import command, flow_id_arg
+from globus_cli.termio import Field, TextMode, display, formatters
+
+
+@command("delete", short_help="Delete a flow")
+@flow_id_arg
+@LoginManager.requires_login(LoginManager.FLOWS_RS)
+def delete_command(login_manager: LoginManager, flow_id: uuid.UUID):
+    """
+    Delete a flow
+    """
+    flows_client = login_manager.get_flows_client()
+
+    fields = [
+        Field("Deleted", "DELETED", formatter=formatters.Bool),
+        Field("Flow ID", "id"),
+        Field("Title", "title"),
+        Field(
+            "Owner",
+            "flow_owner",
+            formatter=formatters.auth.PrincipalURNFormatter(
+                login_manager.get_auth_client()
+            ),
+        ),
+        Field("Created At", "created_at", formatter=formatters.Date),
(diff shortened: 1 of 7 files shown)


src/globus_cli/commands/flows/delete.py
```diff
@@ -0,0 +1,35 @@
...
+
+    fields = [
+        Field("Deleted", "DELETED", formatter=formatters.Bool),
+        Field("Flow ID", "id"),
+        Field("Title", "title"),
+        Field(
+            "Owner",
+            "flow_owner",
+            formatter=formatters.auth.PrincipalURNFormatter(
+                login_manager.get_auth_client()
```
kurtmckee, line 27: The call to the login manager here doesn't have a side-effect (such as some network I/O) that the `flows_client` relies on, does it?

src/globus_cli/parsing/shared_options/id_args.py
```diff
@@ -0,0 +1,39 @@
+"""
+These are semi-standard arg names with overridable metavars.
+
+Basic usage:
+
+>>> @endpoint_id_arg
+>>> def command_func(endpoint_id: uuid.UUID):
+>>>     ...
```
kurtmckee, line 8: Is this pulled out by anything, like autodoc? It may be worthwhile to call this out as:

```rst
..  code-block:: pycon

    >>> @endpoint_id_arg
    ... def command_func(endpoint_id: uuid.UUID):
    ...     ...
```

If not, carry on! 😀

kurtmckee (approved):

src/globus_cli/commands/flows/delete.py
sirosen: No, it's just grabbing credentials out of the on-disk sqlite storage and using them to build a client object.

If you look at how the `login_manager` gets passed, you'll see that it's through a `requires_login` decorator -- that decorator's job is to ensure the user is properly credentialed. If the user doesn't have creds for the clients listed there, the decorator hard-fails the command with a message instructing them to login.

src/globus_cli/parsing/shared_options/id_args.py
sirosen: Yeah, it's not autodoc'ed. I was sort of tempted to remove it when I moved these docs around, but I figure it's not doing any harm and it might help someone get a quick read on these helpers. 😄

PR #699, opened 2022-11-03
Add support for timer create with data_access

sirosen:
EDIT: revised 2023-02-17 after a significant overhaul to use the consents API.

There is at least one noted significant issue identified during final review.

---

Subsequent to the changes below, `globus timer create transfer` is no longer marked hidden and it is included in the changelog. This therefore marks the "release" of `globus timer create transfer` as a command.

In order to support this change, the `timer` scope handling, which also touches on the main login codepath, had to be reworked. There are two reasons for this:

1. This effort uncovered that the scope requested was not sufficient for new transfer timers. The needed scope is `timer[transfer_ap[transfer]]`, but we only requested `timer` previously.

2. It needs to be possible the parametrize the most nested one of these scopes (the `transfer` scope) with dependencies. So a scope-building helper was introduced, `compute_timer_scope`

`compute_timer_scope` encodes the construction of two variant scope strings: `timer[transfer_ap[transfer]]` and
`timer[transfer_ap[transfer[foo/data_access bar/data_access]]]`. One with and one without some set of data_access scopes. The first variant is baked into the LoginManager as the statically required Timer scope.

Client Credentials handling is not included in the data_access handling code for the time being. Therefore, if client credentials are used to run `globus timer create transfer` against a collection with `data_access`, an explicit usage error is thrown for now. In the future, it should be possible to use client creds to request a new token which has the required dependency, but this is set aside as future work.

---

## ⚠️ The remaining "consent gap" in `requires_login` (resolved! ✔️ )

Original text preserved below, but this problem was solved.

`requires_login` cannot -- even with our recent changes to detect needed scopes -- detect changes in scope dependencies. Therefore, it is possible that a user will have logged in in the past, have a timer token which does not have the dependent scope chain, and will *not* be prompted to login again.
We could simply recommend that anyone who finds that timer creation fails do a logout and login, but I find that unsatisfactory as a workaround. I'd rather figure out how we want to track the addition of scope dependencies.

I'm going to work on this now, but it may be better to flip the `hidden=True` flag back to `False` so that this could merge and we're not announcing support for something unfinished, while still getting the rest of this body of work merged.

---

## Post-rebase

I've rebased and, in order to have any kind of sensible messaging in the git log, squashed everything down to one commit.
We kept pulling bits and pieces out of this, so the rebase stopped really making sense when I was looking through the commits.

I'm going to self-review the results here at least once, and I think this would benefit from a final round of testing as well.

### DIFF
--- a/changelog.d/20221102_223217_sirosen_timer_prompt.md
+++ b/changelog.d/20221102_223217_sirosen_timer_prompt.md
@@ -1,0 +1,9 @@
+### Enhancements
+
+* Add `globus timer create transfer` as a new command for creating new timers
+
+** The command prompts for login if data_access consents are detected as a
+   requirement
+
+* `globus session consent` now supports a `--timer-data-access` flag, specifically
+   to help support timer creation
--- a/src/globus_cli/commands/timer/create/__init__.py
+++ b/src/globus_cli/commands/timer/create/__init__.py
@@ -4,7 +4,6 @@ from globus_cli.parsing import group
 @group(
     "create",
     short_help="Submit a Timer job",
-    hidden=True,
     lazy_subcommands={"transfer": (".transfer", "transfer_command")},
 )
 def create_command() -> None:
--- a/src/globus_cli/commands/timer/create/transfer.py
+++ b/src/globus_cli/commands/timer/create/transfer.py
@@ -9,7 +9,11 @@ import click
 import globus_sdk
 
 from globus_cli.endpointish import Endpointish
-from globus_cli.login_manager import LoginManager
+from globus_cli.login_manager import (
+    LoginManager,
+    is_client_login,
+    read_well_known_config,
+)
 from globus_cli.parsing import (
     ENDPOINT_PLUS_OPTPATH,
     TimedeltaType,
@@ -32,11 +36,6 @@ if sys.version_info >= (3, 8):
     from typing import Literal
 else:
     from typing_extensions import Literal
-
-# XXX: this may need to be parametrized over data_access scopes?
-_TRANSFER_AP_SCOPE = (
-    "https://auth.globus.org/scopes/actions.globus.org/transfer/transfer"
-)
 
 
 INTERVAL_HELP = """\
@@ -96,7 +95,7 @@ def resolve_start_time(start: datetime.datetime | None) -> datetime.datetime:
     type=click.IntRange(min=1),
     help="Stop running the transfer after this number of runs have happened",
 )
-@LoginManager.requires_login("timer", "transfer")
+@LoginManager.requires_login("auth", "timer", "transfer")
 def transfer_command(
     *,
     login_manager: LoginManager,
@@ -130,6 +129,7 @@ def transfer_command(
     """
     from globus_cli.services.transfer import add_batch_to_transfer_data, autoactivate
 
+    auth_client = login_manager.get_auth_client()
     timer_client = login_manager.get_timer_client()
     transfer_client = login_manager.get_transfer_client()
 
@@ -162,9 +162,13 @@ def transfer_command(
 
     # Check endpoint activation, figure out scopes needed.
 
+    # the autoactivate helper may present output and exit in the case of v4 endpoints
+    # which need activation (e.g. OA4MP)
+    autoactivate(transfer_client, source_endpoint, if_expires_in=86400)
+    autoactivate(transfer_client, dest_endpoint, if_expires_in=86400)
+
     # check if either source or dest requires the data_access scope, and if so
-    # FIXME: hard fail for now (in the future, we should pick up on the requirement and
-    # generate a scope string to check against our logins)
+    # prompt the user to go through the requisite login flow
     source_epish = Endpointish(source_endpoint, transfer_client=transfer_client)
     dest_epish = Endpointish(dest_endpoint, transfer_client=transfer_client)
     needs_data_access: list[str] = []
@@ -172,20 +176,41 @@ def transfer_command(
         needs_data_access.append(str(source_endpoint))
     if dest_epish.requires_data_access_scope:
         needs_data_access.append(str(dest_endpoint))
+
+    # this list will only be populated *if* one of the two endpoints requires
+    # data_access, so if it's empty, we can skip any handling
     if needs_data_access:
-        raise click.UsageError(
-            "Unsupported operation. 'globus timer create transfer' does not currently "
-            "support collections which use the data_access scope: "
-            f"{','.join(needs_data_access)}"
-        )
+        # if the user is using client credentials, we cannot support the incremental
+        # auth step in the current implementation
+        #
+        # TODO: think through how we can use the client creds to request the
+        # requisite token in this case; it should be possible
+        if is_client_login():
+            raise click.UsageError(
+                "Unsupported operation. When using client credentials, "
+                "'globus timer create transfer' does not currently support "
+                "collections which use the data_access scope: "
+                f"{','.join(needs_data_access)}"
+            )
 
-    # Note this will provide help text on activating endpoints.
-    autoactivate(transfer_client, source_endpoint, if_expires_in=86400)
-    autoactivate(transfer_client, dest_endpoint, if_expires_in=86400)
+        request_data_access = _derive_needed_scopes(auth_client, needs_data_access)
 
-    # XXX: this section needs to be re-evaluated when 'timer create transfer' gets more
-    # capabilities to handle endpoints with scope requirements
-    # this would likely be the place to insert scope related checks
+        if request_data_access:
+            scope_request_opts = " ".join(
+                f"--timer-data-access '{target}'" for target in request_data_access
+            )
+            click.echo(
+                f"""\
+A collection you are trying to use in this timer requires you to grant consent
+for the Globus CLI to access it.
+
+Please run
+
+  globus session consent {scope_request_opts}
+
+to login with the required scopes"""
+            )
+            click.get_current_context().exit(4)
 
     transfer_data = globus_sdk.TransferData(
         source_endpoint=source_endpoint,
@@ -222,7 +247,98 @@ def transfer_command(
             name=name,
             stop_after=stop_after_date,
             stop_after_n=stop_after_runs,
-            scope=_TRANSFER_AP_SCOPE,
+            # the transfer AP scope string (without any dependencies)
+            scope="https://auth.globus.org/scopes/actions.globus.org/transfer/transfer",
         )
     )
     display(response, text_mode=TextMode.text_record, fields=JOB_FORMAT_FIELDS)
+
+
+def _derive_needed_scopes(
+    auth_client: globus_sdk.AuthClient,
+    needs_data_access: list[str],
+) -> list[str]:
+    from globus_sdk.scopes import GCSCollectionScopeBuilder
+
+    # read the identity ID stored from the login flow
+    # we will semi-gracefully handle the case of the data having been damaged/corrupted
+    user_data = read_well_known_config("auth_user_data")
+    if user_data is None:
+        raise RuntimeError(
+            "Identity ID was unexpectedly not visible in storage. "
+            "A new login should fix the issue. "
+            "Consider using `globus login --force`"
+        )
+    user_identity_id = user_data["sub"]
+
+    # get the user's Globus CLI consents
+    consents = auth_client.get(f"/v2/api/identities/{user_identity_id}/consents")[
+        "consents"
+    ]
+
+    # we need to now find the relevant consents which might match the data_access scopes
+    #
+    # this takes the form of a tree traversal
+    # the first parts of the tree should always be present, as they indicate the Timer
+    # consent which the CLI requests statically on login...
+
+    # find the top-level Timer consent
+    for consent in consents:
+        if (
+            consent["scope_name"] == globus_sdk.TimerClient.scopes.timer
+            and len(consent["dependency_path"]) == 1
+        ):
+            timer_consent = consent
+            break
+    else:
+        raise LookupError("could not find timer consent")
+
+    # find the Timer->TransferAP consent
+    first_order_dependencies = {
+        c["scope_name"]: c
+        for c in consents
+        if len(c["dependency_path"]) == 2
+        and timer_consent["id"] in c["dependency_path"]
+    }
+    timer2transferAP_consent = first_order_dependencies[
+        "https://auth.globus.org/scopes/actions.globus.org/transfer/transfer"
+    ]
+
+    # find the Timer->TransferAP->Transfer consent
+    second_order_dependencies = {
+        c["scope_name"]: c
+        for c in consents
+        if len(c["dependency_path"]) == 3
+        and timer_consent["id"] in c["dependency_path"]
+        and timer2transferAP_consent["id"] in c["dependency_path"]
+    }
+    timer2transferAP2transfer_consent = second_order_dependencies[
+        globus_sdk.TransferClient.scopes.all
+    ]
+
+    # find all of the Timer->TransferAP->Transfer->* consents
+    third_order_dependencies = {
+        c["scope_name"]: c
+        for c in consents
+        if len(c["dependency_path"]) == 4
+        and timer_consent["id"] in c["dependency_path"]
+        and timer2transferAP_consent["id"] in c["dependency_path"]
+        and timer2transferAP2transfer_consent["id"] in c["dependency_path"]
+    }
+
+    # in that last step, we reached the leaves of the tree
+    # (Okay, actually, that's a lie. We don't know what other values might exist
+    # further down in the tree. But luckily, it doesn't matter. We only care about the
+    # children of the node we've reached.)
+    # now we need to evaluate those leaves against our requirements
+
+    # check the 'needs_data_access' scope names against the 3rd-order dependencies
+    # of the Timer scope and record the names of the ones which we need to request
+    will_request_data_access: list[str] = []
+    for name in needs_data_access:
+        scope_name = GCSCollectionScopeBuilder(name).data_access
+        if scope_name not in third_order_dependencies:
+            will_request_data_access.append(name)
+
+    # return these ultimately filtered requirements
+    return will_request_data_access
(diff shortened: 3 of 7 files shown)


src/globus_cli/commands/timer/create/transfer.py
```diff
@@ -161,30 +167,47 @@ def transfer_command(
...
+    # which need activation (e.g. OA4MP)
     autoactivate(transfer_client, source_endpoint, if_expires_in=86400)
     autoactivate(transfer_client, dest_endpoint, if_expires_in=86400)
 
-    # XXX: this section needs to be re-evaluated when 'timer create transfer' gets more
-    # capabilities to handle endpoints with scope requirements
-    # this would likely be the place to insert scope related checks
+    # if it is a client login or `--skip-data-access-check` has been used, skip the
+    # following handling for data_access
+    if not skip_data_access_check:
```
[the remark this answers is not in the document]
sirosen: The comment is out of date with the last point that I reached with this work. I believe it predates the `is_client_login()` check being included below.

My original thinking was that we didn't need to check `data_access` when client credentials are used, but if I recall correctly we do need to check and potentially request a new token.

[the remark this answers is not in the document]
kurtmckee: The `is_client_login()` check was present when I reviewed this. Based on the comment written directly above this, I'm wondering whether the condition should be something like:

```python
# if it is a client login or `--skip-data-access-check` has been used, skip
if not skip_data_access_check and not is_client_login():
    # suggested by comment:   ^^^^^^^^^^^^^^^^^^^^^^^^^
    ...
```

sirosen:
I'm converting this back to a draft until we get the scope parsing piece and can see how it may improve the logic here.

src/globus_cli/commands/timer/create/transfer.py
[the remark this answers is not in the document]
sirosen: I need to double-check this, but I think the current logic is more right than the comment. But I'll circle back on this and really sort it out before saying more, since I don't remember what's here perfectly.

kurtmckee (approved):

tests/conftest.py
```diff
@@ -233,22 +247,31 @@ def func(line, assert_exit_code=0, stdin=None, matcher=False):
...
+Line:
+  {}
+exited with {} when expecting {}
+
+stdout:
+{}
+stderr:
+{}
+network calls recorded:
+{}"""
```
kurtmckee, line 261: Similar feedback for this: I recommend an f-string here.

kurtmckee (approved):
I left feedback, but I approve this! :+1:

src/globus_cli/commands/timer/create/transfer.py
```diff
@@ -258,3 +263,113 @@ def transfer_command(
...
+        raise RuntimeError(
+            "TimerClient.authorizer was not set when checking data_access"
+        )
+    if not timer_authorization.startsiwth("Bearer "):  # pragma: no cover
+        raise RuntimeError(
+            "TimerClient.authorizer was not using Bearer Token Authorization"
+        )
+
+    timer_token = timer_authorization[len("Bearer ") :]
+    """
```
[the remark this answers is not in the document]
sirosen: Yes, I was hoping that there would be more to be learned about the consents API such that we could avoid the view_consents scope and possibly use this token. It turns out, we can use the API without that scope _and_ without this token either, which surprised me. This was removed as part of my recent final prep to make this ready to go.

tests/conftest.py
sirosen: This predated the independent PR for this change, where we put together revisions. So if I did my rebase correctly, this change should disappear from the history.

src/globus_cli/commands/logout.py
```diff
@@ -142,7 +142,7 @@ def logout_command(*, login_manager: LoginManager, ignore_errors):
 
         adapter.remove_tokens_for_resource_server(rs)
 
-    adapter.remove_config(_STORE_CONFIG_USERINFO)
+    remove_well_known_config("auth_user_data")
```
sirosen, line 145: `remove_well_known_config` is basically the same as the above line, but it relies on `Literal[...]` type-safety to ensure we always target the correct strings.

Therefore, when I made the change which introduced it, and the other helpers like it, I made sure to do a quick pass over the annotations for the modules where they were used to be certain that `mypy` was type-checking there. That's why each place you see one of these `*_well_known_config` helpers used, you will likely see other changes to annotations.

sirosen:
This is finally ready for review and should be good to merge shortly. I'm doing a self-review pass because I spotted a couple of minor errors already. But the command works and is very close to what I think we should be shipping for this.

changelog.d/20221102_223217_sirosen_timer_prompt.md
```diff
@@ -0,0 +1,9 @@
+### Enhancements
+
+* Add `globus timer create transfer` as a new command for creating new timers
+
+** The command prompts for login if data_access consents are detected as a
+   requirement
+
+* `globus login` will now request a `view_consents` scope and new logins may
+  be required in order to run some commands
```
sirosen, line 9:
```suggestion

* `globus session consent` now supports a `--timer-data-access` flag, specifically
   to help support timer creation
```

I removed the new scope (not actually needed), and added a flag. I think this reflects the changes pretty well.

src/globus_cli/commands/timer/create/__init__.py
```diff
@@ -4,7 +4,6 @@
 @group(
     "create",
     short_help="Submit a Timer job",
-    hidden=True,
```
sirosen, line 7: This is a good place to note or remind people: the code here existed and was even in existing CLI releases, but was marked `hidden=True`, meaning it wasn't published in docs or visible in helptext. It was latent, but not officially released or supported.

So this change marks our last chance for highly backwards incompatible changes to the command before we consider it subject to our support guarantees.

src/globus_cli/commands/timer/create/transfer.py
```diff
@@ -9,13 +9,18 @@
...
+    is_client_login,
+    read_well_known_config,
+)
 from globus_cli.parsing import (
     ENDPOINT_PLUS_OPTPATH,
     TimedeltaType,
     command,
     encrypt_data_option,
     fail_on_quota_errors_option,
+    no_local_server_option,
```
sirosen, line 23:
```suggestion
```

Need to remove this and the usage. It was in an earlier version of the command but is not what we want to be doing.

src/globus_cli/commands/timer/create/transfer.py
```diff
@@ -96,7 +96,17 @@ def resolve_start_time(start: datetime.datetime | None) -> datetime.datetime:
...
 )
-@LoginManager.requires_login(LoginManager.TIMER_RS, LoginManager.TRANSFER_RS)
+@click.option(
+    "--skip-data-access-check",
+    is_flag=True,
+    help="Skip the check for endpoints requiring the data_access scope. "
+    "This flag also avoids prompting for immediate login when endpoints require "
+    "data_access.",
+)
+@no_local_server_option
```
sirosen, line 106:
```suggestion
```

Not only should we remove `no_local_server_option` (per above), but this skip-data-access-check idea was a workaround from an earlier draft in which we couldn't get at the consents API.

Although it might turn out to be useful to someone in the future to create a timer without checking, I think that feature should have its own independent justification. So I'll be removing it here, for now.

src/globus_cli/commands/timer/create/transfer.py
```diff
@@ -117,6 +127,8 @@ def transfer_command(
     skip_source_errors: bool,
     fail_on_quota_errors: bool,
     notify: dict[str, bool],
+    skip_data_access_check: bool,
+    no_local_server: bool,
```
sirosen, line 131:
```suggestion
```

src/globus_cli/login_manager/manager.py
```diff
@@ -66,6 +69,16 @@ class LoginManager:
...
+    # a second map is needed to track the duality of
+    #   server -> top level scope (for checking)
+    # vs
+    #   server -> top level scope with dependencies (for requesting)
+    STATIC_SCOPES_WITH_DEPENDENCIES: dict[str, tuple[list[str], list[str]]] = {
+        TIMER_RS: (
+            [TimerScopes.timer],
+            [_TIMER_SCOPE_WITH_DEPENDENCIES],
+        ),
+    }
```
sirosen, line 81: There are lots of other near-equivalent ways of phrasing this. Amongst those, I have basically no preference, as long as we allow ourselves to keep the existing scopes as-written and this is only a tweak for the timer scope declaration.

The main thing is that the login manager will need both the top level scope string and the scope string with dependencies to use in login flows.

So if you read this and think "that's a weird hack", I invite alternatives as long as they are simple to implement and fit well into the usages which required this workaround.

src/globus_cli/login_manager/tokenstore.py
```diff
@@ -160,16 +163,42 @@ def internal_auth_client():
...
+    return adapter.read_config(name)
+
+
+def remove_well_known_config(
+    name: Literal["auth_client_data", "auth_user_data"],
+    *,
+    adapter: SQLiteAdapter | None = None,
+) -> None:
+    adapter = adapter or token_storage_adapter()
+    adapter.remove_config(name)
```
sirosen, line 204: Prior to these helpers which abstract away the storage adapter (which is a singleton), there were places in the code, including tests, which pulled in constants from this and one other module in order to access the config data.

On one level, that's fine, but it seemed like the need to read the user's identity ID in `globus timer create transfer` demanded a nicer interface, since the access was no longer isolated to "special" places like the `globus login` implementation.

tests/conftest.py
```diff
@@ -134,6 +140,12 @@ def test_token_storage(mock_login_token_response):
         "auth_client_data",
         {"client_id": "fakeClientIDString", "client_secret": "fakeClientSecret"},
     )
+    # NB: this carefully matches the ID provided by our "foo_user_info" data fixture
+    # in the future, this should be moved to a dedicated fixture providing the current
+    # user's identity ID
+    mockstore.store_config(
+        "auth_user_data", {"sub": "25de0aed-aa83-4600-a1be-a62a910af116"}
+    )
```
sirosen, line 148: There's a broader refactor I'd like to do in order to fix this relationship with the fixture file data, but I'm intentionally putting that out of scope as this work already has so many touches on central parts of the application.

tests/functional/timer/test_transfer_create.py
```diff
@@ -278,24 +328,105 @@ def fake_resolve(dt):
...
+            appears_in_consent_prompt.append(dst)
+    elif has_matching_consent == "source":
+        if data_access_position in ("destination", "both"):
+            appears_in_consent_prompt.append(dst)
+    elif has_matching_consent == "destination":
+        if data_access_position in ("source", "both"):
+            appears_in_consent_prompt.append(src)
+    else:
+        raise NotImplementedError
+    expect_consent_error = bool(appears_in_consent_prompt)
```
sirosen, line 383: The giant block above didn't feel great to me, since we're taking a matrix over two parameters and then using that to "build" the test behaviors dynamically, in the test.
If someone sees a good way through this, but while still testing that full matrix, I'm open to ideas.

I did consider building the matrix and these behavioral parameters outside of the test, and then feeding in the whole generated "suite" as a set of test parameters. But it ended up looking basically the same as this, only in module scope rather than inside of test function scope. So I thought it came out worse.

src/globus_cli/login_manager/manager.py
sirosen: I actually rewrote this to use `MutableScope`, which I realized would work out well for this purpose after starting on the sole remaining outstanding issue, mentioned above.
So I'm marking this as resolved since I think the latest version of this is fine and will be improved further by https://github.com/globus/globus-sdk-python/pull/687

sirosen:
Small status update on this: expect one final change or set of changes here.
I've converted `LoginManager.STATIC_SCOPES` to contain a `MutableScope` object encoding the Timer dependency tree.

I'm going to make a change so that `has_login(TIMER_RESOURCE_SERVER)` will fail if the user hasn't consented to that full tree. I haven't yet decided if that will be done by checking against the consents API and failing if the tree is not present, or if I'll add a new slot to the config storage for tracking the last scope string we requested for things which are in STATIC_SCOPES. I'm a little concerned that this could get quite confusing, and that the consents API, while not a totally clean solve, might be simpler.

The goal is to force new logins for anyone running timer commands, let `globus login` pass it's early-exit check, generally put in place a strategy for keeping track of dynamic scope dependencies from `STATIC_SCOPES` as part of our base `LoginManager` state.

tests/functional/timer/test_transfer_create.py
kurtmckee: Wow, yeah, this is tough.

Your approach seems like the best for this situation, and for one reason: `NotImplementedError`. This clearly addresses the possibility that something was missed. I started to do the math on creating a test function for each possible configuration. While simpler, it would be tougher to guarantee that every possible combination was hit.

Kudos for this approach!

tests/unit/test_login_manager.py
```diff
@@ -228,3 +241,44 @@ def dummy_command(*, login_manager, collection_id):
...
+    )
+    assert computed.startswith(BASE_TIMER_SCOPE)
+    assert foo_scope in computed
+    assert bar_scope in computed
+    assert baz_scope in computed
+    start_part = f"{BASE_TIMER_SCOPE}[{TRANSFER_AP_SCOPE}[{transfer_scope}["
+    end_part = "]]]"
+    assert computed.startswith(start_part)
+    assert computed.endswith(end_part)
+    assert computed == f"{start_part}*{foo_scope} *{bar_scope} *{baz_scope}{end_part}"
```
kurtmckee, line 284: It seems like only the last line is needed.

kurtmckee (changes requested):

src/globus_cli/commands/timer/create/transfer.py
```diff
@@ -162,30 +164,55 @@ def transfer_command(
 
     # Check endpoint activation, figure out scopes needed.
 
+    # the autoactivate helper may present output and exit in the case of v4 endpoints
+    # which need activation (e.g. OA4MP)
```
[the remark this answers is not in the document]
sirosen: OA4MP is the acronym for "OAuth For MyProxy", which is the mainstream style of activation for v4 endpoints.
It requires that you open up the webapp and do an OAuth 1.0 flow with the endpoint's listed oauth server.

Not because it's exactly what you asked, but I think the full suite of v4 activation styles we see is:
- delegate-proxy (create a proxy X.509 cred by signing a user cert; here be dragons)
- oa4mp / "oauth" (the above, generating a user cert on the server which is handed back to our services)
- myproxy (send username+password to Transfer, have it exchange those for an X.509 user cert directly)
- autoactivation (almost a no-op, the X.509 user cert is owned by and created by Transfer)

src/globus_cli/commands/timer/create/transfer.py
```diff
@@ -162,30 +164,55 @@ def transfer_command(
...
     if needs_data_access:
-        raise click.UsageError(
-            "Unsupported operation. 'globus timer create transfer' does not currently "
-            "support collections which use the data_access scope: "
-            f"{','.join(needs_data_access)}"
-        )
+        # if the user is using client credentials, we cannot support the incremental
+        # auth step in the current implementation
+        #
+        # TODO: think through how we can use the client creds to request the
```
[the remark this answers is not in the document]
sirosen: 🤷 

I think this is well within our reach, but I don't want to try to solve this problem at the same time as the human-user case.

src/globus_cli/commands/timer/create/transfer.py
```diff
@@ -222,7 +249,98 @@ def transfer_command(
             name=name,
             stop_after=stop_after_date,
             stop_after_n=stop_after_runs,
-            scope=_TRANSFER_AP_SCOPE,
+            # the transfer AP scope string (without any dependencies)
+            scope="https://auth.globus.org/scopes/actions.globus.org/transfer/transfer",
```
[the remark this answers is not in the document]
sirosen: _I sure hope so._ 😂 

It does appear to be today.

tests/unit/test_login_manager.py
sirosen: _scratches head_

Yeah, you'd have to ask Past Stephen why it wasn't just that one. And he's not around anymore.

src/globus_cli/commands/timer/create/transfer.py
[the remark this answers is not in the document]
kurtmckee: Ah, okay. I misunderstood and thought that OA4MP was a specific example of a v4 endpoint!

kurtmckee (approved):

src/globus_cli/login_manager/__init__.py
```diff
@@ -15,6 +15,7 @@
 
 __all__ = [
     "MissingLoginError",
+    "compute_timer_scope",
```
sirosen, line 18:
```suggestion
```

This is an artifact of the rebase which I missed. It's included at the bottom of the list.

tests/conftest.py
```diff
@@ -80,8 +80,14 @@ def mock_login_token_response():
...
-            "openid profile email "
-            "urn:globus:auth:scope:auth.globus.org:view_identity_set",
+            " ".join(
+                [
+                    "openid",
+                    "profile",
+                    "email",
+                    "urn:globus:auth:scope:auth.globus.org:view_identity_set",
+                ]
+            ),
```
sirosen, line 90: Technically an unrelated and non-semantic change, I decided to leave this in because it reads better this way.

It made sense to be touching these lines when we added (and then removed) the `view_consents` scope.

tests/functional/timer/test_transfer_create.py
sirosen: Coming back to this, I don't have any new ideas for a better way to write it, and with one person approving of the approach taken, I'm considering this resolved.

It still doesn't "feel" resolved to me, but I think this is the sort of thing which we can live with until/unless it really _needs_ to be rewritten.

PR #700, opened 2022-11-10
Support Python 3.11

kurtmckee:
### Other

* Support Python 3.11.
* List scriv and tox as development dependencies.


sirosen (approved):
I vaguely object to listing `scriv` and `tox` in the extra which gets installed into the `tox` envs.
I plan to fix this in a follow-up -- the goodness of this change outweighs that objection, for me.

PR #701, opened 2022-11-10
Separate test and dev deps, remove pytest-cov

sirosen:
Introducing a 'test' extra to make the two sets of dependencies independently installable.

Also changed
- make .coveragerc more uniform with globus-sdk .coveragerc
- remove mention of dev deps from the changelog (which is destined for user-facing documentation)

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,5 +1,6 @@
 [run]
 parallel = True
+include = src/*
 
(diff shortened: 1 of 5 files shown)


kurtmckee (approved):
I like these improvements. Thanks for this!

PR #703, opened 2022-11-14
Add support for starting a flow

kurtmckee:
### Enhancements

* Add `globus flow start` as a new command for starting a flow.
* Add `globus login --flow <flow_id>` to trigger a login that asks for consent to start a flow.

### DIFF
--- a/src/globus_cli/commands/flows/start.py
+++ b/src/globus_cli/commands/flows/start.py
@@ -1,0 +1,149 @@
+from __future__ import annotations
+
+import typing as t
+import uuid
+
+import click
+
+from globus_cli.login_manager import LoginManager
+from globus_cli.parsing import JSONStringOrFile, command, flow_id_arg
+from globus_cli.termio import Field, TextMode, display, formatters
+
+ROLE_TYPES = ("flow_viewer", "flow_starter", "flow_administrator", "flow_owner")
+
+
+@command("start", short_help="Start a flow")
+@flow_id_arg
+@click.option(
+    "--input",
+    "input_document",
+    type=JSONStringOrFile(),
+    help="""
+        The JSON input parameters used to start the flow.
+
+        The input document may be specified inline,
+        or it may be a path to a JSON file, prefixed with "file:".
+
+        Example: Inline JSON:
+
+        \b
+            --input '{"src": "~/source"}'
+
+        Example: Path to JSON file:
+
+        \b
+            --input file:parameters.json
+
+        If unspecified, the default is an empty JSON object ('{}').
+    """,
+)
+@click.option(
+    "--label",
+    type=str,
+    help="A label to give the run.",
+)
+@click.option(
+    "--manager",
+    "managers",
+    type=str,
+    multiple=True,
+    help="""
+        A principal that may manage the execution of the run.
+
+        This option can be specified multiple times
+        to create a list of run managers.
+    """,
+)
+@click.option(
+    "--monitor",
+    "monitors",
+    type=str,
+    multiple=True,
+    help="""
+        A principal that may monitor the execution of the run.
+
+        This option can be specified multiple time
+        to create a list of run monitors.
+    """,
+)
+@click.option(
+    "--tag",
+    "tags",
+    type=str,
+    multiple=True,
+    help="""
+        A tag to associate with the run.
+
+        This option can be used multiple times
+        to create a list of tags.
+    """,
+)
+@LoginManager.requires_login(LoginManager.FLOWS_RS)
+def start_command(
+    login_manager: LoginManager,
+    flow_id: uuid.UUID,
+    input_document: dict | None | t.Any,
+    label: str | None,
+    managers: tuple[str],
+    monitors: tuple[str],
+    tags: tuple[str],
+):
+    """
+    Start a flow.
+
+    FLOW_ID must be a UUID.
+    """
+
+    if input_document is None:
+        input_document = {}
+
+    flow_client = login_manager.get_specific_flow_client(flow_id)
+    response = flow_client.run_flow(
+        body=input_document,
+        label=label,
+        tags=list(tags),
+        run_managers=list(managers),
+        run_monitors=list(monitors),
+    )
+
+    auth_client = login_manager.get_auth_client()
+    principal_formatter = formatters.auth.PrincipalURNFormatter(auth_client)
+    for principal_set_name in ("run_managers", "run_monitors"):
+        for value in response.get(principal_set_name, ()):
+            principal_formatter.add_item(value)
+    principal_formatter.add_item(response.get("run_owner"))
+
+    fields = [
+        Field("Flow ID", "flow_id"),
+        Field("Flow title", "flow_title"),
+        Field("Run ID", "run_id"),
+        Field("Run label", "label"),
+        Field(
+            "Run owner",
+            "run_owner",
+            formatter=principal_formatter,
+        ),
+        Field(
+            "Run managers",
+            "run_managers",
+            formatter=formatters.ArrayFormatter(
+                delimiter=", ",
+                element_formatter=principal_formatter,
+            ),
+        ),
+        Field(
+            "Run monitors",
+            "run_monitors",
+            formatter=formatters.ArrayFormatter(
+                delimiter=", ",
+                element_formatter=principal_formatter,
+            ),
+        ),
+        Field(
+            "Run tags",
+            "tags",
+            formatter=formatters.ArrayFormatter(delimiter=", "),
+        ),
+    ]
+
+    display(response, fields=fields, text_mode=TextMode.text_record)
--- a/src/globus_cli/login_manager/manager.py
+++ b/src/globus_cli/login_manager/manager.py
@@ -166,14 +166,16 @@ class LoginManager:
         if epilog is not None:
             click.echo(epilog)
 
-    def assert_logins(self, *resource_servers, assume_gcs=False):
+    def assert_logins(self, *resource_servers, assume_gcs=False, assume_flow=False):
         # determine the set of resource servers missing logins
         missing_servers = {s for s in resource_servers if not self.has_login(s)}
 
         # if we are missing logins, assemble error text
         # text is slightly different for 1, 2, or 3+ missing servers
         if missing_servers:
-            raise MissingLoginError(missing_servers, assume_gcs=assume_gcs)
+            raise MissingLoginError(
+                missing_servers, assume_gcs=assume_gcs, assume_flow=assume_flow
+            )
 
     @classmethod
     def requires_login(cls, *resource_servers: str):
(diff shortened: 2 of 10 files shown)


derek-globus (changes requested):
A couple minor option-structuring requests.

With this change now, there's 3 places now across the sdk & cli that we are defining this scope formatting logic.

Probably makes sense to pull that logic up into a common place soon. What do you think?


```python
rs_name = str(flow_id)
scope_name = f'flow_{str(flow_id).replace("-", "_")}_user'
scopes = [f"https://auth.globus.org/scopes/{flow_id}/{scope_name}"]

```

src/globus_cli/commands/flows/start.py
```diff
@@ -0,0 +1,145 @@
...
+
+ROLE_TYPES = ("flow_viewer", "flow_starter", "flow_administrator", "flow_owner")
+
+
+@command("start", short_help="Start a flow")
+@click.argument(
+    "flow_id",
+    type=uuid.UUID,
+    metavar="FLOW_ID",
+)
```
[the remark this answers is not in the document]
kurtmckee: Ooh, nice. Thanks!

src/globus_cli/commands/flows/start.py
```diff
@@ -0,0 +1,145 @@
...
+            --input file:parameters.json
+
+        If unspecified, the default is an empty JSON object ('{}').
+    """,
+)
+@click.option(
+    "--label",
+    type=str,
+    help="A label to give the run.",
+)
```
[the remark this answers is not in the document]
kurtmckee: The Flows service [doesn't require a label for the run](https://github.com/globusonline/flows/blob/main/globus/flows/views.py#L696-L703), so I didn't enforce anything here.

kurtmckee:
The scope-formatting is indeed a problem for me. However, the bug fix that resolves the hyphens to underscores in the scope string isn't currently available in a published SDK release so I couldn't re-use the scope builder while writing this code.

src/globus_cli/commands/flows/start.py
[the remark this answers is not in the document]
kurtmckee: Switched to this. Marking as resolved.

src/globus_cli/login_manager/manager.py
```diff
@@ -313,6 +315,30 @@ def _get_gcs_info(
...
+            scopes=[f"https://auth.globus.org/scopes/{flow_id}/{scope_name}"],
+        )
+        self.assert_logins(str(flow_id), assume_flow=True)
+
+        authorizer = self._get_client_authorizer(
+            str(flow_id),
+            no_tokens_msg=(
+                f"Could not get login data for flow {flow_id}. "
+                f"Try login with '--flow {flow_id}' to fix."
+            ),
```
sirosen, line 334: I'm just noticing a couple of things, which I had somehow missed when we spoke about this.
1. The `assert_logins` call only passes the flow ID, so it's pretty much guaranteed to print the correct message on failure (somehow when we spoke, I got mixed up and thought we'd be passing a larger set of values to `assert_logins`). I wanted to make sure to mention that I may have made misleading statements in the past.
2. I'm not sure how the `no_tokens_msg` value is reachable

`has_login` insists that the value be present and non-null, so I don't think it's possible for this message to print? But it's not easily statically analyzable as unreachable code. There might be some worthwhile cleanup we could apply here.

src/globus_cli/commands/flows/start.py
```diff
@@ -0,0 +1,141 @@
...
+        Field("Flow ID", "flow_id"),
+        Field("Flow title", "flow_title"),
+        Field("Run ID", "run_id"),
+        Field("Label", "label"),
+        Field(
+            "Run owner",
+            "run_owner",
+            formatter=formatters.auth.PrincipalURNFormatter(
+                login_manager.get_auth_client()
+            ),
```
sirosen, line 122: A somewhat unnecessary nitpick, but I believe you can...
```suggestion
            formatter=principal_formatter
```

src/globus_cli/commands/flows/start.py
```diff
@@ -0,0 +1,141 @@
...
+    response = flow_client.run_flow(
+        body=input_document,
+        label=label,
+        tags=list(tags),
+        run_managers=list(managers),
+        run_monitors=list(monitors),
+    )
+
+    auth_client = login_manager.get_auth_client()
+    principal_formatter = formatters.auth.PrincipalURNFormatter(auth_client)
```
sirosen, line 110: There is an optional performance improvement we can make here, but one which I'd like us to include.

What's needed is to iterate over all the values from `[response["run_owner"]] + response["run_managers"] + response["run_monitors"]` and pass them to `principal_formatter.add_item`.

To explain:

Each PrincipalFormatter gets a `globus_sdk.IdentityMap` which it contains. [PrincipalFormatter.resolved_ids](https://github.com/globus/globus-cli/blob/6fa67a6584ae181bf8f6801f6e587cf467c2b05a/src/globus_cli/termio/formatters/auth.py#L32).

[`add_item`](https://github.com/globus/globus-cli/blob/6fa67a6584ae181bf8f6801f6e587cf467c2b05a/src/globus_cli/termio/formatters/auth.py#L54) will attempt to parse those values and add them to that `IdentityMap`.

When `get()` or [`__getitem__`](https://github.com/globus/globus-sdk-python/blob/07d9b2cba29b7f8ef67ec27c24616c8232a7e382/src/globus_sdk/services/auth/identity_map.py#L244) is used on an `IdentityMap`, it attempts to lookup all of the identities which have been added in a single batch. (A single call to Auth.)

So by prepopulating the map, we get a single callout to Auth, rather than a separate call per element in each of these lists (with caching and cache hits).

sirosen (commented):
This is off to a good start, and coming together more smoothly than I expected. We have some rough edges to smooth out like the scope formatting, but overall I think we should be pretty happy with this.

src/globus_cli/commands/flows/start.py
```diff
@@ -0,0 +1,141 @@
...
+    "--manager",
+    "managers",
+    type=str,
+    multiple=True,
+    help="""
+        A principal that may manage the execution of the run.
+
+        This option can be specified multiple times
+        to create a list of run managers.
+    """,
```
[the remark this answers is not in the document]
kurtmckee: Re: `--manager 'kurt@globus.org'`

I tested exactly this syntax with the current code but forgot to follow up -- this appears to result in an error from the Flows service and may need to be reviewed at a service level.

src/globus_cli/commands/flows/start.py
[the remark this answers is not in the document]
derek-globus: I meant the code usage below made it look like the execution code would error out if `label` wasn't supplied. What happens if you start a flow with this command and don't supply a label?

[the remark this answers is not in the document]
kurtmckee: > What happens if you [...] don't supply a label?

The label will be set to `None`. When displaying the run, the web app will show "[Unlabeled Run 1234]", where "1234" is a large integer. When displaying the run label in a table of information, the web app will show an empty value.

src/globus_cli/commands/flows/start.py
kurtmckee: Ah, got it. I saw this pattern in a nearby file and followed it, but I see the pattern in `show.py` now (and your comment below) and I'm following that, now.

src/globus_cli/commands/flows/start.py
kurtmckee: Thanks for explaining this! I've followed this pattern based on what I see in the nearby `show.py`, now.

src/globus_cli/commands/flows/start.py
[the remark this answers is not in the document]
kurtmckee: I'll investigate making a generic ParamType for this.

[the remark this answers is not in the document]
kurtmckee: Story filed for this, but it won't be part of this effort.

src/globus_cli/commands/flows/start.py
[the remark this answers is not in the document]
kurtmckee: Thanks both for your input! I've updated the Python type annotation so that `label` is obviously nullable.

Note that the click type is still just `str`, as click expects a type that it can instantiate.

derek-globus (approved):

src/globus_cli/login_manager/manager.py
```diff
@@ -313,6 +315,30 @@ def _get_gcs_info(
...
 
+    def get_specific_flow_client(
+        self,
+        flow_id: uuid.UUID,
+    ) -> globus_sdk.SpecificFlowClient:
+        scope_name = f'flow_{str(flow_id).replace("-", "_")}_user'
+        self.add_requirement(
+            str(flow_id),
+            scopes=[f"https://auth.globus.org/scopes/{flow_id}/{scope_name}"],
+        )
```
sirosen, line 326: We should swap this out to the (recently corrected) SDK scope builder value.
I think it might even make sense to do
```python
client = SpecificFlowClient(flow_id)
self.add_requirement(str(flow_id), scopes=[client.scopes.user])
...
client.authorizer = ...
return client
```

The authorizer being created after-the-fact is a little unusual, but it doesn't violate any API/contract with the SDK.

tests/functional/test_login_command.py
```diff
@@ -83,5 +83,34 @@ def test_login_gcs_different_identity(
...
-    result = run_line("globus logout --yes")
+    run_line("globus logout --yes")
     assert test_token_storage.read_config(_STORE_CONFIG_USERINFO) is None
+
+
+def test_login_with_flow(monkeypatch, run_line):
+    """Verify that flow ID's are added as resource servers with correct scopes."""
+
+    uuid1 = str(uuid.uuid4())
+    scope_name = f'flow_{uuid1.replace("-", "_")}_user'
```
sirosen, line 94: Do you want to replace this scope-string building with the SDK method as well? I can see arguments either way, but my initial take is that we want to avoid having `.replace("-", "_")` anywhere in this repo. 😸

tests/functional/flows/test_start_flow.py
```diff
@@ -0,0 +1,122 @@
...
+    add_flow_login(flow_id)
+
+    # Configure identities.
+    owner_identity = {
+        "username": "yogi@jellystone.park",
+        "name": "Yogi",
+        "id": response.json["run_owner"].split(":")[-1],
+        "identity_provider": "c8abac57-560c-46c8-b386-f116ed8793d5",
+        "organization": "Hanna-Barbera",
+        "status": "used",
```
sirosen, line 25:
```suggestion
        "status": "smarter-than-your-av-er-age-bear",
```

tests/functional/flows/test_start_flow.py
```diff
@@ -0,0 +1,122 @@
...
+        identity["username"] for identity in run_manager_identities
+    )
+
+    monitors_match = re.search(
+        r"^Run monitors:\s+(?P<monitors>.+)$", result.output, flags=re.M
+    )
+    assert monitors_match is not None
+    assert monitors_match.group("monitors") == ", ".join(
+        identity["username"] for identity in run_monitor_identities
+    )
```
sirosen, line 122: 👍 for the way these matches are done on the output. I like this better than an `assert output == "foo: bar\nbar:baz\n"` style exact-match.

I wrote a helper for this at one point (https://github.com/globus/globus-cli/blob/ae17da9ce9d782a4f8d3a73fe1ed9a948a357264/tests/conftest.py#L156) which I would _not_ recommend  using. It would be good to think of some pattern which we want to encourage, similar to what you have here, and codify it in some better-authored utility class. That way, our tests will consistently check the fields in the text output.

src/globus_cli/login_manager/manager.py
kurtmckee: Argh, I missed this (and below). Thanks for catching these!

tests/functional/flows/test_start_flow.py
kurtmckee: I daren't change this lest it be used unexpectedly! :wink:

tests/functional/flows/test_start_flow.py
kurtmckee: Thank you! Yep, it may be beneficial to package output matching in a reusable way.

sirosen (approved):
Excellent! We can look forward to this being present in the next release.
I'm going to squash merge, which I think is consistent with some of our chat about this PR, and makes sense with the current commits here.

The need for `assert client.scopes is not None` is interesting. I wonder if we can (should?) annotate the SpecificFlowClient class with
```
class SpecificFlowClient(BaseClient):
    scopes: SpecificFlowScopeBuilderThingamjig
```

`scopes` is always non-null on an instance of the class. But as a class attribute it will be `None`... 🤔 

I'm not sure that the situation can be well explained to type checkers today.

PR #704, opened 2022-11-15
use entity_type to determine EndpointType

aaschaer:
SC Story: https://app.shortcut.com/globus/story/19989/use-entity-type-when-determining-endpointish-type

Tests are dependent on https://github.com/globus/globus-sdk-python/pull/650, so I'm making this a draft for now.

I'm debating renaming EndpointType to EntityType to match the field?

The `nice_name` changes probably have the most significant user facing effects, but I don't think they are likely to break anyone. I'm open to changing them as well if writing out the full names is too long.


src/globus_cli/endpointish/endpointish.py
```diff
@@ -33,7 +33,7 @@ def __init__(
         log.debug("Endpointish.data=%s", self.data)
 
         log.debug("Endpointish determine ep type")
-        self.ep_type = EndpointType.determine_endpoint_type(self.data)
+        self.ep_type = EndpointType(self.data.get("entity_type"))
```
sirosen, line 36: Do we need to wrap this in error handling at all?
I want the CLI to be robust to potential new entity_type values being introduced.

The previous logic always made _a_ determination, even though it could be incorrect if new cases were introduced. I think we may want a catch-all type for the CLI when entity_type is unrecognized. Even if we don't carefully define the behavior for `EntityType.UNDEFINED` (good name?), it would help us be able to reason about what will happen when a new entity_type is introduced.

src/globus_cli/endpointish/endpoint_type.py
```diff
@@ -1,77 +1,45 @@
...
-    SHARE = auto()
-    NON_GCSV5_ENDPOINT = auto()  # most likely GCSv4, but not necessarily
+    # endpoint / collection types, strings match entity_type value in
+    # transfer endpoint docs
+    GCP_MAPPED = "GCP_mapped_collection"
+    GCP_GUEST = "GCP_guest_collection"
+    GCSV5_ENDPOINT = "GCSv5_endpoint"
+    GCSV5_MAPPED = "GCSv5_mapped_collection"
+    GCSV5_GUEST = "GCSv5_guest_collection"
+    GCSv4_HOST = "GCSv4_host"  # most likely GCSv4, but not necessarily
```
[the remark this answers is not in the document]
aaschaer: I left it in because its still technically true, but if Transfer is calling it "GCSv4_host" maybe we don't care CLI side that it might not be?

src/globus_cli/endpointish/endpointish.py
aaschaer: Ah yeah, I'll make this able to handle a new entity_type.

Might not hurt to return `EntityType.UNDEFINED` if `entity_type` isn't present for some reason too?

src/globus_cli/endpointish/endpoint_type.py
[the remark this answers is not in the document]
kurtmckee: I wondered if this was still technically true but wanted to confirm. I trust your judgement to leave it in!

src/globus_cli/endpointish/endpointish.py
kurtmckee: QQ: Should there be an "unrecognized" and an "undefined" entity type?

kurtmckee (approved):

src/globus_cli/endpointish/endpointish.py
sirosen: If we're worried about the missing vs unrecognized ambiguity, I would go with `UNRECOGNIZED` and say that we don't yet recognize a missing `entity_type` as a meaningful/valid value. 😉

sirosen (approved):

PR #706, opened 2022-11-17
Update testing

kurtmckee:
* git-ignore `.coverage.*` files
* Consistently run `coverage combine`
* Set a minimum bar for code coverage (currently at 87%)
* Remove a coverage CLI option that is redundant due to a configuration in `.coveragerc`
* Always write an HTML coverage report
* Test the version-gathering code that intacts with PyPI
* Refactor code in the version-gathering function that cannot raise `RequestException`.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -5,6 +5,7 @@ include = src/*
 [report]
 show_missing = True
 skip_covered = True
+fail_under = 87
 
 exclude_lines =
     # the pragma to disable coverage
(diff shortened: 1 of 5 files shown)


sirosen (approved):
The only thing I was unsure about here was the `commands_pre` usage.
Kurt explained to me that it makes `tox` ignore the exit code (so that `coverage report` with the fail_under config could fail, but `coverage html` cannot cause that step to fail).

Since `coverage report` is now "test-like", that makes sense.

Potential future refinements include fixing the distutils usage (we have until py3.12), but ❤️ for the new tests.

PR #707, opened 2022-11-17
Better behavior when doing recursive filtering

aaschaer:
For https://app.shortcut.com/globus/story/17393/update-globus-ls-recursive-filter-x-to-use-new-transfer-filter-behavior-to-include-dirs

Depends on https://github.com/globus/globus-sdk-python/pull/652, so I'm making this a draft for now.

Also want to make sure we're all on the same page about expected behavior. Some example output:

```
$ globus ls $EP1:/ --recursive --filter ~*1.txt
home/
mnt/
not shareable/
share/
home/aaschaer/
home/aaschaer/file1.txt
not shareable/godata/
not shareable/godata/file1.txt
share/godata/
share/godata/file1.txt
```

### DIFF
--- a/changelog.d/20221117_103859_aaschaer_recursive_filter.md
+++ b/changelog.d/20221117_103859_aaschaer_recursive_filter.md
@@ -1,4 +1,16 @@
 ### Enhancements
 
-* the `--filter` and `--recursive` options to `globus ls` now combine better
-  with directories that don't match the name filter being explored
+* `globus ls` has improved behavior when the `--filter` and `--recursive` options
+   are used in combination
+
+  * directory names are not matched against the filter, allowing the operation to
+    traverse directories regardless of their names
+
+  * the `--filter` is still applied to filenames in all directories traversed by
+    the `ls` operation
+
+  * directory names can be filtered out of the text output by eliminating
+    lines which end in `/`
+
+  * the behaviors of `globus ls` commands with `--recursive` or `--filter`, but not
+    both, are unchanged
--- a/changelog.d/20221117_103859_aaschaer_recursive_filter.md
+++ b/changelog.d/20221117_103859_aaschaer_recursive_filter.md
@@ -1,0 +1,4 @@
+### Enhancements
+
+* the `--filter` and `--recursive` options to `globus ls` now combine better
+  with directories that don't match the name filter being explored
--- a/src/globus_cli/commands/ls.py
+++ b/src/globus_cli/commands/ls.py
@@ -180,29 +180,32 @@ def ls_command(
     ls_params: dict[str, t.Any] = {"show_hidden": int(show_hidden)}
     if path:
         ls_params["path"] = path
-    if filter_val:
-        # this char has special meaning in the LS API's filter clause
-        # can't be part of the pattern (but we don't support globbing across
-        # dir structures anyway)
-        if "/" in filter_val:
-            raise click.UsageError('--filter cannot contain "/"')
-        # format into a simple filter clause which operates on filenames
-        ls_params["filter"] = f"name:{filter_val}"
+
+    # this char has special meaning in the LS API's filter clause
+    # can't be part of the pattern (but we don't support globbing across
+    # dir structures anyway)
+    if filter_val and "/" in filter_val:
+        raise click.UsageError('--filter cannot contain "/"')
 
     # get the `ls` result
     if recursive:
-        # NOTE:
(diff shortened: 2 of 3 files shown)


changelog.d/20221117_103859_aaschaer_recursive_filter.md
```diff
@@ -0,0 +1,4 @@
+### Enhancements
+
+* the `--filter` and `--recursive` options to `globus ls` now combine better
```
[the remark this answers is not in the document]
aaschaer: Yeah, "combine better" isn't great, still trying to think of a good way to explain this behavior

changelog.d/20221117_103859_aaschaer_recursive_filter.md
```diff
@@ -0,0 +1,4 @@
...
+  with directories that don't match the name filter being explored
```
sirosen, line 4: May I suggest this longer, but hopefully more digestible, explanation?

```suggestion
* `globus ls` has improved behavior when the `--filter` and `--recursive` options
   are used in combination

  * directory names are not matched against the filter, allowing the operation to
    traverse directories regardless of their names

  * the `--filter` is still applied to filenames in all directories traversed by
    the `ls` operation

  * directory names can be filtered out of the text output by eliminating
    lines which end in `/`

  * the behaviors of `globus ls` commands with `--recursive` or `--filter`, but not
    both, are unchanged
```

I'm happy for this to get further refined, but I think we need to explain how they behave when combined here, and it's good to note that we aren't changing how they behave when _not_ combined.

src/globus_cli/commands/ls.py
```diff
@@ -180,29 +180,32 @@ def ls_command(
...
-        # surprising
-        # if we're asked to change or "improve" the behavior in the future, we
-        # could do so with "type:dir" or "type:file" filters added in, and
-        # potentially work out some viable behavior based on what people want
+
+        # if we are doing filtering we need to pass multiple filter params. The
+        # first allows all directories, as we need them for recursive
+        # expansion. The second then filters name by the filter_val
+        if filter_val:
+            ls_params["filter"] = ["type:dir", f"name:{filter_val}"]
```
sirosen, line 197: If I recall our formatting code correctly, we can also leave the `:` formatting to the SDK, like so:
```suggestion
            ls_params["filter"] = [{"type": "dir"}, {"name": filter_val}]
```

Do we like this better? I think it's pretty much the same, but maybe others will find it easier to read or have a stronger opinion.

src/globus_cli/services/transfer/recursive_ls.py
```diff
@@ -60,13 +58,11 @@ def __init__(
         ls_params: dict[str, t.Any],
         *,
         max_depth: int = 3,
-        filter_after_first: bool = True,
```
sirosen, line 63: Was this param simply unused? I can't recall why we needed it, but it seems like maybe just a vestige (in which case, 👍 for removing it!).

sirosen (commented):
I like that we're including the directory names in the output. I don't know if we discussed and agreed on that, but it's easier to ask people to filter them out (e.g. with `grep`) if they don't want them than it is to control in the CLI and give an option.

src/globus_cli/services/transfer/recursive_ls.py
aaschaer: Yeah, I was worried it was something I'd have to figure out how to support, but it was always passed as True, so I just removed it

src/globus_cli/commands/ls.py
aaschaer: Oh, I like that a bit more, I'm just going to add it

aaschaer:
> I like that we're including the directory names in the output. I don't know if we discussed and agreed on that, but it's easier to ask people to filter them out (e.g. with grep) if they don't want them than it is to control in the CLI and give an option.

That was one of the things I wanted to make sure we were all okay with. The original story suggests "And client-side, filter out directories (?)." Which I would be fine adding, but it wasn't clear if we would want to filter out all directories or just directories that didn't match the filter.

changelog.d/20221117_103859_aaschaer_recursive_filter.md
aaschaer: This is significantly better, so I'm going to add the suggested commit, and then we can modify as needed

kurtmckee (approved):

PR #712, opened 2022-11-29
Add mypy as a default tox environment

kurtmckee:
Noticed that mypy wasn't running by default, which resulted in avoidable failures in CI when publishing a PR.


sirosen (approved):
Yep, I've been bitten by this too in the past. Good catch!

PR #713, opened 2022-12-05
Fix 'globus list-commands', improve walk_contexts

sirosen:
We had a walk_contexts helper which was present only in `reference/_generate.py` which could correctly crawl a click CLI. Move this into `globus_cli.utils` for reuse.

Remove the unused `skip_adoc` behavior for commands, which is nowadays covered by `hidden=True`, and remove support for this from `walk_contexts`.

Refactor `list-commands` to leverage `walk_contexts`, and add a test which exercises `list-commands` to ensure it keeps working.

This also effectively means that a chunk of `reference/_generate.py` is now being exercised by our testsuite, whereas it was previously only run by `tox -e reference` or equivalent usage.


sirosen:
I had to push fix because we have `mypy.ini` set to start guiding us towards `disallow_untyped_defs`, which I had missed during my local tests.

Luckily, `mypy` now supports recursive type aliases, so we have a way to accurately describe our walk function.

kurtmckee (approved):

PR #714, opened 2022-12-05
Add initial codeowners

sirosen:


ada-globus (approved):
This looks reasonable to me. I'm not able to interactively preview how GH will interpret this, but based on my understanding, this matches my expectations!

kurtmckee (approved):

PR #717, opened 2022-12-05
Add a test to prevent untyped defs backsliding

kurtmckee:
Exactly what it says on the tin.

I found it necessary to remove the extensive exceptions and default values in `mypy.ini` because mypy was passing the checks with flying colors despite the CLI argument to `disallow-untyped-defs`. So, instead of creating a bare-bones `mypy-untyped-defs.ini` file with zero file exceptions, I instead removed all of the file exceptions in `mypy.ini`. This necessitated removing the `disallow_untyped_defs = true`, which would have caused the existing mypy check to fail after all of the file exceptions were removed.

Output looks like this if the number of files with untyped defs increases:

```
mypy_untyped_defs develop-inst-noop: /home/kurt/dev/globus-cli
mypy_untyped_defs installed: attrs==22.1.0,certifi==2022.9.24,cffi==1.15.1,charset-normalizer==2.1.1,click==8.1.3,coverage==6.5.0,cryptography==36.0.2,execnet==1.9.0,-e git+ssh://git@github.com/globus/globus-cli.git@9c80e391dab2a90f634bbd7580a44e6b86223ba0#egg=globus_cli,globus-sdk==3.15.0,idna==3.4,iniconfig==1.1.1,jmespath==0.10.0,mypy==0.991,mypy-extensions==0.4.3,packaging==21.3,pluggy==1.0.0,py==1.11.0,pycparser==2.21,PyJWT==2.6.0,pyparsing==3.0.9,pytest==6.2.5,pytest-forked==1.4.0,pytest-timeout==1.4.2,pytest-xdist==2.5.0,requests==2.28.1,responses==0.17.0,ruamel.yaml==0.17.16,six==1.16.0,toml==0.10.2,types-cryptography==3.3.23.2,types-jmespath==1.0.2.1,types-jwt==0.1.3,types-requests==2.28.11.5,types-urllib3==1.26.25.4,typing_extensions==4.4.0,urllib3==1.26.13
mypy_untyped_defs run-test-pre: PYTHONHASHSEED='359413110'
mypy_untyped_defs run-test: commands[0] | bash -ec 'if [ "$(mypy --disallow-untyped-defs src/ | grep -oE "Found.+([0-9]+) errors.+([0-9]+) files" | cut -d\  -f5)" -gt $MAX_UNTYPED_DEFS_FILE_COUNT ]; then echo -e "\n\nThe number of files with untyped definitions has increased.\n\n"; false; fi'


The number of files with untyped definitions has increased to 150.


ERROR: InvocationError for command /usr/bin/bash -ec 'if [ "$(mypy --disallow-untyped-defs src/ | grep -oE "Found.+([0-9]+) errors.+([0-9]+) files" | cut -d\  -f5)" -gt $MAX_UNTYPED_DEFS_FILE_COUNT ]; then echo -e "\n\nThe number of files with untyped definitions has increased.\n\n"; false; fi' (exited with code 1)
```


tox.ini
```diff
@@ -6,6 +6,7 @@ envlist =
     cov-combine
     cov-report
     mypy
+    mypy_untyped_defs
```
sirosen, line 9: There is no CI step which runs `tox` sans arguments.

Do you want to add this to the lint job?

https://github.com/globus/globus-cli/blob/5775c9fb3c2bfad8bb7077bda516d513f0ac6c01/.github/workflows/build.yaml#L10

Without it being in CI, I think the check is a weaker protection.

tox.ini
```diff
@@ -65,6 +66,22 @@ deps =
...
+allowlist_externals =
+    bash
+deps =
+    mypy==0.991
+    types-jwt
+    types-requests
+    types-jmespath
+commands =
+    bash -ec 'if [ "$(mypy --disallow-untyped-defs src/ | grep -oE "Found.+([0-9]+) errors.+([0-9]+) files" | cut -d\  -f5)" -gt $MAX_UNTYPED_DEFS_FILE_COUNT ]; then echo -e "\n\nThe number of files with untyped definitions has increased.\n\n"; false; fi'
+
```
sirosen, line 84: As you mentioned yesterday, this is a viable workaround for quickly and easily getting this working.
I'm fine with it, in part because I'd like to see us make strides towards removing it (so the discomfort we might feel with "bash in tox" is offset by the fact that it is meant to have a limited lifetime).

Even so, I'm not sure what I can do other than laugh!
Good thing we don't have a long-lines linter for our config files! 😂

tox.ini
kurtmckee: Right you are! Update on its way.

tox.ini
kurtmckee: Hmm, this check no longer works after rebasing on `main`. I'll take a quick look at this.

sirosen:
We discussed briefly in Slack that this seems to be more deeply in conflict with my recent change (#722) than we realized at first. In particular, the new config means that `--disallow-untyped-defs` no longer does anything on its own when supplied to the `mypy` CLI.

If we were deeply dedicated to this approach (counting violators and doing coverage-fail-under style enforcement), I can think of workarounds. However, I think we like the config which explicitly enumerates files, since it gives us good direction in terms of where to apply fixes.

There's a related-but-not-quite-equivalent change we could do, which is to turn on `--warn-unused-configs`. But that flag only works outside of incremental mode, so it needs to be
```
$ mypy --no-incremental --warn-unused-configs src/
Warning: unused section(s) in mypy.ini: [mypy-globus_cli.commands.foo]
```

And it doesn't seem to cause `exit(1)`, so we may still need some output pattern matching to turn it into an error.

---

Thinking this through, I think we can't merge it because of the conflict, and I value #722 enough that I don't want to revert it. I'm recommending that we close this, and optionally we can add a build step to check for unused configs as a separate (somewhat related) task.

@kurtmckee, as the original PR author, are you amenable to that? Or do you want to either make the case for this over #722 or that the two can be reconciled (e.g. with a secondary mypy config)?

kurtmckee:
Nope, I'm fine closing this. C'est la vie.

PR #719, opened 2022-12-05
Fix changelog formatting

ada-globus:
Signed-off-by: Ada Nikolaidis <ada@globus.org>


sirosen (approved):
For others who see this, we realized there were some formatting issues and agreed to fix prior to the release.

PR #721, opened 2022-12-06
fix session show fields

aaschaer:
`globus session show` is currently broken! Ran into this testing `session update --policy` work.

Would be good to add a unit test as well, but I'll open the PR as is for now, and come back if I have time.

### DIFF
--- a/src/globus_cli/commands/session/show.py
+++ b/src/globus_cli/commands/session/show.py
@@ -1,4 +1,4 @@
-import time
+import datetime
 
 import globus_sdk
 
@@ -88,8 +88,10 @@ def session_show(*, login_manager):
         {
             "id": key,
             "username": resolved_ids.get(key, {}).get("username"),
-            "auth_time": time.strftime(
-                "%Y-%m-%d %H:%M %Z", time.localtime(vals["auth_time"])
+            "auth_time": (
+                datetime.datetime.fromtimestamp(vals["auth_time"])
+                .astimezone()
+                .strftime("%Y-%m-%d %H:%M %Z")
             ),
         }
         for key, vals in authentications.items()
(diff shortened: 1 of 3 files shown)


src/globus_cli/commands/session/show.py
```diff
@@ -94,8 +100,10 @@ def session_show(*, login_manager):
         "  globus whoami\n"
     )
 
+    print(list_data)
```
sirosen, line 103: This looks like a debugging print. Let's remove this and add a test.

sirosen (changes requested):
I'm going to open a PR against this PR to add an appropriate test.

I'd like to see us push v3.10.1 rather quickly since the command is completely broken, but I also want to get the guard against this in place.

src/globus_cli/commands/session/show.py
aaschaer: Ah, yes it is

sirosen:
I've just merged #724 into here, but I hink it's necessary to rebase, based on the build failures we were seeing. I'm going to do that rebase now, which should fix the CI behavior (cross reference: #725).

sirosen (approved):
I'm pretty sure we're good to merge on this (also, kudos for finding the bug!); I'm just waiting for CI to pass. 🙂

sirosen:
The fixed build (after rebasing) revealed a minor bug running the tests on Windows (`time.strftime("%s", ...)` is not valid). I just checked independently to verify that using `time.mktime` is an easy solve for this and I've pushed it to this branch.

Normally, I'd take a more laid-back approach, but I very much want us to get this merged so that we can release it. I intend to squash-merge if the tests pass cleanly this time.

PR #728, opened 2022-12-13
annotate commands/_common.py

aaschaer:
Thought I'd start at the top of the command modules section and work down?

Is one module per PR the right way to do this, or would it be better to do one module per commit and several per PR?

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -128,9 +128,6 @@ disallow_untyped_defs = false
 # command modules
 # listed separately to preserve the readability of the above config
 
-[mypy-globus_cli.commands._common]
-disallow_untyped_defs = false
-
 [mypy-globus_cli.commands.api]
 disallow_untyped_defs = false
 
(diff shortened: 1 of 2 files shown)


kurtmckee (approved):

sirosen:
> Is one module per PR the right way to do this, or would it be better to do one module per commit and several per PR?

I prefer that we do them one-by-one. It keeps the reviews very bite-sized and easy.

If you were to do several and submit a single PR, I think that would be fine as well, as long as none of the modules has subtle changes which you think need much review. The bad case there is that you have 8 changes, 7/8 are good, but 1/8 needs discussion and slows the rest down.

sirosen (approved):
I just took a minute to double-check that the failures are only the things which are fixed (just now) in `main`. Looks good!

I'm merging, and if by some disaster/miracle it fails in `main`, I'll take point on fixing it.

PR #729, opened 2022-12-13
remove timezone unaware call in test_session_show_text

aaschaer:
This test was failing for me locally, I believe because mktime can only handle time_structs that are in local time.

Skimming the python time and datetime docs I didn't see an obvious way to fix this, and hardcoding a test value to match another hardcoded test value seemed a reasonable-ish solution. Any other ideas would be welcome though.

### DIFF
--- a/tests/functional/test_session_show.py
+++ b/tests/functional/test_session_show.py
@@ -84,7 +84,11 @@ def create_introspect_data(user_attrs):
                     "acr": None,
                     "amr": None,
                     "idp": user_attrs["idp_id"],
-                    "auth_time": 613217460,
+                    "auth_time": datetime.datetime.strptime(
+                        user_attrs["auth_time"], "%Y-%m-%d %H:%M %Z"
+                    )
+                    .replace(tzinfo=datetime.timezone.utc)
+                    .timestamp(),
                     "custom_claims": {},
                 }
             },
--- a/tests/functional/test_session_show.py
+++ b/tests/functional/test_session_show.py
@@ -1,5 +1,4 @@
 import datetime
-import time
 import uuid
 
 import pytest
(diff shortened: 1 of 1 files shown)


kurtmckee (changes requested):
Try something like this, which appears to result in the hard-coded value `613217460` as expected:

```python
import datetime

auth_time = "1989-06-07 10:11 UTC"
desired = 613217460
actual = int(
    datetime.datetime.strptime(auth_time, "%Y-%m-%d %H:%M %Z")
    .replace(tzinfo=datetime.timezone.utc)
    .timestamp()
)

print(f"desired={desired}, actual={actual}")
```

aaschaer:
> Try something like this, which appears to result in the hard-coded value `613217460` as expected:
> 
> ```python
> import datetime
> 
> auth_time = "1989-06-07 10:11 UTC"
> desired = 613217460
> actual = int(
>     datetime.datetime.strptime(auth_time, "%Y-%m-%d %H:%M %Z")
>     .replace(tzinfo=datetime.timezone.utc)
>     .timestamp()
> )
> 
> print(f"desired={desired}, actual={actual}")
> ```

Ah yeah, adding the tzinfo back works! Still have to hardcode the timezone, but that's better than hardcoding the whole value, could reasonably pass the timezone as an argument to the helper or something if this gets used elsewhere.

kurtmckee (approved):

sirosen (approved):
The build failures are unrelated (I'm working on that).

I was going to suggest something like what Kurt offered -- it's nice to source the time from the same place. If it weren't possible to do at all, having a matching value (as in the original) wouldn't have been a huge issue, but this is nicer.

Thanks for catching this! It's my fault for keeping my work box in UTC! 😁

PR #730, opened 2022-12-13
Fix builds for tox v4 and any other changes

sirosen:
I suspect it won't just be tox again, but I'm starting with the allowlist_externals tweak.

---

Changes ended up being:
- allowlist_externals
- use tox v3 when testing a very old version of `virtualenv` (there is a story behind this test)


aaschaer (approved):
Sorry if this review is premature, but the build is green and the changes look good to me so I'm going ahead and leaving an approval

kurtmckee (approved):

PR #733, opened 2022-12-14
handle session_required_policies errors

aaschaer:
For https://app.shortcut.com/globus/story/18757/sdk-and-cli-support-for-session-required-policies

Opening as a draft as this depends on https://github.com/globus/globus-sdk-python/pull/658 and tests will fail without it.

As of yesterday Transfer released the fixed `session_required_policies` field name to all envs, and Auth support should also be on all envs now, so I think this is stable enough to at least open the PR. No rush to release until GCS support is live though, as that is where the UI (and docs?) will be.

I've tested end-to-end on sandbox, and can help with the needed `globus api` calls if anyone else wants to try it out.

### DIFF
--- a/tests/functional/test_exception_hooks.py
+++ b/tests/functional/test_exception_hooks.py
@@ -5,8 +5,8 @@ from globus_sdk._testing import RegisteredResponse, load_response, load_response
 
 def test_session_required_policies(run_line):
     """
-    ls on hits a 403 with session_required_policies set
-    confirms correct globus session update command in helptext
+    confirms a correct `globus session update` command is shown in helptext
+    after hitting a 403 with session_required_policies set
     """
     meta = load_response_set("cli.transfer_activate_success").metadata
     ep_id = meta["endpoint_id"]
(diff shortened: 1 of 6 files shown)


tests/functional/test_exception_hooks.py
```diff
@@ -0,0 +1,34 @@
+import uuid
+
+from globus_sdk._testing import RegisteredResponse, load_response, load_response_set
+
+
+def test_session_required_policies(run_line):
+    """
+    ls on hits a 403 with session_required_policies set
+    confirms correct globus session update command in helptext
```
kurtmckee, line 9: This explanation isn't reading correctly the first and second time I read it. Would you see if there's a way to better explain this?

kurtmckee (changes requested):

tests/functional/test_exception_hooks.py
aaschaer: Ah yeah, that needs another pass

kurtmckee (approved):

sirosen (approved):

PR #738, opened 2022-12-20
Add click annotation checker and begin using it

sirosen:
The annotation checker has specialized knowledge of how click params map back to types, and customizations to handle the param types defined in the CLI.

It encodes logic for mapping the parameter list for command objects back to expected sets of annotated arguments. It then checks annotations *in a runtime, test context* against those expectations and raises errors if there are detected deviations.

The new helper is applied to several recently annotated commands and has the following basic results:
- `--notify` needs to be moved to a custom param type so that it can be understood by the annotation checker
- `--batch` was annotated incorrectly in `globus transfer`; missing `| None` in the annotation

Importantly, `--notify` heralds a potential sweeping change, that any custom callback may require special handling. As an alternative, we could introduce custom subclasses of Option and Argument which allow us to specify known types as part of the runtime specification.

The tests only run on py3.10+ in order to allow use of `types.UnionType` and have other useful features.

---

In a second commit, I introduced tests for the new `NotificationParamType`, assuming these would be a somewhat trivial extension of existing unit tests for various param types to give us better coverage. This revealed a bug which the first commit had introduced, and which is fixed therein.

---

Overall impressions after pursuing this strategy:

The parameter-to-type translation is nontrivial. I originally hoped that this could be made solid enough to offer up as a standalone package in the `click` ecosystem, but there are numerous globus-cli specific bits of logic. In some cases, the deduction of correct types is impossible in the general case, but we can rely on internal conventions to state things in a way that the checker can understand.

However! This has already revealed an incorrect annotation and has been applied only to a narrow subset of the CLI. I think it is worth continuing to pursue this new body of code in order to ensure that our various annotations are accurate.


src/globus_cli/parsing/param_types/notify_param.py
```diff
@@ -0,0 +1,63 @@
...
+        # on means don't set anything (default)
+        if on:
+            return {}
+        # off means turn off everything
+        if off:
+            return {
+                "notify_on_succeeded": False,
+                "notify_on_failed": False,
+                "notify_on_inactive": False,
+            }
```
kurtmckee, line 56: It took a little bit to grok this code, as I needed to follow these variables:

* `value` (the string)
* `value` (the list)
* `on` and `off` (bools)
* `vals` (the set excluding "off" and "on")

I took a crack at refactoring this. It subtly changes the behavior, as (for example) `","` would now be converted to "off" as well, and `"off,"` would be accepted as "off". It also isolates and rejects invalid values by name; for example, `--notify succeded,failed` will now explicitly call out the misspelling of "succeeded".

```python
# Build a set of all values, excluding empty strings.
values = {v.strip() for v in value.lower().split(",")} - {""}

# You'll get an empty set if value is "" or "," to start with.
# Special-case it into "off", which helps avoid surprising scripts
# which take notification settings as inputs and build '--notify ""'.
if not values:
    values = {"off"}

# Disallow invalid values.
invalid_values = values - {"off", "on", "succeeded", "failed", "inactive"}
if invalid_values:
    raise click.UsageError(
        "--notify received these invalid values: {}".format(
            list(invalid_values)
        )
    )

# Disallow combining "off" with other values.
values_other_than_off = values - {"off"}
if "off" in values and values_other_than_off:
    raise click.UsageError('--notify cannot accept "off" and another value')

# return the notification options to send!
# on means don't set anything (default)
if "on" in values:
    return {}
# off means turn off everything
if "off" in values:
    return {
        "notify_on_succeeded": False,
        "notify_on_failed": False,
        "notify_on_inactive": False,
    }
```

What are your thoughts on this?

tests/click_types.py
```diff
@@ -0,0 +1,163 @@
...
+    if isinstance(param_type, click.types.StringParamType):
+        return (str,)
+    if isinstance(param_type, click.types.BoolParamType):
+        return (bool,)
+    elif isinstance(param_type, (click.types.IntParamType, click.IntRange)):
+        return (int,)
+    elif isinstance(param_type, (click.types.FloatParamType, click.FloatRange)):
+        return (float,)
+    elif isinstance(param_type, click.Choice):
+        return (t.Literal[tuple(param_type.choices)],)
```
kurtmckee, line 45: I'm not familiar with this syntax. Does this tuple unpacking, or is it valid as-is?

```suggestion
        return (t.Literal[*tuple(param_type.choices)],)
```

tests/click_types.py
```diff
@@ -0,0 +1,163 @@
...
+        #   '--foo/--no-foo' is a bool flag with an explicit default of None
+        #   '--foo' is a count option with a default of 0
+        #   '--foo' uses a param type which converts None to a default value
+        if isinstance(param_obj, click.Option):
+            if _option_defaults_to_none(param_obj):
+                possible_types.add(None.__class__)
+
+        for param_type in _param2types(param_obj):
+            if param_obj.multiple:
+                param_type = list[(param_type,)]
```
kurtmckee, line 135: Seeing `list[(param_type,)]` is starting to reinforce that passing a tuple into a type annotation like this is the correct syntax.

tests/unit/param_types/test_notification_param.py
```diff
@@ -0,0 +1,89 @@
+import click
+import pytest
+
+from globus_cli.parsing.known_callbacks import none_to_empty_dict
+from globus_cli.parsing.param_types import NotificationParamType
+
+
+@click.command()
+@click.option("--notify", type=NotificationParamType(), callback=none_to_empty_dict)
```
kurtmckee, line 9: There's a check in `NotificationParamType` that can return `None`, but I don't see a test for this here.

src/globus_cli/parsing/param_types/notify_param.py
```diff
@@ -0,0 +1,63 @@
+import click
+
+
+class NotificationParamType(click.ParamType):
+    def get_metavar(self, param) -> str:
+        return "{on,off,succeeded,failed,inactive}"
+
+    def convert(self, value, param, ctx):
```
kurtmckee, line 8: This is missing type annotations.

src/globus_cli/parsing/param_types/notify_param.py
```diff
@@ -0,0 +1,63 @@
...
+        """
+        Parse --notify
+        - "" is the same as "off"
+        - parse by lowercase, comma-split, strip spaces
+        - "off,x" is invalid for any x
+        - "on,x" is valid for any valid x (other than "off")
+        - "failed", "succeeded", "inactive" are normal vals
+
+        In code, produces True, False, or a set
```
kurtmckee, line 17: This comment doesn't appear to match current code behavior.

kurtmckee (changes requested):

src/globus_cli/parsing/param_types/notify_param.py
sirosen: I'm more than happy to incorporate this change!

I had only done a lift-and-shift of the existing code, and I agree that it's hard to read.
Overall the behavioral change looks to me like an improvement too.

Look forward to a `Co-authored-by` commit. 😉

tests/click_types.py
sirosen: Apparently, both of these are valid. The argument to `__class_getitem__` is a tuple, but it looks like unpacking is automatic. Here's a little sample from the console:
```
>>> import typing as t
>>> x = (int, str)
>>> y = t.Literal[x]
>>> y
typing.Literal[int, str]
>>> z = t.Literal[*x]
>>> z
typing.Literal[int, str]
>>> w = t.Literal[(x,)]
>>> w
typing.Literal[(<class 'int'>, <class 'str'>)]
```

So that looks like `y` and `z` are equivalent, but `w` is improperly formed/invalid.

I don't have a preference between passing a tuple or an unpacked tuple, but I'd like to be consistent in how we do these constructions. Given the choice, which one would you prefer?

---

I find the syntax here a little odd, and I think the use-cases (like this one) for dynamically constructing types somewhat surprised the original authors of the type annotation syntax and PEPs. However, given the recent course-corrections the SC has made with respect to typing and wanting to support runtime usages, I feel okay about using this.

tests/unit/param_types/test_notification_param.py
sirosen: I think I might need to dig more deeply into how to test that behavior, for all of our parameter types.
There's no way to trigger that with "basic" usage, like these tests are doing. I _think_ that the value can be `None` when the shell completer is running. That was the case in past versions of `click`, so I've retained that habit, but I also know that the completer was largely rewritten in v8.0.

The `None` case might not even be possible anymore -- I might be operating on an old paradigm.

src/globus_cli/parsing/param_types/notify_param.py
sirosen: Good call! I'll apply that fix and the docstring change, noted below.

tests/click_types.py
sirosen: I thought it was necessary, but after my little experiment on this, shown above, it might be valid to write `list[param_type]` as well.
If so, that might make me tend towards preferring unpacked tuples in general. `list[param_type]` reads better, I think.

ada-globus, line 135: Is it possible in the `param_obj.multiple` scenario that we also need to include a list containing any combination of the `param_type`s? Or is that something we know we needn't consider by virtue of the fact that the cases where this applies are limited to the `OrNull`s?

ada-globus (commented):
This looks great and is a nice addition!

tests/click_types.py
```diff
@@ -0,0 +1,163 @@
...
+    for param in f.params:
+        # skip params which do not get passed to the callback
+        if param.expose_value is False:
+            continue
+
+        param_map[param.name] = param
+
+    for name in param_map:
+        if name not in hints:
+            raise ValueError(f"expected parameter '{name}' was not in type hints")
```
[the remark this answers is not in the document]
sirosen: Yep, this seems like a reasonable step!

I also might _remove_ `param_map`. I'm not sure if it's necessary to the logic, and it might be clearer to look repeatedly at `f.params`.
So it's possible that I'll do more extensive cleanup in a little bit.

tests/click_types.py
sirosen: It is possible, at least in theory, that `multiple` parameters will have a type which is a union.
I can imagine, for example,
```python
@click.option("--foo", type=IntOrStr(), multiple=True)
```

I think it's worth rearranging the logic to at least try to handle this, but without a clear example case in which we can apply this, it will likely go unverified for the near term.

The capabilities of this testing tool will have to grow in various directions as we apply it to more and more chunks of the codebase. But it's hard to anticipate all of the possible options today. So I'm trying to build us a reasonable "baseline" which we can grow as we encounter more cases, and get things "mostly right" at the start. But if there's a gap or unhandled case, I'm okay with that.

tests/click_types.py
```diff
@@ -29,7 +28,22 @@
...
+
+    Supports both `click` native types and `globus_cli` custom types
+
+    For example:
+        IntParamType -> (int,)
+        IntRange -> (int,)
+        StringOrNull -> (str, None)
+
+    Returning types as a tuple rather than building a union when there is more than one
+    ensures that we build "flat" unions of the form `Union[x, y, z]` and avoid nexted
```
kurtmckee, line 44:
```suggestion
    ensures that we build "flat" unions of the form `Union[x, y, z]` and avoid nested
```

kurtmckee (changes requested):

tests/click_types.py
sirosen: Thanks! I just amended to include this. 🤫

tests/click_types.py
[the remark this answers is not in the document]
sirosen: The refactor I did was more extensive, but should cover this.

tests/click_types.py
sirosen: I was working through several changes here and I found that it was hard or at least fiddly to make things totally consistent.

`flake8` flags the following syntax as invalid (`E999`):
```
Literal[*tuple(choices)]
```

Now... it is not invalid, but I can understand why it looks that way. `__getitem__` takes a single arg, and I'm guessing the flake8 parser doesn't have updates to handle `__class_getitem__`.

So we could ignore this, but breaking a parser which is being used on our code seems unwise. So I'm ending up mixing styles here based on the usage.

`tuple[param_type, ...]` gets used in one place, `Literal[tuple(choices)]` in another.

I think this is okay, if a _little_ unfortunate, and it just needs to be part of our understanding of how runtime evaluation of types works.

tests/unit/param_types/test_notification_param.py
sirosen: I confirmed as part of my work today on how shell completion now behaves that this is an older paradigm, from `click<8.0`, and we don't receive `None` in any usage anymore. So I've removed the check, rather than adding a test for it.

src/globus_cli/parsing/param_types/notify_param.py
sirosen: This was a good catch, thank you! I should have it nicely fixed, but please let me know if it should be reworded at all.

tests/click_types.py
sirosen: Now that there are self-tests for the click type-checker, I was able to construct some scenarios to ensure that it works with `multiple=True`, union types, and `nargs=-1`.

sirosen:
My latest changes today refactored the core logic pretty significantly, so please take a look. One major change here is that the union comparison has changed.

Previously, this did:
- collect types for a union based on runtime state
- extract types from any union annotation
- compare the two sets

Now, it's doing:
- collect types for a union based on runtime state
- combine those types into a Union type
- compare types for equality

The result is more uniform logic, as it coincides with the logic which compares non-union types.

src/globus_cli/parsing/param_types/notify_param.py
```diff
@@ -0,0 +1,117 @@
...
+        if "," not in incomplete:
+            return [CompletionItem(o) for o in all_options if o.startswith(incomplete)]
+
+        # grab the last partial name from the list
+        # e.g. if the caller used `--notify inactive,succ<TAB>`, then
+        #      collect `succ` as the last incomplete fragment
+        #
+        # also collect the valid completed parts into a set for comparisons
+        last_incomplete_fragment = incomplete.split(",")[-1]
+        already_contains = set(incomplete.split(",")[:-1])
```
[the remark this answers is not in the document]
sirosen: For me, the left-hand-size unpacking syntax falls into the same category as `for-else`.
I can read it when I see it, but it has never made it into my own vocabulary for writing!

I don't know that I frequently see cases for it as good as this one. Let's apply this rewrite. I'll read the other comment to understand your thoughts regarding leading commas.

src/globus_cli/parsing/param_types/notify_param.py
```diff
@@ -0,0 +1,117 @@
...
+            return [CompletionItem(with_trailing_comma + o) for o in possible_options]
+
+        # if the last fragment is nonempty, trim it and replace it with the valid
+        # completions
+        without_last_incomplete = incomplete[: -len(last_incomplete_fragment)]
+        return [
+            CompletionItem(without_last_incomplete + o)
+            for o in possible_options
+            if o.startswith(last_incomplete_fragment)
+        ]
```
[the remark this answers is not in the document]
sirosen: I'm wondering if we should preserve `already_contains` _except_ for empty strings.
That is...
```python
*already_contains, last_fragment = incomplete.lstrip(",").split(",")
already_contains = [s for s in already_contains if s != ""]
```

I'm also thinking maybe we should make sure `""` is accepted (and discarded) by the parsing.

The two options, as I see it, are these two completions:
```
# if we trim empty strings
,inactive,,s -> inactive,succeeded

# if we allow but ignore empty strings
,inactive,,s -> ,inactive,,succeeded
```

I also want to double-check that the bash completer and zsh completer both have the same behavior if we rewrite the beginning of the string to remove excess commas. I think they both work, but I'd like to be sure.

src/globus_cli/parsing/param_types/notify_param.py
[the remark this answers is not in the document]
kurtmckee: It occurs to me that, rather than using `.lstrip()`, outright filtering might be a better play, specifically for cases like `"succeeded,,failing,,"`:

```python
*existing_options, fragment_option = incomplete.split(",")
existing_options = [i for i in existing_options if i]

possible_options = sorted(set(all_compoundable_options) - set(existing_options))

return [
    CompletionItem(",".join(existing_options + [o]))
    for o in possible_options
    if o.startswith(fragment_option)
]
```

src/globus_cli/parsing/param_types/notify_param.py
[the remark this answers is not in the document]
kurtmckee: Haha, we had the exact same idea re: filtering! Awesome!

There may be technical restrictions in the shells, and that's completely unknown to me, so I'm glad you're thinking of it! I would prefer to perform outright cleanup during completion, simply because it's what I'm familiar with in some environments. For example, I'm accustomed to Powershell swapping the direction of path separators the moment I press `[Tab]`, so this type of "DO IT RIGHT, @#$%!" opinionated completion is something I've come to accept.

kurtmckee (approved):

PR #743, opened 2023-01-09
Consolidate the "setattr" parameters for endpointish create and update commands

sirosen:
# 🗺️ Context

Multiple flavors of "endpointish" (i.e. those things which have `entity_type`) have distinct create and update commands.
I believe the full list is
- `globus endpoint create` (deprecated, hidden)
- `globus endpoint update`
- `globus gcp create guest`
- `globus gcp create mapped`
- `globus collection update`

Plus the following commands which _should_ be created in the future:
- `globus collection create mapped`
- `globus collection create guest`

There is _some_ sharing of parameters between these via common decorators, which helps, but there's also a _lot_ of duplication.

In particular, several distinct commands have mirror versions of the same parameter. For example, `--keywords` defined such that it passes its value as a list vs as a string. Or `--description` defined to either convert `""` to `null` or not. These differences are mostly transparent to the user, but are part of why the helpers haven't been unified in the past.

## 🔍 Annotation Inspection

With the introduction of the click annotation checker tests, it's now possible to rewrite these decorators with a much higher degree of confidence. Whereas previously rewriting an option required carefully cross-checking it against the usage, it is now possible to rely on the annotation checker to ensure that the type matches. This is even more clearly the case with options buried in large shared decorators. It was difficult to change a behavior and ensure it was properly propagated to all usage sites.

Annotation checking is not fully the same thing as correctness checking, and it doesn't replace testing. But it makes the process simpler by a significant margin.

# 🔧 The Main Change: `endpointish_params`

We've tried a few ways of writing shared decorators for bulk options, both in the CLI and elsewhere, and had mixed results. Even within the course of this PR authorship, there are a few lessons learned and course corrections.

The core of this change is the introduction of `endpointish_params`, the new and preferred way for us to pass these collective options.

Ultimately, it's hard to try to define some abstraction which properly captures variations between commands.
Instead, this has ended up with a rather direct approach
- `endpointish_params.create` is the "create" version, `endpointish_params.update` is the "update" version
- instruct the decorator if the entity should be called a `collection` or `endpoint` in helptext
- allow overriding certain default behaviors (like `keyword_style="list"`)

That is, there is no high-level abstraction, or any attempt to map `entity_type` values to relevant options.

## 💥 Functional Changes

Most of these changes are harmless and backwards-compatible, but there are some exceptions.
- Several options can now pass explicit nulls with `""` instead of a literal `""`. This mainly applies to `globus endpoint update` options.
- `globus gcp create [mapped|guest]` commands now use "choice" style for "verify" options. This is a newer and nicer way of specifying some mutually exclusive options, like `force_verify` and `disable_verify`. `--disable-verify/--no-disable-verify` is kept for backwards compatibility for now, but warns when used in interactive contexts.

~Since the compatibility contract of the CLI is slightly vague, I think we're in good standing to modify the very newly added options to a better style. I don't think we can make the same call for the longer-standing `globus endpoint update` command.~

## 🏁 Type Checker Changes

In order to get accurate information to the click annotation checker, it's necessary to make some changes. The most significant is the introduction of classes for type-annotated Parameters (`AnnotatedOption`) and Parameter Types (`AnnotatedParamType`).

These classes allow us to declare the type information either statically or via callbacks for presentation back to the click annotation checker. This lets us work around some cases in which the checker can't be expected to deduce a correct type.

## 🧹 Cleanup

- The annotated param type changes allowed for quite a bit of cleanup to the click type checker
  - click types are specified in a static map where possible
  - all unrecognized param types are asked for their callback rather than hardcoded logic
  - the known callbacks disappear in favor of a passed annotation for parameters using those callbacks
- `ExplicitNullType.nullify` and `ExplicitNullType.nullify_dict` are new internal operations to make dictionary scrubbing simpler
- `globus_cli.utils.fold_decorators` is now used to apply "stacks" of decorators to functions
- new methods in the click type checker: `collection update`, `endpoint create`, `endpoint update`, `gcp create guest`, `gcp create mapped`
- `endpointish_params.legacy_transfer_params` is added for the `globus endpoint [create|update]` commands to share options (rather than that being kept in a separate module)

----

EDIT: 2023-01-10, I updated the message here to reflect the current state of the PR.


changelog.d/20230109_172604_sirosen_consolidate_endpointish_setattr.md
```diff
@@ -0,0 +1,14 @@
+### Other
+
+* Improve the uniformity of endpoint and collection option parsing.
+** The `--sharing-restrict-paths` option to `globus collection update` now
+   checks for invalid types (non-dict, non-null data)
+** `globus endpoint update` can now null values which previously could not be
+   unset by passing `""` for `--contact-email`, `--contact-info`,
```
sirosen, line 7: TODO: rewrite

`globus endpoint update` will now treat the empty string as `null` for the following options: ...

src/globus_cli/parsing/param_classes.py
```diff
@@ -1,14 +1,50 @@
...
+
+_SENTINEL = _Sentinel()
+
 
-class OneUseOption(click.Option):
+class AnnotatedOption(click.Option):
+    def __init__(
+        self,
+        *args: t.Any,
+        type_annotation: type | _Sentinel = _SENTINEL,
```
derek-globus, line 21: Can this value validly be `None`?

Confused why we'd introduce the `_SENTINEL` "i didn't specify it" singleton.

src/globus_cli/parsing/param_classes.py
```diff
@@ -64,14 +100,13 @@ def one_use_option(*args, **kwargs):
     if kwargs.get("is_flag"):
         kwargs["is_flag"] = False  # mutually exclusive with count
         kwargs["count"] = True
+        if "type_annotation" not in kwargs:
+            kwargs["type_annotation"] = bool
```
derek-globus, line 104: Why?

src/globus_cli/parsing/param_types/comma_delimited.py
```diff
@@ -16,6 +18,9 @@ def __init__(
         self.convert_values = convert_values
         self.choices = list(choices) if choices is not None else None
 
+    def get_type_annotation(self, param: click.Parameter) -> type:
+        return list[str]
```
derek-globus, line 22: Are the elements in a `CommaDelimitedList` always strings?

src/globus_cli/parsing/shared_options/endpointish.py
```diff
@@ -0,0 +1,319 @@
...
+) -> dict[str, bool]:
+    if value is None:
+        return {}
+    value = value.lower()
+    if value == "force":
+        return {"force_verify": True, "disable_verify": False}
+    elif value == "disable":
+        return {"force_verify": False, "disable_verify": True}
+    else:
+        return {"force_verify": False, "disable_verify": False}
```
derek-globus, line 201:
```suggestion
    return {
        "force_verify": value == "force",
        "disable_verify": value == "disable"
    }
```

src/globus_cli/parsing/shared_options/endpointish.py
```diff
@@ -0,0 +1,319 @@
...
+        )
+
+    return utils.fold_decorators(f, decorators)
+
+
+def _verify_choice_to_dict(
+    ctx: click.Context, param: click.Parameter, value: t.Any
+) -> dict[str, bool]:
+    if value is None:
+        return {}
```
derek-globus, line 194: Should None be treated as default instead of null?

src/globus_cli/parsing/param_classes.py
sirosen: `None` is valid in annotations, so I believe it can be validly expressed as `None`. Such an option would not necessarily be useless, though it would be a little unusual.

For example:
```python
def callback(ctx, param, value):
    warnings.warn("got deprecated option --foo")
    return None

@click.option("--foo", cls=AnnotatedOption, type_annotation=None, hidden=True, callback=callback)
def myfunc(*, foo: None) -> None:
    ...
```

If `--foo` is deprecated, it's not _necessary_ to make its type `None`. The following nearly-equivalent thing is also fine:
```python
def callback(ctx, param, value):
    warnings.warn("got deprecated option --foo")
    return value

@click.option("--foo", hidden=True, callback=callback)
def myfunc(*, foo: str | None) -> None:
    ...
```

But since this is meant to be at the core of the click type checker, I wanted to account for the possibility.

src/globus_cli/parsing/param_classes.py
sirosen: I think this deserves a comment. `one_use_option` is a bit weird (and maybe we should remove it in the future).
It takes any option and converts it to a form which allows us to count and detect multiple usages.

I would describe it as doing the following two transformations:
```python
# --foo is a bool flag
@one_use_option("--foo", is_flag=True)

# rewrites to --foo as a count which converts back to bool
def callback(ctx, param, value):
    if value > 1:
        raise UsageError("bad")
    return bool(value)

@click.option("--foo", count=True, callback=callback)
```
and
```python
# --foo is a non-flag option
@one_use_option("--foo")

# rewrites to --foo as a multiple option which unpacks its only value
def callback(ctx, param, value):
    if len(value) > 1:
        raise UsageError("bad")
    if value:
        return value[0]
    return None

@click.option("--foo", multiple=True, callback=callback)
```

This means that the type checker won't see `is_flag=True` on the created option, but `count=True` instead. Under normal logic, as for `click.option("-v", "--verbose", count=True)`, it will determine that the type is `int`. But that "hidden" callback in `one_use_option` makes it `bool`.

src/globus_cli/parsing/param_types/comma_delimited.py
sirosen: Yes, we're guaranteed it. `convert_values` looks like it could modify things, but it's typed as a `str -> str` function.

src/globus_cli/parsing/param_types/location.py
```diff
@@ -1,18 +1,32 @@
...
+        # ignore for now
+        return tuple[float, float]  # type: ignore[no-any-return,misc]
+
+    def convert(
+        self, value: t.Any, param: click.Parameter | None, ctx: click.Context | None
+    ) -> t.Any:
         try:
             match = re.match("^(.*),(.*)$", value)
+            if not match:
+                raise ValueError()
```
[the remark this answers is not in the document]
sirosen: I think I should do a bigger change here. I didn't like seeing that we're catching `ValueError` and `AttributeError` in the first place. I think I added this as a quick hack to pacify `mypy` and that it deserves more attention.

src/globus_cli/parsing/shared_options/endpointish.py
sirosen: I'm not sure I follow the question.

To explain how I see it:

The callback runs even if no option is passed. So this ensures that we can write
```python
@decorator_here
def foo(..., verify: dict[str, bool]): ...
```
If we made it `None`, then the functions would change
```python
@decorator_here
def foo(..., verify: dict[str, bool] | None): ...
```

And the way that `verify` is being used is in usages like `dict(foo=foo, bar=bar, **verify)`.

We could handle this option a lot of different ways -- I'm not overly attached to what's here -- but I'm not sure what you're proposing as the replacement logic.

src/globus_cli/parsing/shared_options/endpointish.py
sirosen: Yes please and thank you! 😄

src/globus_cli/parsing/param_types/location.py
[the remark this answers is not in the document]
sirosen: I'm going to mark this resolved with the rewrite I just pushed, which redoes this function and adds some unit tests for it. Feel free to reopen, but I think this is fixed.

src/globus_cli/parsing/param_classes.py
sirosen: Upon review, I decided to pull this out and instead pass `type_annotation=bool` in the 3 relevant usage sites.
I considered making the OneUseOption class responsible for this and various other "clever" solutions.

None of them stacked up, IMO, well against accepting that if you want to use `one_use_option("--foo", is_flag=True)`, then you must additionally pass `type_annotation=bool`. There's less to explain this way, which I see as the primary advantage.

src/globus_cli/parsing/shared_options/endpointish.py
derek-globus: Ah that makes sense. I was missing the invocation scope on this one, thanks for explaining

src/globus_cli/parsing/param_types/location.py
```diff
@@ -18,20 +18,27 @@ class LocationType(AnnotatedParamType):
...
     ) -> t.Any:
+        match_result = re.match(r"^([^,]+),([^,]+)$", value)
+        if not match_result:
+            self.fail(
+                f"location '{value}' does not match the expected "
+                "'latitude,longitude' format"
+            )
+
+        maybe_lat = match_result.group(1)
+        maybe_lon = match_result.group(2)
```
derek-globus, line 34: I've never seen `maybe_x` variable naming convention and I genuinely can't tell if I love it or hate it.

derek-globus (approved):
Approve of the change.

I'd ask we wait until Aaron or someone more familiar with the potentially affected API routes approves as well before merging.

src/globus_cli/parsing/param_types/location.py
sirosen: I still remember learning Haskell, which calls optional types "Maybe". `Maybe a` is what python writes as `Optional[a]`.
Somehow, that word, and the way that it adds uncertainty, has stuck with me as something I use from time to time.

Not really an explanation or even a defense of the naming! Just a bit of color. 🎨

sirosen:
Aaron and I discussed this a bit to get him up to speed with the changes, and I've tested this out a bit against Transfer. No new errors are introduced, although `--display-name ""` remains invalid and gets back an API error.

PR #748, opened 2023-01-20
Improve test and CI performance

kurtmckee:
This decreases our CI usage (sum of all CI job runtimes) and decreases our CI total duration (wall clock time that passes to get a pass/fail).

----

*   Configure tox to build a platform-independent wheel and share it among all test environments.

    This requires a change to the code coverage `fail-under` value,
    because coverage now sees three files that are untested:

    *   `src/globus_cli/globus_cli_flake8.py`
    *   `src/globus_cli/login_manager/_old_config.py`
    *   `src/globus_cli/login_manager/local_server.py`

    This change does not result in any performance improvements during local testing.
*   Cache downloaded PyPI packages, Python virtual environments, and tox environments in CI.

    Performance numbers:

    ```
                    +-------------------+-----------+
                    | Total duration    | CI usage  |
    +---------------+-------------------+-----------+
    | No caching    | 2m 39s            | 11m 52s   |
    | Cache miss    | 2m 05s            | 14m 53s   |
    | Cache hit     | 1m 45s            |  8m 25s   |
    +---------------+-------------------+-----------+
    ```

    Note that, after this change is merged, cache misses are expected to be infrequent.
    This is because CI runs in branches have access to caches created in the `main` branch,
    and CI runs against the `main` branch every Monday at 4 A.M UTC.
    Therefore, it is expected that cache hits will be common.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,11 +1,17 @@
 [run]
 parallel = True
-include = src/*
+source =
+    globus_cli
+
+[paths]
+source =
+    src
+    */site-packages
 
 [report]
 show_missing = True
 skip_covered = True
-fail_under = 87
+fail_under = 85
 
 exclude_lines =
     # the pragma to disable coverage
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -15,13 +15,46 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3
+      - name: Identify week number
+        shell: bash
+        run: |
+          date +'%V' > week-number.txt
+          date +'week-number=%V' >> $GITHUB_ENV
       - uses: actions/setup-python@v4
+        id: setup-python
         with:
           python-version: "3.10"
+          cache: "pip"
+          cache-dependency-path: |
+            .github/workflows/build.yaml
+            setup.cfg
+            setup.py
+            tox.ini
+            week-number.txt
+      - name: Restore cache
+        id: restore-cache
+        uses: actions/cache@v3
+        with:
+          path: |
+            .tox/
+            .venv/
+          key: >
+            lint
+            week=${{ env.week-number }}
+            os=${{ matrix.os }}
+            python=${{ steps.setup-python.outputs.python-version }}
+            hash=${{ hashFiles('.github/workflows/build.yaml', 'tox.ini', 'setup.cfg', 'setup.py') }}
+      - name: Identify venv path
+        shell: bash
+        run: echo 'venv-path=${{ runner.os == 'Windows' && '.venv/Scripts' || '.venv/bin' }}' >> $GITHUB_ENV
       - name: install tox
-        run: python -m pip install -U tox
+        if: steps.restore-cache.outputs.cache-hit == false
+        run: |
+          python -m venv .venv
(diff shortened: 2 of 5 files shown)


.coveragerc
```diff
@@ -1,11 +1,17 @@
 [run]
 parallel = True
-include = src/*
+source =
+    globus_cli
+
+[paths]
+source =
+    src
+    */site-packages
```
sirosen, line 9: I was just putting in a PR ( https://github.com/globus/globus-sdk-python/pull/670 ) which contained something similar, but more complex than this. I like this better; I'll test and switch to match.

.github/workflows/build.yaml
```diff
@@ -15,13 +15,46 @@ jobs:
...
+        with:
+          path: |
+            .tox/
+            .venv/
+          key: >
+            lint
+            week=${{ env.week-number }}
+            os=${{ matrix.os }}
+            python=${{ steps.setup-python.outputs.python-version }}
+            hash=${{ hashFiles('.github/workflows/build.yaml', 'tox.ini', 'setup.cfg', 'setup.py') }}
```
sirosen, line 46: Tangent: let's switch to declarative `setup.cfg` package metadata and make `setup.py` just a call to `setup()`, so we can remove `setup.py` from these file hash lists! 😉

.github/workflows/build.yaml
```diff
@@ -39,50 +72,170 @@ jobs:
...
+        uses: actions/cache@v3
+        with:
+          path: |
+            .tox
+            .venv
+          key: >
+            test
+            week=${{ env.week-number }}
+            os=${{ matrix.os }}
+            python=${{ steps.setup-python.outputs.python-version }}
```
sirosen, line 102: This applies to all jobs using this as a cache key, not just this one.

I think I saw another build using the python location from the env.
Two (conflicting) thoughts:
- I like this better, it reads more cleanly
- will this potentially result in corrupt/bad caches when github updates point releases (e.g. a 3.11.0 virtualenv running with a 3.11.1 interpreter)?

sirosen (commented):
I am always excited for faster CI!

There's one question I want to settle before approving, about the python version used as a cache key.

.github/workflows/build.yaml
sirosen: In another thread/context it came out that this version includes the patch version number, not just the input/requested version number. So I think I have no concern here. 👍

.github/workflows/build.yaml
```diff
@@ -15,13 +15,46 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3
+      - name: Identify week number
+        shell: bash
+        run: |
+          date +'%V' > week-number.txt
+          date +'week-number=%V' >> $GITHUB_ENV
```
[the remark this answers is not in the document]
kurtmckee: > Is it tied to the scheduled runs?

Yes, excellent observation! `%U` rolls the week numbers on Sundays. `%V` rolls the week numbers on Mondays. Since this codebase has a cron job for Mondays, I aligned with that and used `%V`.

The difference between repos exists because when we first discussed this, `%U` was the first solve and I didn't think anything of it until I started working on this repo.

sirosen (approved):

PR #754, opened 2023-02-07
Move user_message{,_link} into shared options

### LINKED ISSUE #476: Add user_message and user_message_link support to endpoint create/update
Field descriptions can be found in the transfer doc:
https://docs.globus.org/api/transfer/endpoint/#endpoint_fields

sirosen:
What's easier than creating a new issue in another issue tracker to move #476 over? Resolving #476.

I didn't bother about any new tests, since I think a combination of some quick manual tests and the type checking work we've done suffices to cover this change.

---

These parameters are now part of the shared create/update parameters for endpointish objects. They can be set on collections and host endpoints, but are noted in documentation as not being valid on shares. As such, they are not included in the options for `globus gcp create guest`.

In order to achieve the above semantics, the shared option decorators have gained a new ability, to accept a tuple of strings to denote options which should be omitted or "skipped".
The guest collection creation command passes

    skip=("user_message", "user_message_link")

and this is enough to signal that these two options are skipped. To do this, a mild refactor is applied, turning a list of decorators into a map of names to decorators. The final step, when they are applied, is a filter and conversion to a list.

closes #476

### DIFF
--- a/src/globus_cli/parsing/shared_options/endpointish.py
+++ b/src/globus_cli/parsing/shared_options/endpointish.py
@@ -158,7 +158,7 @@ def _apply_endpointish_create_or_update_params(
                 "--user-message",
                 help=(
                     "A message for clients to display to users when interacting "
-                    "with this {name}"
+                    f"with this {name}"
                 ),
                 type=StringOrNull(),
             ),
@@ -166,7 +166,7 @@ def _apply_endpointish_create_or_update_params(
                 "--user-message-link",
                 help=(
                     "Link to additional messaging for clients to display to users "
-                    "when interacting with this {name}. Should be an HTTP or HTTPS URL "
+                    f"when interacting with this {name}. Should be an HTTP or HTTPS URL "
                 ),
                 type=UrlOrNull(),
             ),
--- a/src/globus_cli/parsing/shared_options/endpointish.py
+++ b/src/globus_cli/parsing/shared_options/endpointish.py
@@ -166,7 +166,8 @@ def _apply_endpointish_create_or_update_params(
                 "--user-message-link",
                 help=(
                     "Link to additional messaging for clients to display to users "
-                    f"when interacting with this {name}. Should be an HTTP or HTTPS URL "
+                    f"when interacting with this {name}. "
+                    "Should be an HTTP or HTTPS URL "
                 ),
                 type=UrlOrNull(),
             ),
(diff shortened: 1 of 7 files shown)


src/globus_cli/parsing/shared_options/endpointish.py
```diff
@@ -89,64 +94,83 @@ def _apply_endpointish_create_or_update_params(
...
                 "--force-encryption/--no-force-encryption",
                 default=None,
                 help=f"Force the {name} to encrypt transfers",
             ),
-        ]
+            user_message=click.option(
+                "--user-message",
+                help=(
+                    "A message for clients to display to users when interacting "
+                    "with this {name}"
```
kurtmckee, line 161:
```suggestion
                    f"with this {name}"
```

src/globus_cli/parsing/shared_options/endpointish.py
```diff
@@ -89,64 +94,83 @@ def _apply_endpointish_create_or_update_params(
...
+                ),
+                type=StringOrNull(),
+            ),
+            user_message_link=click.option(
+                "--user-message-link",
+                help=(
+                    "Link to additional messaging for clients to display to users "
+                    "when interacting with this {name}. Should be an HTTP or HTTPS URL "
```
kurtmckee, line 169:
```suggestion
                    f"when interacting with this {name}. Should be an HTTP or HTTPS URL "
```

src/globus_cli/parsing/shared_options/endpointish.py
```diff
@@ -162,29 +186,25 @@ def _apply_endpointish_create_or_update_params(
                 " When set on mapped collections, this policy is inherited by any "
                 "guest collections"
```
kurtmckee, line 187: _Brain nudge: the whitespace in `verify_help` (here, where it's concatenated, and above in its definition) should be standardized at a later date._

kurtmckee (changes requested):

src/globus_cli/parsing/shared_options/endpointish.py
sirosen: Oy vey... I keep on making this mistake lately. I think I need to slow down or drink more coffee. ... Yeah, definitely more coffee. ☕☕☕

Thank you again for spotting these.

src/globus_cli/parsing/shared_options/endpointish.py
sirosen: At one point for gcs-cli, I wrote a series of tests for helptext because there were various tricky issues to run down. Between whitespace and my f-string typos, maybe we'd benefit from something similar here.

Perhaps a suite of tests which invokes `--help` on every command and asserts things about the output. I'll chew on this idea today.

kurtmckee (approved):

PR #755, opened 2023-02-07
Fix test response registration time

sirosen:
The registration of our response data can take a non-negligible amount of time (YAML parsing is slow and there are many files).

Unfortunately, structuring this as an autouse session fixture puts that within the timing for the execution of the first test. In concert with pytest-timeout being applied, this can in rare cases push a test over the timeout. (e.g. If run on a machine under heavy load.)

The issue can be dodged entirely by moving the registration of our response data into pytest_configure or pytest_sessionstart, which happen before any tests are running and being timed.

---

I got caught by timeouts twice in the last couple of days and I've never seen it happen before. I'm guessing this is partly to do with me simply having loaded up my dev box more than usual, but the fix is still a good improvement IMO.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -23,6 +23,12 @@ import globus_cli
 yaml = YAML()
 log = logging.getLogger(__name__)
 
+_test_file_dir = os.path.normpath(os.path.join(os.path.dirname(__file__), "files"))
+
+
+def pytest_configure(config):
+    _register_all_response_sets()
+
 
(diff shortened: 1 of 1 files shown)


ada-globus (approved):
This looks good! I marked it as "no-news-is-good-news" because this is a change that should only impact loading for tests, but feel free to change if you disagree :)

sirosen:
Indeed, I meant to mark it with that label! Thanks for doing that for me. Oops.

PR #758, opened 2023-02-11
Improve the formatting of failed test output

sirosen:
The `run_line` fixture tries to present all of the behavior of a command which fails in the form of an exception message. But the message formatting is currently hard to read, and it can easily be improved with some better whitespace. This doesn't impact any tests passing/failing, as the output only appears in the case of a failing test.

---

This was an improvement I made during other work, but which I think is worth reviewing and keeping on its own merits.
`run_line` might be due for a deeper rethink/refactor, but for now this makes it behave better.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,6 +4,7 @@ import logging
 import os
 import re
 import shlex
+import textwrap
 import time
 import uuid
 from unittest import mock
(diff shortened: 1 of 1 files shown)


kurtmckee (commented):
I feel like this might become more readable by using f-strings (rather than `.format()`), by creating the message prior to raising it (to reduce the parenthetical levels), and by using `textwrap.dedent()` so the actual text can stay in-line with the code. Perhaps something like:

```python
message = textwrap.dedent(f"""\
    ...
    stdout:
    {textwrap.indent(result.stdout, " " * 6}
    ...
    """)
raise Exception(message)
```

This introduces sensitivity to the current indentation level (that `6` has to accommodate the current indentation level *of the code*, plus whatever indentation you actually want in the output), but may read more easily. What do you think about that approach?

sirosen:
The `6` bothers me. IMO, aligning indents is nice when you can do it with a trivial `dedent`, but not important enough to make other significant compromises.

However, switching to an f-string definitely helps. I've made a compromise change to rewrite mostly as you suggested, but with the string dedented.

kurtmckee (approved):

PR #767, opened 2023-03-01
Annotate task commands and fix bugs in `globus task list` filters

sirosen:
Two commits here.

First, this annotates all of the `globus task` subcommands and adds them to the annotation tests while removing them from `mypy.ini`. This improves overall annotation coverage and quality.
In that work, I found a mismatch between `tuple` (produced by options with `multiple=True`) and `list` (being checked in a helper for `globus task list`. This was resolved by switching to one of their common bases, `Sequence`, which should fix some of the behavior.

The presence of the `tuple`/`list` bug revealed that there are not sufficient tests for `globus task list` commands. Adding those tests revealed at least one additional bug, in which the request_time and completion_time filters were incorrectly assembled.


sirosen:
As happens to me increasingly often, I tested on py3.11 and didn't bother to test the older pythons locally, thinking "of course I've done it right!". _Narrator: but he did not do it right._

I'm fixing it now and should have a version passing CI shortly.

kurtmckee (approved):

PR #780, opened 2023-03-22
fix: updates CLI Documentation link in the login_manager

jbottigliero:
Existing URL (https://globus.github.io/globus-cli/) does not resolve... as far as I know, https://docs.globus.org/cli/ is the official CLI documentation URL.


sirosen (approved):
Oh wow, I didn't even remember that we had a link in there.

I've marked this as not requiring a changelog (`no-news-is-good-news`) to speed our review/merge process.
Thanks for spotting this and putting in the fix!

PR #781, opened 2023-03-22
add base_python to mypy tox config

aaschaer:
`tox -e mypy` was failing when running with my default of py3.8. We discovered they also fail on 3.9, so this config forces 3.11 or 3.10.


sirosen (approved):
I'm pretty sure the CI failure is transient and will resolve on the rerun I just triggered. :crossed_fingers:

kurtmckee (approved):

PR #782, opened 2023-03-22
add --include option to globus transfer

aaschaer:
For https://app.shortcut.com/globus/story/22583/cli-add-support-for-include-filter-rules

Opening as draft because mypy will fail until https://github.com/globus/globus-sdk-python/pull/712 is released which is itself blocked on underlying transfer support being released.

After a reasonable amount of investigation on the team's part, there doesn't seem to be any better way to preserve ordering of options than to override `parse_args`.

It should be noted that this is technically a breaking change, as we are making `--exclude` only apply to files to support the `--include wanted_stuff.txt --exclude *` use case. Without this change, `--exclude *` would prevent traversing all directories, breaking recursion. I can get usage data from Transfer if we are worried about this.

### DIFF
--- a/changelog.d/20230321_115826_aaschaer_include.md
+++ b/changelog.d/20230321_115826_aaschaer_include.md
@@ -1,0 +1,4 @@
+### Enhancements
+
+* The `--exclude` option for `globus transfer` only applies to files.
+* Add `--include` option to `globus transfer` allowing ordered overrides of `--exclude` rules.
--- a/changelog.d/20230321_115826_aaschaer_include.md
+++ b/changelog.d/20230321_115826_aaschaer_include.md
@@ -1,4 +1,4 @@
 ### Enhancements
 
-* The `--exclude` option for `globus transfer` only applies to files.
+* Modify the `--exclude` option for `globus transfer` so it only applies to files.
 * Add `--include` option to `globus transfer` allowing ordered overrides of `--exclude` rules.
--- a/changelog.d/20230321_115826_aaschaer_include.md
+++ b/changelog.d/20230321_115826_aaschaer_include.md
@@ -1,4 +1,8 @@
 ### Enhancements
 
-* Modify the `--exclude` option for `globus transfer` so it only applies to files.
 * Add `--include` option to `globus transfer` allowing ordered overrides of `--exclude` rules.
+
+### Breaking Changes
+
+* The `--exclude` option for `globus transfer` now only applies to files to better
+  support excluding files within a directory structure
--- a/src/globus_cli/commands/transfer.py
+++ b/src/globus_cli/commands/transfer.py
@@ -33,6 +33,10 @@ else:
 
 @command(
     "transfer",
+    opts_to_combine={
+        "include": "filter_rules",
+        "exclude": "filter_rules",
+    },
     short_help="Submit a transfer task (asynchronous)",
     adoc_examples="""Transfer a single file:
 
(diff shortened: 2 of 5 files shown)


src/globus_cli/parsing/commands.py
```diff
@@ -74,7 +75,28 @@ def parse_args(self, ctx: click.Context, args: list[str]) -> list[str]:
...
+                combined_opts: dict[str, list[tuple[str, str]]] = {
+                    combined_name: [] for combined_name in self.opts_to_combine.values()
+                }
+                parser = self.make_parser(ctx)
+                values, _, order = parser.parse_args(args=list(args))
+                # values is a dict of value lists keyed by their option name
+                # in order for that value and opts is a list of option names
+                # in the order they were given at the command line
+                # we want a list of (name, value) tuples for multiple options
+                # in the order they were given at the command line
```
kurtmckee, line 89: > `...and opts is a list of option names`

This sentence suggests that a variable named "opts" may have been renamed to "order", which is now seen in the code. I think either the variable or the comment may need to be updated.

kurtmckee (changes requested):

changelog.d/20230321_115826_aaschaer_include.md
```diff
@@ -0,0 +1,4 @@
+### Enhancements
+
+* The `--exclude` option for `globus transfer` only applies to files.
```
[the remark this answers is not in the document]
aaschaer: I added a Breaking Changes section and a short explanation to why we're making the breaking change

src/globus_cli/commands/transfer.py
```diff
@@ -33,6 +33,10 @@
 
 @command(
     "transfer",
+    opts_to_combine={
+        "include": "filter_rules",
+        "exclude": "filter_rules",
+    },
```
sirosen, line 39: I like the interface you added here. It's relatively easy to read and guess what it does.

I wish we could attach this somehow to the options themselves, rather than the command, but I don't see a cheap and easy path to doing that. At best, we'd need custom option classes to store some extra values.

src/globus_cli/parsing/commands.py
```diff
@@ -74,7 +75,28 @@ def parse_args(self, ctx: click.Context, args: list[str]) -> list[str]:
         # args will be consumed, so check it before super()
         had_args = bool(args)
         try:
+            # if we have any opts to be combined in order, do that now
+            if self.opts_to_combine:
+                combined_opts: dict[str, list[tuple[str, str]]] = {
+                    combined_name: [] for combined_name in self.opts_to_combine.values()
+                }
+                parser = self.make_parser(ctx)
+                values, _, order = parser.parse_args(args=list(args))
```
sirosen, line 84: I think it would be valuable to note what the type of `parser` is here. I think it's an optparse parser?

sirosen (approved):
We've been well-served by the prior review here. Looks good!

Before merging, I'd like to see if we can add a comment to describe the internal parser object a little. It will make it easier to come back to this code in the future. That said, we can do without that if we want to release this shortly.

src/globus_cli/parsing/commands.py
aaschaer: > Before merging, I'd like to see if we can add a comment to describe the internal parser object a little. It will make it easier to come back to this code in the future. That said, we can do without that if we want to release this shortly.

@sirosen I added some explanation and example to the `GlobusCommand` help text, not sure that's exactly what you wanted?

PR #784, opened 2023-03-31
Add server timing header ASCII graph with hidden flag control

sirosen:
I started on this as a port of the version which was originally in globus-search-cli.
The main thing which is missing right now is some kind of tests, which is why it's currently just a draft.

I'm also not sure that it's hooked into the right place (e.g. for paginated methods), but this might be a good thing in most cases?
In particular, I'm looking at `globus api` commands' error behaviors, which use `click.echo` on the error body text rather than `display(...)`. Ideas about where and how to wire it up are welcome.

### DIFF
--- a/src/globus_cli/termio/server_timing.py
+++ b/src/globus_cli/termio/server_timing.py
@@ -143,7 +143,7 @@ def _parse_simple_metric_part(metric: str) -> Metric:
         value = float(unparsed_value)
     except ValueError as e:
         raise ServerTimingParseError("Metric value did not parse as float") from e
-    return Metric(name=name, duration=value)
+    return Metric(name=name.strip(), duration=value)
 
 
 DEFAULT_PARSER = Draft2017Parser()
--- a/src/globus_cli/termio/server_timing.py
+++ b/src/globus_cli/termio/server_timing.py
@@ -90,9 +90,7 @@ class ServerTimingParser(abc.ABC):
             try:
                 ret.append(self.parse_single_metric(item))
             except ServerTimingParseError:
-                if skip_errors:
-                    pass
-                else:
+                if not skip_errors:
                     raise
         return ret
 
@@ -122,33 +120,30 @@ class Draft2017Parser(ServerTimingParser):
     spec_reference = "https://www.w3.org/TR/2017/WD-server-timing-20171018/"
 
     def parse_single_metric(self, metric_str: str) -> Metric:
-        parts = [p.strip() for p in metric_str.split(";")]
-        if len(parts) == 1:
-            return _parse_simple_metric_part(parts[0])
-        elif len(parts) == 2:
-            metric = _parse_simple_metric_part(parts[0])
-            metric.description = parts[1].strip('"')
-            return metric
-        else:
+        part, *optionals = [p.strip() for p in metric_str.split(";")]
+        if len(optionals) > 1:
             raise ServerTimingParseError(
                 "Too many semicolons in timing item, cannot parse"
             )
+        metric = _parse_simple_metric_part(part)
+        if optionals:
+            metric.description = optionals[0].strip('"')
+        return metric
 
 
 def _parse_simple_metric_part(metric: str) -> Metric:
-    metric = metric.strip()
     if not metric:
         raise ServerTimingParseError("encountered empty metric")
 
-    if "=" in metric:
-        name, _, unparsed_value = metric.partition("=")
-        try:
-            value = float(unparsed_value)
-        except ValueError as e:
-            raise ServerTimingParseError("Metric value did not parse as float") from e
-        return Metric(name=name, duration=value)
-    else:
+    if "=" not in metric:
         return Metric(name=metric)
+
+    name, _, unparsed_value = metric.partition("=")
+    try:
+        value = float(unparsed_value)
+    except ValueError as e:
+        raise ServerTimingParseError("Metric value did not parse as float") from e
+    return Metric(name=name, duration=value)
 
 
 DEFAULT_PARSER = Draft2017Parser()
--- a/src/globus_cli/termio/server_timing.py
+++ b/src/globus_cli/termio/server_timing.py
@@ -1,7 +1,9 @@
 from __future__ import annotations
 
-import os
+import abc
+import dataclasses
 import shutil
+import typing as t
 
 import click
 import globus_sdk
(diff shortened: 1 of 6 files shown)


src/globus_cli/termio/server_timing.py
```diff
@@ -0,0 +1,92 @@
...
+    Given a Server Timing value as a string, parse it into a dict of the format
+      nice_name: value
+
+    For example
+
+      'a=1, "alpha"; b=2'
+
+    will parse as
+
+      {"alpha": 1, "b": 2}
```
kurtmckee, line 30: I'm confused by this example and the code below. It seems like the code below should split the string on semicolons so `"alpha"` is kept with `a=1`. However, it splits the overall string on comma, which would result in:

```python
[
  'a=1',
  '"alpha"; b=2',
]
```

I was anticipating this to be the result of line 51:

```python
[
  'a=1, "alpha"',
  'b=2',
]
```

sirosen: Oh. Huh. The code does not match this doc. Confusing indeed!

The Server-Timing header spec lets you use "nice names" which are quoted strings after the key-value pairs. (I have little idea why the spec is written this way, since it seems like the nice names aren't that important, but it doesn't bother me much.)

A good example might be
```
x0=50.2, "Authentication Callout"; x1=1.8, "Database Read"

=>

Authentication Callout: 50.2ms
Database Read: 1.8ms
```

I'm guessing that at one point I meant to use the nice names and then later I decided to ignore them.
But even if that's the output we want (and I think we should prefer the nice names, per the example above?), the parse should still start with semicolons.

sirosen: Er... I went to double-check the spec and it seems that I'm a bit wrong about what's allowed and supported.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server-Timing

:crossed_fingers: that our Server-Timing header builders in our applications are correct. I think they are -- I was looking much more at the spec when I wrote them.
But this all needs a bit of reassessment.

sirosen: I've started to dig deeper. It seems like the Server-Timing spec has changed in the time that I've been using it.
And it's continuing to change:
https://www.w3.org/standards/history/server-timing

Here's one of the older drafts which looks similar to what I remembered:
https://www.w3.org/TR/2017/WD-server-timing-20170725/

For right now, I'm writing down a method, `parse_metric_draft_2017` which declares itself as matching that draft.
I haven't figured out how we'll detect which draft is in use though, or if we will even want to do so.

The idea of writing a library with full detection and support for all of these drafts is simultaneously fun and not where I think we want to be spending our time. :sweat_smile:

sirosen: I just finished fairly thoroughly rewriting the parsing code. It should be a bit more readable (I hope) and makes it possible to support multiple spec versions in the future.

I'm going to resolve this because the relevant docstring was moved and rewritten, but the new parsing behavior would definitely benefit from a careful review. I did test it against live services and saw good results, however.

src/globus_cli/termio/server_timing.py
```diff
@@ -0,0 +1,154 @@
...
+        metric_items = header_str.split(",")
+        ret: list[Metric] = []
+        for item in metric_items:
+            try:
+                ret.append(self.parse_single_metric(item))
+            except ServerTimingParseError:
+                if skip_errors:
+                    pass
+                else:
+                    raise
```
kurtmckee, line 96: I think this can be simplified:

```python
        for item in metric_items:
            try:
                ret.append(self.parse_single_metric(item))
            except ServerTimingParseError:
                if not skip_errors:
                    raise
```

FUN ALTERNATIVE! DO NOT RECOMMEND!

```python
        for item in metric_items:
            try:
                ret.append(self.parse_single_metric(item))
            finally:
                if skip_errors:
                    continue
```

src/globus_cli/termio/server_timing.py
```diff
@@ -0,0 +1,154 @@
...
+        if len(parts) == 1:
+            return _parse_simple_metric_part(parts[0])
+        elif len(parts) == 2:
+            metric = _parse_simple_metric_part(parts[0])
+            metric.description = parts[1].strip('"')
+            return metric
+        else:
+            raise ServerTimingParseError(
+                "Too many semicolons in timing item, cannot parse"
+            )
```
kurtmckee, line 135: I prefer for the exception case to appear first, and to only call `_parse_simple_metric_part()` once.

```suggestion
    def parse_single_metric(self, metric_str: str) -> Metric:
        part, *optionals = [p.strip() for p in metric_str.split(";")]
        if len(optionals) > 1:
            raise ServerTimingParseError(
                "Too many semicolons in timing item, cannot parse"
            )
        metric = _parse_simple_metric_part(part)
        if optionals:
            metric.description = optionals[0].strip('"')
        return metric
```

src/globus_cli/termio/server_timing.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+def _parse_simple_metric_part(metric: str) -> Metric:
+    metric = metric.strip()
```
kurtmckee, line 139: It looks like the string is already stripped before it's passed in; is there another route by which this might be called, or is this being thorough?

```suggestion
```

src/globus_cli/termio/server_timing.py
```diff
@@ -0,0 +1,154 @@
...
+
+    if "=" in metric:
+        name, _, unparsed_value = metric.partition("=")
+        try:
+            value = float(unparsed_value)
+        except ValueError as e:
+            raise ServerTimingParseError("Metric value did not parse as float") from e
+        return Metric(name=name, duration=value)
+    else:
+        return Metric(name=metric)
```
kurtmckee, line 151: I prefer the simple case to appear first, and then dedent the rest of the block:

```suggestion
    if "=" not in metric:
        return Metric(name=metric)

    name, _, unparsed_value = metric.partition("=")
    try:
        value = float(unparsed_value)
    except ValueError as e:
        raise ServerTimingParseError("Metric value did not parse as float") from e
    return Metric(name=name, duration=value)
```

tests/unit/test_server_timing.py
```diff
@@ -0,0 +1,97 @@
...
+    (
+        (
+            "a=1, b=2.2",
+            [
+                Metric(name="a", duration=1.0),
+                Metric(name="b", duration=2.2),
+            ],
+        ),
+        (
+            'a=1, b, c=2.2; "callout"',
```
kurtmckee, line 55: I haven't seen a case that exercises the `.strip()` calls. Perhaps:

```python
' a=1 , b , c=2.2 ; "callout" '
```

(It might be worthwhile to test spaces on either side of the equals sign...don't know if the spec allows for that, though.)

kurtmckee (changes requested):
I think my feedback might have gone down a self-challenge to code golf this a little...sorry!

The only thing I'm actually blocking on is a test that exercises the `.strip()` calls!

tests/unit/test_server_timing.py
sirosen: I decided to make the test slightly more stringent and put in ` a = 1 `, which then promptly failed to parse as I wanted ("a " vs "a").
So the test change also includes _another_ call to `strip()` to get that one to come out as `"a"`.

That said, overall I feel good about where the tests put us on this front.

kurtmckee (approved):
Nice!

PR #788, opened 2023-04-26
Add `globus flows create` command

ada-globus:
# Context
Implements [[sc-18412](https://app.shortcut.com/globus/story/18412/implement-initial-version-of-globus-flows-create)].

This adds the `globus flows create` command.

# Notes
- The title and definition are required arguments and thus positional. All others are optional. (The input schema is presumed to be `{}` if omitted).
- The definition and input schema support the JSONStringOrFile parameter and can accept a path to a JSON file (prefixed with `file:`) or a valid JSON string.
- I had initially pursued work on a principal ParamType, but ended up deciding based on conversation (where we agreed that we prefer that the service provides this feedback) that this was best set aside to keep the scope of this work constrained.
- There is a gap here where the SDK doesn't handle Flows's error formats. As a prerequisite, we'll need upstream extension of the FlowsAPIError in order for the CLI to handle this (subsequently, a custom hook for this error type may be added to the CLI). I've considered this non-blocking for this change.
- I tried to keep the test definition somewhat generic so that it shouldn't take too much work if we add more samples to responses. This will require a little work to fully genericize it, but I believe should be mostly around creating a proper list of the expected principals.
- There is an unreleased change in the SDK that will allow us to remove the overridden values on L82-85 and test distinct identities for these fields. I've added these lines for now to allow this to move forward.

### DIFF
--- a/src/globus_cli/commands/flows/create.py
+++ b/src/globus_cli/commands/flows/create.py
@@ -5,6 +5,7 @@ import click
 from globus_cli.login_manager import LoginManager
 from globus_cli.parsing import JSONStringOrFile, ParsedJSONData, command
 from globus_cli.termio import Field, TextMode, display, formatters
+from globus_cli.types import JsonValue
 
 ROLE_TYPES = ("flow_viewer", "flow_starter", "flow_administrator", "flow_owner")
 
@@ -140,13 +141,27 @@ def create_command(
             globus flows create 'My Other Flow' definition.json
     """
 
+    # Ensure that the definition is a JSON object
+    if not isinstance(definition.data, dict):
+        raise click.UsageError("Flow definition must be a JSON object")
+    definition_doc = definition.data
+
+    # Ensure the input schema is a JSON object
+    if input_schema is None:
+        input_schema_doc: dict[str, JsonValue] = {}
+    else:
+        if not isinstance(input_schema.data, dict):
+            raise click.UsageError("--input-schema must be a JSON object")
+        input_schema_doc = input_schema.data
+
+    # Configure clients
     flows_client = login_manager.get_flows_client()
     auth_client = login_manager.get_auth_client()
 
     res = flows_client.create_flow(
         title=title,
-        definition=definition.data,
-        input_schema=input_schema.data if input_schema is not None else {},
+        definition=definition_doc,
+        input_schema=input_schema_doc,
         subtitle=subtitle,
         description=description,
         flow_viewers=list(viewers),
@@ -155,6 +170,7 @@ def create_command(
         keywords=list(keywords),
     )
 
+    # Configure formatters for principals
     principal_formatter = formatters.auth.PrincipalURNFormatter(auth_client)
     for principal_set_name in ("flow_administrators", "flow_viewers", "flow_starters"):
         for value in res.get(principal_set_name, ()):
--- a/src/globus_cli/commands/flows/create.py
+++ b/src/globus_cli/commands/flows/create.py
@@ -35,12 +35,12 @@ ROLE_TYPES = ("flow_viewer", "flow_starter", "flow_administrator", "flow_owner")
         Example: Inline JSON:
 
         \b
-            --input_schema '{"properties": {"src": {"type": "string"}}}'
+            --input-schema '{"properties": {"src": {"type": "string"}}}'
 
         Example: Path to JSON file:
 
         \b
-            --input_schema file:schema.json
+            --input-schema file:schema.json
 
         If unspecified, the default is an empty JSON object ('{}').
     """,
--- a/src/globus_cli/commands/flows/create.py
+++ b/src/globus_cli/commands/flows/create.py
@@ -76,7 +76,7 @@ ROLE_TYPES = ("flow_viewer", "flow_starter", "flow_administrator", "flow_owner")
     help="""
         A principal that may run the flow.
 
-        Use "all_authenticated_users" to make the flow visible to everyone.
+        Use "all_authenticated_users" to allow any authenticated user to run the flow.
 
         This option can be specified multiple times
         to create a list of flow starters.
@@ -90,7 +90,7 @@ ROLE_TYPES = ("flow_viewer", "flow_starter", "flow_administrator", "flow_owner")
     help="""
         A principal that may view the flow.
 
-        Use "public" to allow any authenticated user to run the flow.
+        Use "public" to make the flow visible to everyone.
 
         This option can be specified multiple times
         to create a list of flow viewers.
--- a/src/globus_cli/commands/flows/create.py
+++ b/src/globus_cli/commands/flows/create.py
@@ -29,8 +29,7 @@ ROLE_TYPES = ("flow_viewer", "flow_starter", "flow_administrator", "flow_owner")
         The JSON input schema that governs the parameters
         used to start the flow.
 
-        The input document may be specified inline,
-        or it may be a path to a JSON file, prefixed with "file:".
+        The input document may be specified inline, or it may be a path to a JSON file.
 
         Example: Inline JSON:
 
@@ -40,7 +39,7 @@ ROLE_TYPES = ("flow_viewer", "flow_starter", "flow_administrator", "flow_owner")
         Example: Path to JSON file:
 
         \b
-            --input-schema file:schema.json
+            --input-schema schema.json
 
         If unspecified, the default is an empty JSON object ('{}').
     """,
@@ -128,8 +127,8 @@ def create_command(
     TITLE is the name of the flow.
 
     DEFINITION is the JSON document that defines the flow's instructions.
-    The definition document may be specified inline,
-    or it may be a path to a JSON file, prefixed with "file:".
+    The definition document may be specified inline, or it may be
+    a path to a JSON file.
 
         Example: Inline JSON:
 
@@ -140,7 +139,7 @@ def create_command(
         Example: Path to JSON file:
 
         \b
-            globus flows create 'My Other Flow' file:definition.json
+            globus flows create 'My Other Flow' definition.json
     """
 
     if input_schema is None:
--- a/src/globus_cli/commands/flows/create.py
+++ b/src/globus_cli/commands/flows/create.py
@@ -72,9 +72,10 @@ ROLE_TYPES = ("flow_viewer", "flow_starter", "flow_administrator", "flow_owner")
     type=str,
     multiple=True,
     help="""
-        A principal that may run the flow.
+        A principal that may start a new run of the flow.
 
-        Use "all_authenticated_users" to allow any authenticated user to run the flow.
+        Use "all_authenticated_users" to allow any authenticated user
+        to start a new run of the flow.
 
         This option can be specified multiple times
         to create a list of flow starters.
--- a/src/globus_cli/commands/flows/create.py
+++ b/src/globus_cli/commands/flows/create.py
@@ -1,0 +1,196 @@
+from __future__ import annotations
+
+import typing as t
+
+import click
+
+from globus_cli.login_manager import LoginManager
+from globus_cli.parsing import JSONStringOrFile, command
+from globus_cli.termio import Field, TextMode, display, formatters
+
+ROLE_TYPES = ("flow_viewer", "flow_starter", "flow_administrator", "flow_owner")
+
+
+@command("create", short_help="Create a flow")
+@click.argument(
+    "title",
+    type=str,
+)
+@click.argument(
+    "definition",
+    type=JSONStringOrFile(),
+    metavar="DEFINITION",
+)
+@click.option(
+    "--input-schema",
+    "input_schema",
+    type=JSONStringOrFile(),
+    help="""
+        The JSON input schema that governs the parameters
+        used to start the flow.
+
+        The input document may be specified inline,
+        or it may be a path to a JSON file, prefixed with "file:".
+
+        Example: Inline JSON:
+
+        \b
+            --input_schema '{"properties": {"src": {"type": "string"}}}'
+
+        Example: Path to JSON file:
+
+        \b
+            --input_schema file:schema.json
+
+        If unspecified, the default is an empty JSON object ('{}').
+    """,
+)
+@click.option(
+    "--subtitle",
+    type=str,
+    help="A concise summary of the flow's purpose.",
+)
+@click.option(
+    "--description",
+    type=str,
+    help="A detailed description of the flow's purpose.",
+)
+@click.option(
+    "--administrator",
+    "administrators",
+    type=str,
+    multiple=True,
+    help="""
+        A principal that may perform administrative operations
+        on the flow (e.g., update, delete).
+
+        This option can be specified multiple times
+        to create a list of flow administrators.
+    """,
+)
+@click.option(
+    "--starter",
+    "starters",
+    type=str,
+    multiple=True,
+    help="""
+        A principal that may run the flow.
+
+        Use "all_authenticated_users" to make the flow visible to everyone.
+
+        This option can be specified multiple times
+        to create a list of flow starters.
+    """,
+)
+@click.option(
+    "--viewer",
+    "viewers",
+    type=str,
+    multiple=True,
+    help="""
+        A principal that may view the flow.
+
+        Use "public" to allow any authenticated user to run the flow.
+
+        This option can be specified multiple times
+        to create a list of flow viewers.
+    """,
+)
+@click.option(
+    "--keyword",
+    "keywords",
+    type=str,
+    multiple=True,
+    help="""
+        A term used to help discover this flow when
+        browsing and searching.
+
+        This option can be specified multiple times
+        to create a list of keywords.
+    """,
+)
+@LoginManager.requires_login("flows")
+def create_command(
+    login_manager: LoginManager,
+    title: str,
+    definition: dict,
+    input_schema: dict | None | t.Any,
+    subtitle: str | None,
+    description: str | None,
+    administrators: tuple[str],
+    starters: tuple[str],
+    viewers: tuple[str],
+    keywords: tuple[str],
+) -> None:
+    """
+    Create a new flow.
+
+    TITLE is the name of the flow.
+
+    DEFINITION is the JSON document that defines the flow's instructions.
+    The definition document may be specified inline,
+    or it may be a path to a JSON file, prefixed with "file:".
+
+        Example: Inline JSON:
+
+        \b
+            globus flows create 'My Cool Flow' \\
+            '{{"StartAt": "a", "States": {{"a": {{"Type": "Pass", "End": true}}}}}}'
+
+        Example: Path to JSON file:
+
+        \b
+            globus flows create 'My Other Flow' file:definition.json
+    """
+
+    if input_schema is None:
+        input_schema = {}
+
+    flows_client = login_manager.get_flows_client()
+    auth_client = login_manager.get_auth_client()
+
+    res = flows_client.create_flow(
+        title=title,
+        definition=definition,
+        input_schema=input_schema,
+        subtitle=subtitle,
+        description=description,
+        flow_viewers=list(viewers),
+        flow_starters=list(starters),
+        flow_administrators=list(administrators),
+        keywords=list(keywords),
+    )
+
+    principal_formatter = formatters.auth.PrincipalURNFormatter(auth_client)
+    for principal_set_name in ("flow_administrators", "flow_viewers", "flow_starters"):
+        for value in res.get(principal_set_name, ()):
+            principal_formatter.add_item(value)
+    principal_formatter.add_item(res.get("flow_owner"))
+
+    fields = [
+        Field("Flow ID", "id"),
+        Field("Title", "title"),
+        Field("Subtitle", "subtitle"),
+        Field("Description", "description"),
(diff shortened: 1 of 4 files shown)


src/globus_cli/commands/flows/create.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+@command("create", short_help="Create a flow")
+@click.argument(
+    "title",
+    type=str,
+)
+@click.argument(
+    "definition",
+    type=JSONStringOrFile(),
```
kurtmckee, line 21: The more I think about this, the less I like the requirement to specify `file:` as a prefix to the filename. I'd prefer to lean away from that syntax and instead assume that if the file exists, it's a file; and if it doesn't exist, fall back to loading the text as JSON.

I'm also leery of tab completion getting broken by mandating a `file:` prefix.

kurtmckee (changes requested):

src/globus_cli/commands/flows/create.py
sirosen: With time, I've also started to wonder if we can redo the json-or-file parsing. I originally introduced it in gcs-cli because we had other things which needed a similar parsing strategy, and then I carried it back to globus-cli to keep things consistent.

I'd like to talk about this today and see if we can write something like... `JSONStringOrFileV2` which has behavior more to our liking.

sirosen: I'm hoping that I have this resolved in a way which we will like in #790.
I'd like us to hold this PR until that can get review, as the resolution here may be (:crossed_fingers:) a simple rebase.

src/globus_cli/commands/flows/create.py
```diff
@@ -0,0 +1,196 @@
...
+
+        Example: Inline JSON:
+
+        \b
+            --input_schema '{"properties": {"src": {"type": "string"}}}'
+
+        Example: Path to JSON file:
+
+        \b
+            --input_schema file:schema.json
```
sirosen, line 43: _Assuming_ we take the JSONStringOrFile rewrite, I'd change this to `--input-schema schema.json` to bias us away from the `file:` prefix syntax.

I also notice `--input_schema` here and above, which need minor correction.

src/globus_cli/commands/flows/create.py
```diff
@@ -0,0 +1,196 @@
...
+    type=str,
+    multiple=True,
+    help="""
+        A term used to help discover this flow when
+        browsing and searching.
+
+        This option can be specified multiple times
+        to create a list of keywords.
+    """,
+)
```
sirosen, line 111: Rather than allowing `--keyword foo --keyword bar`, would a better CLI usage be `--keywords foo,bar`?
I think comma-delimited lists are nicer for most users, even if it forbids commas inside of keywords.

src/globus_cli/commands/flows/create.py
```diff
@@ -0,0 +1,196 @@
...
+
+    res = flows_client.create_flow(
+        title=title,
+        definition=definition,
+        input_schema=input_schema,
+        subtitle=subtitle,
+        description=description,
+        flow_viewers=list(viewers),
+        flow_starters=list(starters),
+        flow_administrators=list(administrators),
```
sirosen, line 160: The current behavior here will be to pass these inputs verbatim, which means that users won't be able to pass usernames or unadorned UUIDs (which we could then decorate as URNs).

I could see an argument against accepting bare UUIDs, but I'd like us to accept usernames. They're much more user-friendly.

What about doing the following to each of these?
```python
new = []
for p in principal_list:
    if p.startswith("urn:"):
        new.append(p)
        continue
    resolved = auth_client.maybe_lookup_identity_id(p)
    if resolved:
        new.append(f"urn:globus:auth:identity:id:{resolved}"
    else:
        raise click.UsageError("bad, no such user")
```

Note that the way `maybe_lookup_identity_id` is written, it will treat a UUID with a no-op. We could adjust that for this case, or explicitly check for (and reject) UUIDs.

tests/functional/flows/test_create_flow.py
```diff
@@ -0,0 +1,170 @@
...
+    },
+    "monica": {
+        "username": "monica@kreb.scouts",
+        "name": "Monica Perling",
+        "identity_provider": "c8abac57-560c-46c8-b386-f116ed8793d5",
+        "organization": "Kreb Scouts",
+        "status": "used",
+        "email": "monica@kreb.scouts",
+    },
+}
```
sirosen, line 39: I had to look these up and it was time well-spent. :grin:

src/globus_cli/commands/flows/create.py
kurtmckee: Totally missed the underscore-vs-hyphen. Good catch!

src/globus_cli/commands/flows/create.py
kurtmckee: I think that allowing CSVs would break with the existing idea that, for example, `--administrator` must be specified multiple times.

src/globus_cli/commands/flows/create.py
ada-globus: I'm pretty sure I am a 'no' on bare UUIDs, but I think—though I was originally leaning against it— I've come around on attempting to resolve usernames. I'll update to address.

src/globus_cli/commands/flows/create.py
```diff
@@ -0,0 +1,196 @@
...
+)
+@click.option(
+    "--starter",
+    "starters",
+    type=str,
+    multiple=True,
+    help="""
+        A principal that may run the flow.
+
+        Use "all_authenticated_users" to make the flow visible to everyone.
```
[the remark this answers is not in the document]
ada-globus: Thanks for catching! I think I had the arguments reversed originally and forgot to fix the description 🤦‍♀️

src/globus_cli/commands/flows/create.py
ada-globus: I have a very soft lean toward multiple instances of the argument, but if there is substantial prior art for comma-delimited lists in the CLI then I'm willing to be persuaded that it's better to keep consistency at a CLI-level.

src/globus_cli/commands/flows/create.py
ada-globus: As I am starting to work on this, a couple things that I'm realizing:

a) this kind of contradicts the prior conversation we had as a group about wanting to have the validation of these values happen at the API-level so that it's controlled by the service. I'd originally had written a principal ParamType that did stricter validation here of the values along with nice user messaging, but I felt like there was some debate over whether we shouldn't just be passing more or less anything the user inputs here and let the CLI sort it out, so I put that on ice in order to move this forward.

b) this gets uglier when we start to add handling for the special principals

I'm up for talking through how we want to address, but would sort of like to stet this for the scope of this PR so we can decide holistically how we want to handle this sort of situation. If we want to go this route (and I still think I could be convinced), I'm inclined to resurrect the principal ParamType and add the username resolution to it, but I think that will make this PR quite a bit larger.

Thoughts?

ada-globus:
Initial guess is that the CI failure is due to the change I made to the `TWO_HOP_TRANSFER_FLOW_DEFINITION` over in the SDK which has now landed in `main`. If correct, we can do an SDK release and I can just rip out the temporary kludge that makes it work with the currently released SDK, _or_ I can revise the kludge to detect which to use. (This genre of conditional in a test makes me vaguely uneasy but it seems pretty okay in this particular case, esp as it will likely be short-lived). I'm going to let it rest for a minute since we have other changes going on here in the interim.

src/globus_cli/commands/flows/create.py
sirosen: We allow for both comma-delimited lists and multi-option usages. To summarize my views on the two modes of use in two points:

1. Multi-option use is very explicit, allows for more robust and complex parsing, but is also more onerous for users. Comma-delimited lists are very easy and lightweight, but more limited.
2. They're two equally good APIs for users, and should be tailored to specific use-cases, not one-size-fits-all.

GCS and Transfer both treat keywords as a comma-delimited lists, which is worth keeping on our radar. That's partly driven by a service behavior, in that Transfer actually requires keywords for endpoint search to be provided as a string field, with comma-delimited semantics service-side.

<details><summary>GCS tangent</summary>
GCS is a little bit inbetween, accepting a list of strings, but then needing to serialize that and send it to Transfer (to reflect the data there) as a comma-delimited list. So even though GCS might accept `["foo,bar", "baz"]`, it can't mean anything different from `"foo,bar,baz"`.
</details>

Whether or not cross-service consistency is a goal (I would vaguely like it but also don't think it's terribly realistic to expect it at present), I think the meaning of keywords for search and retrieval is better aligned with the more lightweight comma-delimited usage.
Regarding the other flavor of consistency -- that within a command or API -- I'm not sure that it makes sense to align the mode of use for `--keyword` with `--administrator` (to stick with that example) because they're two different kinds of data.

kurtmckee: > GCS and Transfer both treat keywords as a comma-delimited lists,

and

> I'm not sure that it makes sense to align the mode of use for `--keyword` with `--administrator` (to stick with that example) because they're two different kinds of data.

Flows accepts a list of strings, not a CSV. I don't currently think it's onerous to use multiple keywords.

src/globus_cli/commands/flows/create.py
sirosen: I'm pretty clear on my long-term vision here. I'm not sure what we should do in the short term.

### Long Term

We already have a principal type in the CLI:
https://github.com/globus/globus-cli/blob/0791f166299a77ef2f00f889bf8ea88e26c6ffa6/src/globus_cli/parsing/param_types/identity_type.py#L36

It's just underpowered for what we need for Flows. And probably misnamed. And maybe the domain-handling case should be a special subclass. And... :sweat_smile: 

I'd like to fix this up into something which supports URN and non-URN syntax, can take options to enable or disable known special principals like `public`, etc, etc. All of that work is well beyond the scope of this PR.

There should be a type which handles all forms of Principals and which can be tuned and parametrized, or subclassed, as appropriate. We should gracefully handle errors from the services and have user-friendly support in the CLI, meaning that if the CLI is ever wrong, users experience a clear error.

I think this long term plan is somewhat uncontroversial -- it's a CLI which "does the right thing" in as many cases as possible.

Regarding the apparent conflict:
> this kind of contradicts the prior conversation we had as a group about wanting to have the validation of these values happen at the API-level so that it's controlled by the service.

I think my attitude is that a CLI which can do fancy things is very nice, and we should work on that. But a CLI which can't handle API errors correctly is broken and needs to be fixed.

I like the CLI to do minimal transformations to the data it receives, and trust service-side validation to work and to be rendered correctly. Drawing the line on "minimal transformations" is less science than art, but I think the error-handling behavior -- and the expectation that it's a rock-solid piece of the API-SDK-CLI interaction which we can rely upon -- is pretty straightforward.

### Short Term

We can drop this from the create flow command, but I think it needs to be a fast-follow.
URN-formatted principals are not particularly friendly as a CLI user experience, and I think users will find it difficult to use if we don't make accommodations.

For a user who isn't already writing data in the URN format, I think we need to look at what their experience will be if they want to share with a given user by username, e.g. `sirosen@globusid.org`. How do you convert that into an input for this command?

> Why, simply run with `--administrator "urn:globus:auth:identities:id:$(globus get-identities sirosen@globusid.org)"`!

That's not a user experience I want to advertise very much. It works, but it's not very easy or even very explainable.

Dropping any identity resolution from this PR might be the best choice, but I think it's important to solve this problem. Not so important that I'd object to doing a release without it though, so I'd be okay merging without it.

The main alternative I see is to define an interface here (no global / generic param type, just special-cased stuff for this command) which we want the long-term thing to conform to. That gives us tighter parameters on what the PrincipalType param type will have to support, but it also lets us more rapidly deliver a version of this command which supports our desired interface.

---

I guess I'm ping-pong-ing this back to you. How do you want to proceed? Intentionally leave it out, but agree that we want it handled in the near term? Or do a one-off here and let that guide how we want our generic solution to work?

ada-globus: Per our discussion yesterday, I think we're in alignment on the eventual goal. I think that the thing we want to create for this is complex enough that it should be explicitly outside of this PR. I've created a story that attempts to sketch it out here: [[SC-24057](https://app.shortcut.com/globus/story/24057)].

I already have some initial work done toward this that I pulled out of this PR in an earlier phase and some general intuition about the shape this could take, and I'm happy to unstash into a new branch and continue that work once we've refined the story enough to be sure we all agree on what we want this to look like (such that it is of the greatest value possible for other CLI commands, both currently existing and in the future).

IMO, ideally this should support the functionality of the existing IdentityType enabling it to either replace IdentityType or allow IdentityType to become a lightweight subclass (I don't know if we want the b32 or domain stuff in the PrincipalType class itself...?).

sirosen: > (I don't know if we want the b32 or domain stuff in the PrincipalType class itself...?).

Yeah, this is a perfect example of "special weird extension", I think. Either it's a subclass or it's some opt-in behavior for the cases which want it.
I see the support for those being useful only for Transfer (where the b32-encoded IDs are used) and Auth (where you might want to cross-reference said data).

Let's consider this thread resolved. We're clear enough on what needs to happen that it shouldn't block this PR.

ada-globus:
I believe this is in good shape now and has been reconciled with the changes to `JSONStringOrFile`. In order to resolve the discrepancy between the test data in the last release of the SDK and the new test data in `main`, I initially created a temporary conditional to reshape the expectations but ...then changed course to move ahead with genericizing the test (I _think_ should now work on any reasonably shaped input data and properly handles things like repeated identities across roles, etc) which will allow us to easily parametrize this when additional responses are available.

tests/functional/flows/test_create_flow.py
```diff
@@ -0,0 +1,220 @@
...
+        existing_identities=identities,
+    )
+    flow_starter_identities = assign_identities_for_principal_set(
+        principal_set=flow_starters,
+        existing_identities=identities,
+    )
+    flow_viewer_identities = assign_identities_for_principal_set(
+        principal_set=flow_viewers,
+        existing_identities=identities,
+    )
```
sirosen, line 126: I found this stack of calls hard to follow at first because it's doing two data manipulations at once -- returning a list and updating the `identities` dict.

I think it might be easier to follow with a very small object to wrap it, like...

```python
class IdentityPool:
    IDENTITY_DATA = {"pete": ..., "monica": ..., ...}

    def __init__(self):
        self.identities = {}
        self.assigned_sets = {}

    def assign(self, set_name, principal_set):
        ...  # see usage below to guess at what this should do

...

pool = IdentityPool()
pool.assign("owner", [response.json["flow_owner"]])
pool.assign("administrator", response.json["flow_administrators"])
...
owner_identity = pool.assigned_sets["owner"][0]
flow_administrator_identities = pool.assigned_sets["administrator"]
...

load_response(
    RegisteredResponse(
        service="auth",
        path="/v2/api/identities",
        json={
            "identities": list(pool.identities.values()),
        },
    )
)
```

This is maybe a bit more prescriptive than I mean to be, but I'm just looking for ways to avoid having to juggle quite as many details as a reader. Splitting up the assignment from reading back the assigned values is the first thing which occurs to me as a way of doing that.

tests/functional/flows/test_create_flow.py
```diff
@@ -0,0 +1,220 @@
...
+    for keyword in keywords:
+        arguments.extend(("--keyword", f"'{keyword}'"))
+    if input_schema is not None:
+        arguments.extend(("--input-schema", f"'{json.dumps(input_schema)}'"))
+    if subtitle is not None:
+        arguments.extend(("--subtitle", f"'{subtitle}'"))
+    if description is not None:
+        arguments.extend(("--description", f"'{description}'"))
+
+    result = run_line(f"globus flows create {' '.join(arguments)}")
```
sirosen, line 159: I see a lot of careful quoting here, and I think this is a good case for a refactor to use the non-shlex-splitting API for a click CliRunner. I'm going to suggest a refactor here, but it's untested so please check before assuming it's good. :smile: 

```suggestion
    command = [
        "globus",
        "flows",
        "create",
        title,
        json.dumps(definition),
    ]
    for flow_administrator in flow_administrators:
        command.extend(("--administrator", {flow_administrator}))
    for flow_starter in flow_starters:
        command.extend(("--starter", flow_starter))
    for flow_viewer in flow_viewers:
        command.extend(("--viewer", flow_viewer))
    for keyword in keywords:
        command.extend(("--keyword", keyword))
    if input_schema is not None:
        command.extend(("--input-schema", json.dumps(input_schema)))
    if subtitle is not None:
        command.extend(("--subtitle", subtitle))
    if description is not None:
        command.extend(("--description", description))

    result = run_line(command)
```

tests/functional/flows/test_create_flow.py
```diff
@@ -0,0 +1,220 @@
...
+    return identity_set
+
+
+def value_for_field_from_output(name, output):
+    """
+    Return the value for a specified field from the output of a command.
+    """
+    match = re.search(rf"^{name}:[^\S\n\r]+(?P<value>.*)$", output, flags=re.M)
+    assert match is not None
+    return match.group("value")
```
sirosen, line 93: I like this helper and I'm thinking about how we can make this into a fixture which we can expose more globally within the testsuite. I think I did something like this at some point in the distant past, but not with a good interface. So this is a problem I know we keep seeing and solving per-test, and I want to put it into our hands more consistently.

sirosen (commented):
This is very, very close. I have some optional changes in feedback, but I want us to discuss and resolve those threads before I approve.

tests/functional/flows/test_create_flow.py
```diff
@@ -0,0 +1,220 @@
...
+    )
+    flow_starter_identities = assign_identities_for_principal_set(
+        principal_set=flow_starters,
+        existing_identities=identities,
+    )
+    flow_viewer_identities = assign_identities_for_principal_set(
+        principal_set=flow_viewers,
+        existing_identities=identities,
+    )
+    print(identities)
```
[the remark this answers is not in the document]
ada-globus: I do, too—I simply missed this one! Thanks :)

tests/functional/flows/test_create_flow.py
ada-globus: I didn't know this was possible—perfect 👍

tests/functional/flows/test_create_flow.py
ada-globus: Yeah, this seems reasonable to me. I'll take a quick swing at it.

ada-globus: I've done something much along the lines of this suggestion that I think improves readability meaningfully. Could potentially improve further w/ more investment in the IdentityPool class, but have refrained from doing so unless/until we think it might prove useful for other tests. LMK what you think!

sirosen (approved):

kurtmckee (changes requested):

src/globus_cli/commands/flows/create.py
```diff
@@ -0,0 +1,206 @@
...
+
+    DEFINITION is the JSON document that defines the flow's instructions.
+    The definition document may be specified inline, or it may be
+    a path to a JSON file.
+
+        Example: Inline JSON:
+
+        \b
+            globus flows create 'My Cool Flow' \\
+            '{{"StartAt": "a", "States": {{"a": {{"Type": "Pass", "End": true}}}}}}'
```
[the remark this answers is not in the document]
ada-globus: IIRC it does, because CLI automatically does substitution on this downstream

src/globus_cli/commands/flows/create.py
```diff
@@ -0,0 +1,206 @@
...
+)
+@click.option(
+    "--starter",
+    "starters",
+    type=str,
+    multiple=True,
+    help="""
+        A principal that may run the flow.
+
+        Use "all_authenticated_users" to allow any authenticated user to run the flow.
```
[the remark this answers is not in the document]
ada-globus: What would you think about "start a new run of the flow"? I am a little hesitant to remove the word "run" only because it may make the meaning a little ambiguous (I do like the separation of using "start" as a verb and "run" as a noun so that we're not "running runs"—but while it seems clear to me that "starting" a flow creates a "run", I don't think that's necessarily obvious to a new user to the service).

src/globus_cli/commands/flows/create.py
```diff
@@ -0,0 +1,206 @@
...
+    help="""
+        The JSON input schema that governs the parameters
+        used to start the flow.
+
+        The input document may be specified inline, or it may be a path to a JSON file.
+
+        Example: Inline JSON:
+
+        \b
+            --input-schema '{"properties": {"src": {"type": "string"}}}'
```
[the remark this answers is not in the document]
ada-globus: AFAICT, it does not—the help text for the arguments does not appear to be treated the same way as the docstring.

Screenshot 2023-05-15 at 07 37 24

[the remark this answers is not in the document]
kurtmckee: Perfect!

src/globus_cli/commands/flows/create.py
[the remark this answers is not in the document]
kurtmckee: > "start a new run of the flow"

Nice. I like it!

PR #790, opened 2023-04-28
Rewrite JSONStringOrFile to accept filenames without the 'file:' prefix and have a simpler implementation

sirosen:
This work was motivated by [this thread in #788](https://github.com/globus/globus-cli/pull/788#discussion_r1180456544), and a subsequent conversation we had about what the ideal (backwards compatible) behavior would be.
The rewrite is intentionally done in many small phases, to make review feasible while still achieving the desired outcome.

In order, the changes are...
- Introduce `JSONStringorFileV2`, a rewritten implementation
- Convert commands, one by one, to `JSONStringOrFileV2`, fixing issues uncovered by this change along the way
- Remove `JSONStringOrFile`
- Rename `JSONStringOrFileV2` to `JSONStringOrFile` (rewrite ends here)
- Use `JSONStringOrFile` to replace any uses of `json.load` called on a `click.File` argument

### Enhancements

* JSON file parsing throughout the CLI has been made more uniform and robust. Commands which required files to be specified with the `file:` prefix now allow for filenames without the prefix, improving tab-completion. All commands which accept JSON data as inputs now allow for files or JSON-formatted arguments.


sirosen:
I found that this made my local `coverage report` fail the fail-under check because it was still seeing the old `prefix_mapper.py` file. Turns out, that was in the `build/` dir for `tox` and was getting picked up even through tox rebuilding.
I did a quick `find . -name 'prefix_mapper.py' -exec 'rm' '{}' ';'` to fix it, but if anyone else sees this...
- that `find` could be handy
- something might be wrong with the tox config, such that a stale file wasn't removed correctly?

src/globus_cli/commands/collection/update.py
```diff
@@ -151,11 +152,12 @@ def collection_update(
...
-    if sharing_restrict_paths is not None:
-        if not isinstance(sharing_restrict_paths, dict):
-            raise click.UsageError(
-                "--sharing-restrict-paths may not contain non-object JSON data"
-            )
+    if isinstance(sharing_restrict_paths, ParsedJSONData) and not isinstance(
+        sharing_restrict_paths.data, dict
+    ):
+        raise click.UsageError(
+            "--sharing-restrict-paths may not contain non-object JSON data"
```
kurtmckee, line 159: I recommend converting this from "may not contain non-object JSON data" to "must be a JSON object". I think that this will be a clearer statement of the requirements.

This suggestion applies to all locations that raise a UsageError based on the `.data` attribute.

Note that updating this will affect all of the tests that rely on the current double-negative texts.

src/globus_cli/parsing/param_types/json_strorfile.py
```diff
@@ -0,0 +1,107 @@
...
+    `file:` prefix which disambiguates an input as being a filename. e.g.
+        --opt 'file:./foo/bar.json'
+
+    The parsing happens in the following order of precedence:
+        - 'file:' prefix (filename)
+        - bare filename, including `-` for stdin
+        - JSON parse
+
+    This ordering protects against odd behavior if a filename happens to be valid JSON.
+    Most likely to happen with a JSON literal like an integer, e.g. `--opt 100`
```
kurtmckee, line 46: Based on my own past experience, I would recommend changing this example to `--opt 1`, which seems like a more likely filename if STDERR is piped incorrectly to STDOUT and results in piping to a file literally named `1`.

```suggestion
    Most likely to happen with a JSON literal like an integer, e.g. `--opt 1`
```

src/globus_cli/parsing/param_types/json_strorfile.py
```diff
@@ -0,0 +1,107 @@
...
+    """
+
+    def __init__(
+        self, *args: t.Any, null: t.Any | None = None, **kwargs: t.Any
+    ) -> None:
+        self.null = null
+        super().__init__(*args, **kwargs)
+
+    def get_metavar(self, param: click.Parameter) -> str:
+        return "[JSON_FILE|JSON|file:JSON_FILE]"
```
kurtmckee, line 60: My gut is that this won't be completely clear to a user. Does this change seem clearer, or is it already clear enough?

```suggestion
        return "[JSON_FILE|INLINE_JSON|file:JSON_FILE]"
```

kurtmckee (changes requested):

src/globus_cli/commands/collection/update.py
sirosen: Yeah, I like that much better. I'll run through and make the change in a jiffy.

src/globus_cli/parsing/param_types/json_strorfile.py
sirosen: I don't know that anything will suffice to totally clarify for users other than an example usage.

I'd be somewhat inclined to leave it as-is? I worry that `INLINE_JSON` could generate more confusion than the bare `JSON` -- e.g. users thinking that it's some limitation on the format to have no newlines.

This is all weakly held, but that's where I'm tending.

kurtmckee: Makes sense. I'm on board!

kurtmckee (approved):
Looks good to me. There's a failing unit test but after that, let's merge!

src/globus_cli/parsing/param_types/json_strorfile.py
```diff
@@ -0,0 +1,107 @@
...
+            return ParsedJSONData(filename=None, data=json.loads(value))
+        except json.JSONDecodeError:
+            # apply a *very weak* heuristic to try to get users accurate error
+            # messages when they make typos
+            #
+            # if the value looks like it's meant to be JSON, show an invalid JSON error
+            # but otherwise, it's a file-not-found error
+            if value.startswith("{"):
+                raise click.UsageError("parameter value did not contain valid JSON")
+            raise click.UsageError(f"FileNotFound: {value} does not exist")
```
ada-globus, line 107: I feel like this error would make more sense in `convert` since this method isn't involved with file handling.

tests/functional/search/test_delete_by_query.py
```diff
@@ -103,6 +103,29 @@ def test_dbq_query_string_and_document_mutex(run_line, tmp_path):
     assert "mutually exclusive" in result.stderr
 
 
+def test_dbq_rejects_non_object_document(run_line, tmp_path):
+    """
+    Check that `-q` and `--query-document` cannot be used together
+    """
```
ada-globus, line 109: This isn't what this is testing, right?

tests/functional/search/test_query.py
```diff
@@ -120,3 +120,26 @@ def test_query_required(run_line):
         assert_exit_code=2,
     )
     assert "Either '-q' or '--query-document' must be provided" in result.stderr
+
+
+def test_query_rejects_non_object_document(run_line, tmp_path):
+    """
+    Check that `-q` and `--query-document` cannot be used together
+    """
```
ada-globus, line 128: Same as above, I think this came along for the ride on a copy-and-paste :)

src/globus_cli/commands/flows/start.py
```diff
@@ -95,11 +95,15 @@ def start_command(
     """
 
     if input_document is None:
-        input_document = {}
+        input_document_json: dict[str, JsonValue] = {}
+    else:
+        if not isinstance(input_document.data, dict):
+            raise click.UsageError("Flow input cannot be non-object JSON data")
```
ada-globus, line 101: Can we change this one as well per @kurtmckee's note?

src/globus_cli/commands/search/delete_by_query.py
```diff
@@ -50,8 +53,12 @@ def delete_by_query_command(
 
     if q:
         doc: dict[str, t.Any] = {"q": q}
-    elif query_document:
-        doc = json.load(query_document)
+    elif query_document is not None:
+        if not isinstance(query_document.data, dict):
+            raise click.UsageError(
+                "--query-document cannot contain non-object JSON data"
```
ada-globus, line 59: Same as above, could we update?

ada-globus (commented):
This is a really nice improvement! Have just a couple very minor notes, but otherwise really excited to see this merge!

sirosen:
We time-spliced on review vs merge, but I'm going to look at the comments and see if we need a quick PR to fixup.

src/globus_cli/parsing/param_types/json_strorfile.py
sirosen: This was the only bit of feedback about which I'm hesitant, so I'm going to leave it for now.
Although it's a bit odd to see this message here, it's a side-effect of the fact that the parse is ambiguous in this case (and only this case).

I'm not sure how to structure this best, so I'm leaving it in its current clumsy state.

PR #791, opened 2023-04-28
Organize reflection code into 'globus_cli.reflect' and use it to improve annotation checking

sirosen:
- Move reflection tooling to a dedicated module, `globus_cli.reflect`, pulling from the doc generator and `globus_cli.utils`
- Use `reflect` to improve the implementation of the click annotation checker test to automatically pick up on any new modules

Once this is done, road-test the changes to the annotation checker by removing modules from the deny-list to ensure that the checker picks them up.
Using `globus endpoint activate`, verify that the checker can still be used to identify and fix a minor issue after the refactor.

This also enables the checker on many modules where it already passes, expanding the list of checks.

### DIFF
--- a/reference/_generate.py
+++ b/reference/_generate.py
@@ -11,11 +11,10 @@ import time
 
 import click
 import requests
-from pkg_resources import load_entry_point
 
-from globus_cli.utils import walk_contexts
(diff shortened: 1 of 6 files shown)


kurtmckee (approved):
Looks okay to me!

PR #792, opened 2023-04-28
Bugfix: the server-timing display always showed

sirosen:
The cause in this case was that the callback for `--show-server-timing` always fires, but should exit early if the flag is not set.


ada-globus (approved):
Was wondering what the cause was after seeing the thread where this was spotted! 👍

PR #794, opened 2023-05-01
Tighten up type checking on Search, Groups, and user-credential commands

sirosen:
For each category of commands targeted here, remove the exemptions from mypy.ini and the click type checker, then fix any resulting issues.
There are almost no behavioral changes here, other than some arguments which are now more consistently validated as `click.UUID` types.

Because so much "disable" config is removed, this is a negative diff, but that's somewhat deceptive -- in `src/` it's probably closer to break-even.


src/globus_cli/commands/search/index/role/create.py
```diff
@@ -10,6 +11,11 @@
 
 from ..._common import index_id_arg, resolved_principals_field
 
+if sys.version_info >= (3, 8):
+    from typing import Literal
+else:
+    from typing_extensions import Literal
+
```
kurtmckee, line 18: Elsewhere (for example, in `group/set_policies.py`), this is expressed slightly differently:

```python
if sys.version_info < (3, 9):
    from typing_extensions import Literal
else:
    from typing import Literal
```

This suggests that `typing.Literal` is available in Python 3.8 (in this file) but not in Python 3.8 (in `group/set_policies.py`).

Please double-check which of these conditions is correct.

kurtmckee (changes requested):

src/globus_cli/commands/search/index/role/create.py
sirosen: As we discussed on Friday, but for the public record:
The check is wrong. `Literal` was added in 3.8.

I started writing a lint rule which to catch this and require `>=` comparisons only (to simplify the check), but I realized that there are some complex potential cases outside of typing, e.g.
```python
if (3,7) <= sys.version_info < (3,9): ...
```

I don't know why you would need a ternary version comparison like that, but a linter would need to handle it.

So for now I'm going to table that work. Maybe if the only thing inside of the `if` branch is a `typing` or `typing_extensions` import, that would be a good way to identify it... I'll think about it.

sirosen: In resolving this, I expanded the scope of the PR slightly, to touch all `sys.verison_info` comparisons for typing. So there are a few more files touched, but now it's all uniformly using `>=`.

kurtmckee (approved):

PR #809, opened 2023-06-20
Support deleting runs in the Flows service

kurtmckee:
Also, bump the SDK version 3.21.0 to support this operation.

This forms the basis of additional, upcoming work to manage runs in the Flows service.

### DIFF
--- a/changelog.d/20230620_150848_kurtmckee_add_run_delete_sc_18418.md
+++ b/changelog.d/20230620_150848_kurtmckee_add_run_delete_sc_18418.md
@@ -1,0 +1,3 @@
+### Enhancements
+
+* Add a `globus flows run delete` command.
--- a/src/globus_cli/commands/flows/__init__.py
+++ b/src/globus_cli/commands/flows/__init__.py
@@ -9,6 +9,8 @@ from globus_cli.parsing import group
         "list": (".list", "list_command"),
         "show": (".show", "show_command"),
         "start": (".start", "start_command"),
+        # "run" is a subgroup of commands.
+        "run": (".run", "run_command"),
     },
 )
 def flows_command():
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ setup(
     package_dir={"": "src"},
     python_requires=">=3.7",
     install_requires=[
-        "globus-sdk==3.20.1",
+        "globus-sdk==3.21.0",
         "click>=8.0.0,<9",
         "jmespath==1.0.1",
         "packaging>=17.0",
--- a/src/globus_cli/commands/flows/run/__init__.py
+++ b/src/globus_cli/commands/flows/run/__init__.py
@@ -1,0 +1,11 @@
+from globus_cli.parsing import group
+
+
+@group(
+    "run",
+    lazy_subcommands={
+        "delete": (".delete", "delete_command"),
+    },
+)
+def run_command() -> None:
+    """Interact with a run in the Globus Flows service"""
--- a/src/globus_cli/commands/flows/run/delete.py
+++ b/src/globus_cli/commands/flows/run/delete.py
@@ -1,0 +1,31 @@
+from __future__ import annotations
+
+import uuid
+
+from globus_cli.login_manager import LoginManager
+from globus_cli.parsing import command, run_id_arg
+from globus_cli.termio import Field, TextMode, display, formatters
+
+
+@command("delete", short_help="Delete a run")
+@run_id_arg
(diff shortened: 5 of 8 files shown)


changelog.d/20230620_150848_kurtmckee_add_run_delete_sc_18418.md
```diff
@@ -0,0 +1,3 @@
+### Enhancements
+
+* Support deleting runs in the Flows service.
```
sirosen, line 3: Let's enhance this changelog to list the new command's name, `globus flows runs delete`.

Or `globus flows run delete` if we take that naming approach.

src/globus_cli/commands/flows/__init__.py
```diff
@@ -9,6 +9,8 @@
         "list": (".list", "list_command"),
         "show": (".show", "show_command"),
         "start": (".start", "start_command"),
+        # "runs" is a subgroup of commands.
+        "runs": (".runs", "runs_command"),
```
sirosen, line 13: I don't have a strong opinion, but I think we should discuss `runs` vs `run` as the name here.

`flows` is plural because the name of the service is Globus Flows -> `globus flows ...`
But a `run` is a singular object -> `run delete`

We already have several singulars like `globus task` (Transfer Task), `globus search task`, etc.

We also have _one_ case (as far as I know) where we made a mistake and have a singular for a service name which is supposed to be pluralized: Globus Groups -> `globus group`. (oops!)

What I'm saying is:
- there is some prior art which suggests a singular for an object name, and a plural/singular choice to match the service name
- but we haven't necessarily applied that set of rules consistently
- it's therefore unclear what the best choice is for `run` vs `runs`

I think I tend towards `globus flows run` (singular) and treating `globus group ...` as a minor bug which we should fix at some point by aliasing to `globus groups ...`. But I could be convinced otherwise.

sirosen (changes requested):
I'd like to see the addition of a run deletion test -- just something which gets the body of the function to run.

I have comments on `run` vs `runs` but I don't want it to sound like I have a strong opinion. More: I want us to make sure we are making an active choice about the name, with as much of the information available as is possible.

src/globus_cli/commands/flows/__init__.py
kurtmckee: I'm on board for `run` (singular). I had thought I saw a "permissions" directory -- plural -- and followed that without double-checking that initial glance. So! Singular it is.

changelog.d/20230620_150848_kurtmckee_add_run_delete_sc_18418.md
kurtmckee: Done.

src/globus_cli/commands/flows/__init__.py
sirosen: > I'm on board for `run` (singular). I had thought I saw a "permissions" directory -- plural -- and followed that without double-checking that initial glance. So! Singular it is.

FWIW, I'm pretty sure you saw correctly and it's another place which is inconsistent with the other cases. (At what point are the inconsistent things so numerous that we can't talk confidently about consistency? :sweat_smile: )

I'd like to put together a game plan for aliasing and deprecating old names, but I haven't written anything down yet.

derek-globus: As a general rule (if we're defining the norm), idk how I feel about `globus <service> <domain obj> <verb>` as a cli structure

I think `globus <service> <operation>` makes more sense (so `globus flows delete_run`)

derek-globus: I guess our current flows approach is to treat the `domain obj` as implied if the object name matches the cli structure.

So `globus flows show` shows a flow because `flows` implies `globus flows flow show` 

But that pattern feels decently wrong to me.

derek-globus: Eh, I'm not sure whether this PR is the right place to have this discussion.

I guess it does add a new sub-keyword to the CLI so maybe it is

derek-globus: Discussed offline w/ Kurt & Stephen.

We're going to go with `globus flows run delete` for internal consistency in the cli (even if the CLI is not entirely consistent, discrepencies are considered the exception not the rule for this version)

We'll evaluate the naming structure of `globus <service> <operation>` as a potential migration in a subsequent major version (with the idea of maintaining aliasing for backwards compatability)

derek-globus (approved):

sirosen (approved):

Issue #806, opened 2023-06-17
How necessary is the upper bound on cryptography version?


xylar:
We are having a problem on conda-forge where `cryptography` only migrated to `openssl` 3 as of version 38.  However, the newest packages are being built only with `openssl` 3 (after an interim period where packages were being built with both `openssl` 1 and 3).  This means that `globus-cli` cannot be installed with some of the latest packages (notably but not limited to `dask`) because of the constraint on cryptography:
https://github.com/globus/globus-cli/blob/main/setup.py#L55
My question is how necessary this constraint might be and what it would take to get support for cryptography 38 (and hopefully beyond).  This would help me a great deal.

If I can clarify further, don't hesitate to ask.

sirosen:
I think we can completely remove the upper bound. I'll revisit this on Monday when I'm back at my desk to make sure the team agrees.

Having the version capped used to be more important to us, but I don't think it's necessary anymore.

---

Aside: thanks for the way you engaged on this and opened this issue! It made it easy to understand your use case and I appreciate that you were generous towards us and assumed we had a reason for the upper bound.

xylar:
@sirosen, thanks very much for being willing to have a look at this.  There isn't a big rush from my perspective but having this sorted out some way or other by the beginning of August would be a big help to me because I maintain and deploy a conda metapackage on US DOE machines that has `globus-cli` along with a number of other conda packages that we're working to update to the latest versions.

Let me know how your testing goes and how I can be of help.

sirosen:
I've just put in a PR to remove the upper bound and @-mentioned you so that you should receive a notification when it merges. I'll also hold off on closing this until we do a release with an open upper bound on `cryptography`.

There is one thing that concerns me slightly, which is this part of your message:

> I maintain and deploy a conda metapackage on US DOE machines that has `globus-cli` along with a number of other conda packages that we're working to update to the latest versions.

Generally speaking, we recommend installing `globus-cli` into a dedicated virtualenv. Since it's an application and not designed to be used as a library (i.e. nobody should ever `import globus_cli` in their code), installing it into an environment with other packages isn't a use case we actively take measures to support.

Although it will often work to install `globus-cli` alongside other packages, we can only be flexible on some of our requirements data, and there are certain dependency conflicts which we would consider not to be bugs.
In particular, we pin the exact version of `globus-sdk` always, and we may pin other packages (e.g., today we have `jmespath` pinned) as necessary.

[...]

PR #810, opened 2023-06-21
Remove `cryptography` upper bound

sirosen:
Prompted by #806, we don't need to be capping `cryptography` anymore (and besides, our upper bound is out of date).

I don't think this belongs in the more widely distributed changelog, so I'll update that issue to make sure @xylar is informed directly when we do our next release.


kurtmckee (approved):

xylar:
Thanks!  Consider me informed!

PR #812, opened 2023-06-22
Add `globus flows update` command

kurtmckee:
### Enhancements

* Add a `globus flows update` command.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -11,7 +11,7 @@ source =
 [report]
 show_missing = True
 skip_covered = True
-fail_under = 85
+fail_under = 86
 
 exclude_lines =
     # the pragma to disable coverage
--- a/src/globus_cli/commands/flows/update.py
+++ b/src/globus_cli/commands/flows/update.py
@@ -137,23 +137,6 @@ def update_command(
             raise click.UsageError("--input-schema must be a JSON object")
         input_schema_doc = input_schema.data
 
-    # Empty lists (like `xs = []`) will result in erasure of the option.
-    # Only pass values if `--x ...` was passed (in which case `xs` will be truthy)
-    # or `--no-xs` was specified (in which case `xs` will be an empty list).
-    # These conditions are guaranteed because `--x` and `--no-xs` are mutex options.
-    prepared_administrators: list[str] | None = None
-    if administrators is not None:
-        prepared_administrators = list(administrators)
-    prepared_starters: list[str] | None = None
-    if starters is not None:
-        prepared_starters = list(starters)
-    prepared_viewers: list[str] | None = None
-    if viewers is not None:
-        prepared_viewers = list(viewers)
-    prepared_keywords: list[str] | None = None
-    if keywords is not None:
-        prepared_keywords = list(keywords)
-
     # Configure clients
     flows_client = login_manager.get_flows_client()
     auth_client = login_manager.get_auth_client()
@@ -166,10 +149,10 @@ def update_command(
         subtitle=subtitle,
         description=description,
         flow_owner=owner,
-        flow_administrators=prepared_administrators,
-        flow_starters=prepared_starters,
-        flow_viewers=prepared_viewers,
-        keywords=prepared_keywords,
+        flow_administrators=administrators,
+        flow_starters=starters,
+        flow_viewers=viewers,
+        keywords=keywords,
     )
 
     # Configure formatters for principals
--- a/src/globus_cli/commands/flows/update.py
+++ b/src/globus_cli/commands/flows/update.py
@@ -5,22 +5,17 @@ import uuid
 import click
 
 from globus_cli.commands.flows._common import (
-    administrators_option,
     description_option,
     input_schema_option,
-    keywords_option,
-    starters_option,
     subtitle_option,
-    viewers_option,
 )
 from globus_cli.login_manager import LoginManager
 from globus_cli.parsing import (
+    CommaDelimitedList,
     JSONStringOrFile,
-    MutexInfo,
     ParsedJSONData,
     command,
     flow_id_arg,
-    mutex_option_group,
 )
 from globus_cli.termio import Field, TextMode, display, formatters
 from globus_cli.types import JsonValue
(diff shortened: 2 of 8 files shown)


.coveragerc
```diff
@@ -11,7 +11,7 @@ source =
 [report]
 show_missing = True
 skip_covered = True
-fail_under = 85
+fail_under = 86
```
ada-globus, line 14: Love to see it!

src/globus_cli/commands/flows/update.py
```diff
@@ -0,0 +1,210 @@
...
+        definition_doc = definition.data
+
+    # Ensure the input schema is a JSON object (if provided)
+    input_schema_doc: dict[str, JsonValue] | None = None
+    if input_schema is not None:
+        if not isinstance(input_schema.data, dict):
+            raise click.UsageError("--input-schema must be a JSON object")
+        input_schema_doc = input_schema.data
+
+    # Empty lists (like `xs = []`) will result in erasure of the option.
```
ada-globus, line 142:
```suggestion
    # Passing an empty role list (like `xs = []`) will result in erasure of the option.
```

Might be slightly clearer that this is referring to the behavior of the API rather than intended behavior or the CLI?

tests/functional/flows/test_update_flow.py
```diff
@@ -0,0 +1,243 @@
...
+        return identity
+
+
+def value_for_field_from_output(name, output):
+    """
+    Return the value for a specified field from the output of a command.
+    """
+    match = re.search(rf"^{name}:[^\S\n\r]+(?P<value>.*)$", output, flags=re.M)
+    assert match is not None
+    return match.group("value")
```
ada-globus, line 110: At a glance this looks the same as the helpers in `test_create_flow`, in which case maybe we should move to a shared module?

src/globus_cli/commands/flows/update.py
```diff
@@ -0,0 +1,210 @@
...
+    """,
+)
+@subtitle_option
+@description_option
+@input_schema_option
+# "--no-Xs" options must exist to clear lists of things (like administrators).
+# These must be mutually exclusive with the flags that set options.
+@administrators_option
+@click.option(
+    "--no-administrators",
```
ada-globus, line 73: I think `no-` options are probably fine, but would it be clearer if we went with something like `clear-`, `reset-`, or `erase-`?

ada-globus (commented):
Overall this looks really good. I have a few thoughts, but I could be persuaded otherwise if I am missing context or if you have strong feelings to the contrary!

sirosen (commented):
I'm stating this as a bit of a matter of record about how/why we're changing this. Please correct any of this if I'm misstating our discussion.

---

We discussed the fact that `--administrator/--no-administrators` diverges from the existing two patterns for managing a list field (including clearing it).
Those are exemplified by options to `globus collection update`. Specifically, (1) `--sharing-users-allow ""` and (2) `--keywords ""`. (1) is that we accept a `multiple=True` string field and convert `""` to mean "clear the field". (2) is that we accept a `CommaDelimitedList` and convert `""` to mean "clear".

After a lengthy discussion, we agreed on the merits of (2) as the overall approach, changing the option string to `--administrators`. We also agreed that `globus flows create` should be updated in future work to match.

The important item here in terms of direction and interface for the CLI is that we are modeling list fields as `CommaDelimitedList`. I would go so far as to state that this is a norm we are now establishing and which we can follow and converge towards (where possible) in the future.

CSV has the advantage that the empty list is naturally encoded as `""`.

tests/functional/flows/test_update_flow.py
kurtmckee: You are correct! I thought about doing that, but now I'll actually do it!

src/globus_cli/commands/flows/update.py
kurtmckee: As discussed, we'll target a comma-separated list of strings, and allow an empty string to act as the sentinel value for erasure.

tests/functional/flows/test_update_flow.py
kurtmckee: Done.

tests/functional/flows/test_update_flow.py
```diff
@@ -0,0 +1,162 @@
...
+    omitted_keys = {"flow_administrators", "flow_starters", "flow_viewers", "keywords"}
+    assert json.loads(request.body).keys() & omitted_keys == set()
+
+
+@pytest.mark.parametrize("option", ["definition", "input-schema"])
+def test_json_object_option_validation(option, run_line):
+    """Ensure options that must be JSON objects are validated as such."""
+
+    command = f"globus flows update {uuid.uuid4()} --{option} '[]'"
+    run_line(command, assert_exit_code=2, matcher="must be a JSON object")
```
sirosen, line 162: I wish that `matcher` did this! But it does something a bit weirder/clumsier:
https://github.com/globus/globus-cli/blob/3e27e695631bc5f0122901f02c4130f66cfcb37f/tests/conftest.py#L275

I would be onboard to change the interface to support this usage (may I suggest we follow `pytest`'s `match=...` arg and support regexes?), and find some other way of handling the cases which use the current `matcher=True` behavior.

For now though, let's make this capture output and do an `assert ... in output`?

sirosen (changes requested):
I need to think more about how we'll make the changes to `globus flows create` to match this, but that's future work. I think we're good to merge this, except for one snag in a test.

tests/functional/flows/test_update_flow.py
kurtmckee: This is resolved now. Thanks for catching this!

src/globus_cli/commands/flows/update.py
kurtmckee: This is resolved.

sirosen (approved):

ada-globus:
> I need to think more about how we'll make the changes to `globus flows create` to match this, but that's future work. I think we're good to merge this, except for one snag in a test.

I think I'd be onboard with just breaking the existing args. IMO the last release was recent enough, we just need to call it out, explain why we changed it (i.e., "we realized that flow update required a different format, and they should be consistent") and apologize but remove the prior args ASAP to prevent future confusion.

PR #813, opened 2023-07-03
Add a `globus flows run update` command

kurtmckee:
### Enhancements

* Add a `globus flows run update` command.

### DIFF
--- a/src/globus_cli/commands/flows/run/update.py
+++ b/src/globus_cli/commands/flows/run/update.py
@@ -1,0 +1,108 @@
+from __future__ import annotations
+
+import uuid
+
+import click
+
+from globus_cli.login_manager import LoginManager
+from globus_cli.parsing import CommaDelimitedList, command, run_id_arg
+from globus_cli.termio import Field, TextMode, display, formatters
+
+
+@command("update", short_help="Update a run")
+@run_id_arg
+@click.option(
+    "--label",
+    type=str,
+    help="A label to give the run.",
+)
+@click.option(
+    "--managers",
+    "run_managers",
+    type=CommaDelimitedList(),
+    help="""
+        A comma-separated list of principals that may manage the execution of the run.
+
+        Passing an empty string will clear any existing run managers.
+    """,
+)
+@click.option(
+    "--monitors",
+    "run_monitors",
+    type=CommaDelimitedList(),
+    help="""
+        A comma-separated list of principals that may monitor the execution of the run.
+
+        Passing an empty string will clear any existing run monitors.
+    """,
+)
+@click.option(
+    "--tags",
+    "tags",
+    type=CommaDelimitedList(),
+    help="""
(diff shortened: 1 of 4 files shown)


src/globus_cli/commands/flows/run/update.py
```diff
@@ -0,0 +1,108 @@
...
+@command("update", short_help="Update a run")
+@run_id_arg
+@click.option(
+    "--label",
+    type=str,
+    help="A label to give the run.",
+)
+@click.option(
+    "--managers",
+    "run_managers",
```
ada-globus, line 21: Is there a reason why you adopted an alternate name here but not in the principals for the `flows update` command (I'm guessing we haven't aliased these at the API level...)? No objection, just double checking.

ada-globus (approved):

src/globus_cli/commands/flows/run/update.py
kurtmckee: Do you mean "not in the principals for the `flow start` command"?

Flows use different terms (administrators, starters, viewers) than Runs (monitors, managers).

As for the difference with `flows start`, I thought the plan was to modify how those commands work, and asap.

ada-globus: Sorry, I should have been more precise: In other commands we are not using `flows_` or `run_` prefixes. I was just wondering why this one is.

tests/functional/flows/test_update_run.py
```diff
@@ -0,0 +1,118 @@
...
+        "Run ID",
+        "Run Label",
+        "Run Managers",
+        "Run Monitors",
+        "Run Tags",
+    }
+    for field in expected_fields:
+        assert field in result.output, f"'{field}' is not present in the output"
+
+    tag_match = re.search(r"^Run Tags:\s+(?P<tags>.+)$", result.output, flags=re.M)
```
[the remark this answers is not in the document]
ada-globus: Yeah, agree, that would be a bit more intuitive!

src/globus_cli/commands/flows/run/update.py
kurtmckee: Ah, gotcha.

I did that so it was clearer where that would map to in the API call.

kurtmckee: I can change that if desired.

tests/functional/flows/test_update_run.py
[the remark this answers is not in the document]
sirosen: Okay, I'm going to have to circle back to this. I think the `result` is coming directly from `click`, so we'll either want to customize or wrap it.

src/globus_cli/commands/flows/run/update.py
ada-globus: I don't have a very strong feeling, I just wanted to be sure I understood the intent. This is fine as-is, IMO.

PR #815, opened 2023-07-06
Add a `globus flows run show` command

kurtmckee:
### Enhancements

* Add a `globus flows run show` command.


ada-globus (approved):
This looks great, and local tests work well. One thing I noticed that I hadn't before: We're adding a comma and a space as our delimiter for tags and keywords, but as configured the principals are using a comma with no trailing space. Might be worth making consistent if it's easy, but regardless I'm going to approve as I don't think that's a blocking issue (IMO).

PR #817, opened 2023-07-11
Rewrite the test output matching utilities to be simpler to use

sirosen:
The overall goal here was to replace `matcher=True` (which changes the return type and produces an `OutputMatcher`. Awkward!) with something that feels more similar to the `match=...` behavior in `pytest.raises` and `pytest.warns`, in which you pass a regex.

I encourage commit-by-commit reading for this PR.
The commits are done in small units to show the evolution of a replacement in which `run_line` now accepts `match_out=...` and `match_err=...` for regex matching.
First the new helpers are introduced, then rolled out across the tests, and then finally `OutputMatcher` is dropped and the new helpers are refined a little.

I've included a bit of a "bonus" feature here which is particularly helpful for us. `match_out=("Foo", "bar")` is allowed as a shorthand for the regex which matches `^Foo:\s+bar$`. That aligns well with our colon-delimited output.

I couldn't find a better way to capture pytest verbosity than recording it during `pytest_configure`, but I'm open to refining that more.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -291,8 +291,13 @@ network calls recorded:
 
 
 def _assert_matches(text, text_name, match):
-    if isinstance(match, str) or isinstance(match, re.Pattern):
+    if isinstance(match, (str, re.Pattern, tuple)):
         match = [match]
+    elif not isinstance(match, list):
+        raise NotImplementedError(
+            "match_{out,err} got unexpected arg type: {type(match)}"
+        )
+
     match = [_convert_match_tuple(m) for m in match]
 
     compiled_matches = [
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -239,7 +239,14 @@ def run_line(cli_runner, request, patch_tokenstorage):
     for easier debugging.
     """
 
-    def func(line, assert_exit_code=0, stdin=None, matcher=False):
+    def func(
+        line,
+        assert_exit_code=0,
+        stdin=None,
+        match_out=None,
+        match_err=None,
+        matcher=False,
+    ):
         from globus_cli import main
 
         # split line into args and confirm line starts with "globus"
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -272,13 +279,43 @@ def func(line, assert_exit_code=0, stdin=None, matcher=False):
...
+    if isinstance(match, str) or isinstance(match, re.Pattern):
+        match = [match]
+    match = [_convert_match_tuple(m) for m in match]
+
+    compiled_matches = [
+        m if isinstance(m, re.Pattern) else re.compile(m, re.MULTILINE) for m in match
+    ]
+    for pattern in compiled_matches:
+        assert (
+            pattern.search(text) is not None
```
kurtmckee: `.search()` and `.match()` have different characteristics. Would it be sensible to call the parameters `search_stdout` and `search_stderr`?

tests/conftest.py
```diff
@@ -272,13 +279,43 @@ def func(line, assert_exit_code=0, stdin=None, matcher=False):
...
+        ), f"Pattern {pattern} not found in {text_name}"
+
+
+def _convert_match_tuple(match):
```
kurtmckee, line 293: Same as above -- `.search()` is used, not `.match()`.

tests/conftest.py
```diff
@@ -272,13 +279,43 @@ def func(line, assert_exit_code=0, stdin=None, matcher=False):
...
+    # tuple of ("Foo", "bar") converts to a regex for
+    #       "Foo:  bar"
+    if isinstance(match, tuple):
+        assert len(match) == 2
+        field_name, field_value = match
+        assert isinstance(field_name, str)
+        assert isinstance(field_value, str)
+        match = f"^{re.escape(field_name)}:\\s+{re.escape(field_value)}$"
```
kurtmckee, line 301: The trailing `$` precludes prefix matching.

tests/conftest.py
```diff
@@ -239,7 +239,14 @@ def run_line(cli_runner, request, patch_tokenstorage):
     for easier debugging.
     """
 
-    def func(line, assert_exit_code=0, stdin=None, matcher=False):
+    def func(
+        line,
+        assert_exit_code=0,
+        stdin=None,
+        match_out=None,
+        match_err=None,
```
kurtmckee, line 221: I have a preference to make it explicit that STDOUT and STDERR are being searched.

```suggestion
        match_stdout=None,
        match_stderr=None,
```

tests/functional/search/test_index.py
```diff
@@ -40,14 +42,15 @@ def test_index_create(run_line):
...
 
 def test_index_delete(run_line):
     meta = load_response_set("cli.search").metadata
     index_id = meta["index_id"]
 
-    result = run_line(f"globus search index delete {index_id}")
-    assert f"Index {index_id} is now marked for deletion." in result.output
+    run_line(
+        f"globus search index delete {index_id}",
+        match_out=f"Index {re.escape(index_id)} is now marked for deletion.",
```
kurtmckee, line 55: I think the `re.escape()` call is harmless, but I don't think it's needed -- UUIDs don't trigger special regex behavior.

tests/conftest.py
sirosen: Yeah, I'm okay with using `search_*` to indicate that it's a regex search.

tests/functional/search/test_index.py
sirosen: Yes, it is harmless... I waffled about whether or not to include it, since it's a no-op.
Why don't I remove it, now that it's come up?

I like `re.escape`-ing anything which looks like "input", but have no particular attachment to doing it in this case.

tests/conftest.py
kurtmckee: Yeah, that makes sense. I'm onboard.

kurtmckee (approved):

PR #824, opened 2023-07-20
Update to globus-sdk v3.25.0

sirosen:
This picks up globus/globus-sdk-python#790, which is a nice internal fix but which I don't think is worthy of any public changelog note here.

We had a conversation in which we agreed to use the SDK default of 0.5s for leeway from that fix, rather than passing an explicit value here.


ada-globus (approved):
👍

PR #826, opened 2023-07-20
Replace use of the SDK paginator table with 'wrap'

sirosen:
The paginator table attached to client objects provides a convenient syntax for accessing paginators which wrap the relevant underlying client methods.

However, it is not presently possible to explain it to type checkers that this is a type-manipulating proxy over the client class. The paginator table is just one valid syntax for accessing a Paginator object built from the method being paginated. The paginator building is actually done by `Paginator.wrap`, which inspects the method for special pagination attributes and dispatches to its relevant subclass.

The major advantage of `Paginator.wrap()` is that it preserves type annotation information correctly, returning a paginator object parametrized over the return type of the underlying method.

As a result, we should prefer `Paginator.wrap` for mypy-checked code, and [the current SDK docs even have a note to this effect](https://globus-sdk-python.readthedocs.io/en/stable/core/paging.html#typed-paginators-with-paginator-wrap).

In order to ensure the move is done everywhere and holds into the future, add a pygrep hook which looks for access to a `.paginated` attribute. If this runs into false positives in the future, it could be replaced with an AST-based lint or something else more sophisticated, but on the current codebase it is fully accurate.

The change can be verified to be effective by introducing unuspported arguments and seeing that `mypy` catches them. For example:

```diff
-    res = paginator(task_id).items()
+    res = paginator(task_id, foo=bar).items()
```

is flagged:

```
$ mypy src/
src/globus_cli/commands/task/show.py:71: error: Unexpected keyword argument "foo"  [call-arg]
```

Such a check was not possible via the paginator-table interface (the `.paginated` attribute).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -41,3 +41,16 @@ repos:
       entry: ./changelog.d/post-fix-changelog.py
       language: python
       files: ^changelog\.adoc$
+    - id: reject-paginator-table-usage
+      name: "Require use of 'Paginator.wrap()' not 'client.paginated.foo'"
+      types: [python]
+      language: pygrep
+      # look for either
+      #     x = client.paginated.foo(...)
+      # or
+      #     x = (
+      #         client
+      #         .paginated
+      #         .foo(...)
+      #     )
+      entry: '(\.paginated\.)|(^\s+\.paginated$)'
--- a/src/globus_cli/commands/endpoint/search.py
+++ b/src/globus_cli/commands/endpoint/search.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import click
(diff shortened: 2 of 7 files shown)


.pre-commit-config.yaml
```diff
@@ -41,3 +41,16 @@ repos:
...
+      language: pygrep
+      # look for either
+      #     x = client.paginated.foo(...)
+      # or
+      #     x = (
+      #         client
+      #         .paginated
+      #         .foo(...)
+      #     )
+      entry: '(\.paginated\.)|(^\s+\.paginated$)'
```
kurtmckee, line 56: ❤️

kurtmckee (approved):
Love the use of a pre-commit hook here!

PR #829, opened 2023-07-21
Add type annotations to the login manager subpackage

sirosen:
This was done in two commits because the second part has a surprising and outsized impact.
The first commit just applies some simple annotations in a few modules.

The second commit's full message is reproduced below.

---

[Type annotate LoginManager and accommodate typing](https://github.com/globus/globus-cli/commit/5d6ccc263d938f88d419089c0e78a7371bb5eb48)

```
Most of the annotation changes have no broad impact.
However, one change in particular is nefariously tricky:
`requires_login`.

This is a decorator factory which needs to be annotated in a way to
indicate that it adds an argument. ParamSpec and typing do not support
annotating the behavior we have today, in which the decorator adds the
`login_manager` keyword argument. The closest possible phrasing is to
use ParamSpec+Concatenate to indicate that we pass it as the first
positional argument. Therefore, *change the runtime logic to use this
pattern* and apply the ParamSpec-based annotations needed to validate
this in all commands.

In all cases where `login_manager` is passed via `requires_login`,
move it to be the only positional argument and annotate all other
command parameters as keyword-only, using the `*` separator. This also
uniformizes several commands which had drifted away from the preferred
style of keyword-only argument declarations.

The final detail on this change is that `mypy` has recently
experienced a regression in the handling of ParamSpecs under complex
scenarios like decorator factories. There is at least one potential
fix in progress, but it is unclear when it might release as it does
not appear to be ready for the upcoming v1.5 release.
However, `mypy` *does* correctly understand the actual type of such a
decorator factory -- this can be tested with `reveal_type` -- so we
can use this feature but with a type-ignore in a relatively
critical-looking location (the return value of the decorator
factory).

In addition to these changes, applying this change to
`session show` resulted in some other annotation issues being flagged
by mypy, so those are fixed here as well.
```

### DIFF
--- a/src/globus_cli/commands/session/show.py
+++ b/src/globus_cli/commands/session/show.py
@@ -1,4 +1,7 @@
+from __future__ import annotations
+
 import datetime
+import typing as t
 
 import globus_sdk
 
@@ -44,7 +47,7 @@ $ globus session show --format json
 """,
 )
 @LoginManager.requires_login("auth")
-def session_show(*, login_manager):
+def session_show(login_manager: LoginManager):
     """List all identities in your current CLI auth session.
 
     Lists identities that are in the session tied to the CLI's access tokens along with
@@ -56,6 +59,7 @@ def session_show(*, login_manager):
     # get a token to introspect, refreshing if necessary
     try:
         # may force a refresh if the token is expired
+        assert auth_client.authorizer
         auth_client.authorizer.get_authorization_header()
     except AttributeError:  # if we have no RefreshTokenAuthorizor
         pass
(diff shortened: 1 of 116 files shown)


src/globus_cli/commands/session/show.py
```diff
@@ -56,15 +59,16 @@ def session_show(*, login_manager):
     # get a token to introspect, refreshing if necessary
     try:
         # may force a refresh if the token is expired
+        assert auth_client.authorizer
```
kurtmckee, line 62: This line is for mypy's benefit, correct?

kurtmckee (approved):

PR #836, opened 2023-08-08
Implement complex run resume using experimental SDK components

sirosen:
Note that this changeset updates to SDK v3.26.0 in order to pull in the new functionality.

The src-tree changes turned out even smaller and simpler than I had hoped, although the testing is a bit verbose.

Experimental GARE parsing is used to convert from an Inactive Run document to a `required_scopes` list.
Experimental scope parsing is used to convert that `required_scopes` list to a forest.
The same subtree matching which was written for MutableScope is now used to compare the forest against the auth-response.

Prompting is done via a new mechanism.
A new internal error class is introduced for explicitly triggering the ConsentRequired handling path -- think of this as a "concrete auth requirements error" whereas the GARE parse gave us an abstract one. Note that it only supports `required_scopes` for now.

I decided to normalize `MutableScope` -> `scope_parser.Scope` in a somewhat suboptimal but very easy way by calling `parse(str(...))`, but this could be changed easily.

### DIFF
--- a/changelog.d/20230808_165447_sirosen_complex_run_resume.md
+++ b/changelog.d/20230808_165447_sirosen_complex_run_resume.md
@@ -1,0 +1,5 @@
+### Enhancements
+
+* Improve `globus flows run resume` to be capable of detecting missing consents and prompt
+  for reauthentication via `globus session consent`. The consent check can also
+  be skipped with `--skip-inactive-reason-check`.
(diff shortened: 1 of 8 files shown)


kurtmckee (approved):
I approve this, though as I mentioned previously, it will be nice if the `isinstance --> Scope` conditional branch is tested as well.

sirosen:
> I approve this, though as I mentioned previously, it will be nice if the `isinstance --> Scope` conditional branch is tested as well.

Yeah, that will be worth going after. I'll fix the linting issue first though (whatever that is) so that this can be unblocked.
And then I'll try to close up any test coverage gaps _today_, potentially after a release.

PR #837, opened 2023-08-10
Add a `globus flows run show-definition` command

kurtmckee:
### Enhancements

* Add a `globus flows run show-definition` command.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -11,7 +11,7 @@ source =
 [report]
 show_missing = True
 skip_covered = True
-fail_under = 86
+fail_under = 87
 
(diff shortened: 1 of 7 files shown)


ada-globus (commented):
This looks good! I like the addition to opt-out of sorting keys (unfortunate that the service reorders what the user gives us, but at least this prevents further reordering on top of that). Would love to see a test or two. Will wait smash approve til CI is passing (understand this is blocked by SDK) but otherwise I'm good with what's here so far!

sirosen (approved):

ada-globus (approved):

PR #838, opened 2023-08-10
Add `globus flows run show-logs` command

ada-globus:
image

# Notes
- I ran into a bit of difficulty getting a test written for pagination without being able to disable strict matching. I do know that pagination is working in practice against the production service, but I'll follow up with a test demonstrating this once I work out the matching issue. I do not expect any changes to the implementation.
- There are two modes: The default tabular mode and the `--details` text record mode
- Per conversation in Slack, we've agreed that `--format json` is adequate for the "piping" TAC in the associated story
- When printing details, the command manually converts the details object back to JSON to facilitate better copy/paste and/or parsing by subsequent commands. Happy to drop this if we have serious objections, but seemed like a nice affordance.

### DIFF
--- a/changelog.d/20230810_084859_ada_sc_18432.md
+++ b/changelog.d/20230810_084859_ada_sc_18432.md
@@ -1,0 +1,3 @@
+### Enhancements
+
+* Add `globus flows run show-logs` for showing a *run*'s log entries
--- a/src/globus_cli/commands/flows/run/__init__.py
+++ b/src/globus_cli/commands/flows/run/__init__.py
@@ -5,6 +5,7 @@ from globus_cli.parsing import group
     "run",
     lazy_subcommands={
         "show": (".show", "show_command"),
+        "show-logs": (".show_logs", "show_logs_command"),
         "list": (".list", "list_command"),
         "update": (".update", "update_command"),
         "delete": (".delete", "delete_command"),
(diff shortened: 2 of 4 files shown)


kurtmckee (approved):

kurtmckee (approved):

sirosen (approved):
I like everything in this changeset, but the image especially much.

I've tried out a bunch of usages and I'm pleased to say that the results on my moderately-wide terminal are as readable as one could hope.

There's a minor usage issue which I flagged for some internal discussion, but which exceeds the bounds of this body of work. `globus flows run show-logs` shows an unintuitive error if the log is expired/missing. The result at first may make it seem like `run list` and `run show-logs` are in disagreement, but this is all a service-side behavior which I think needs to have some custom error messaging.

ada-globus:
> There's a minor usage issue which I flagged for some internal discussion, but which exceeds the bounds of this body of work. `globus flows run show-logs` shows an unintuitive error if the log is expired/missing. The result at first may make it seem like `run list` and `run show-logs` are in disagreement, but this is all a service-side behavior which I think needs to have some custom error messaging.

For posterity, just capturing the thread from Slack here: A concurrent, ongoing workstream will soon bring the timelines for run expiration into agreement within the service, so these errors should become consistent and significantly less surprising everywhere users interact with them!

PR #839, opened 2023-08-10
Add support for `flows list --orderby`

sirosen:
There is a slightly extraneous improvement to the click type checker, but one which I think is worth keeping because it's so small.
I originally wrote this using `click.Tuple`, but then it didn't have the autocompletion behavior I wanted, so I changed tack.

This implements a `ColonDelimitedChoiceTuple` type (naming, I know! :grimacing: ) which allows for options of the form
```python
@click.option(
    "--foo",
    type=ColonDelimitedChoiceTuple(("a", "b"), ("c", "d")),
    help="Use '--foo a:c' or '--foo b:d' or '--foo a:d' or '--foo b:c'"
)
def bar(*, foo: tuple[Literal["a", "b"], Literal["c", "d"]]): ...
```
Note that this is a matrix (product of vectors) rather than a zip.

Colons have precedent in our CLI for the EndpointPlusPath type (`<id>:<path>`) and are not interpreted by shells in the way that some other fun delimiters like `|` might be. Commas were considered off-limits because they appear in the API-side of this, and could therefore confuse more advanced users.

Once we have that parsing completely locked down, all that remains is
- setup `--orderby` to use it for a matrix of parameters over `ASC` and `DESC`
- make that a `multiple=True` option so that it can be listified
- temporarily remove some values from the `--orderby` choices (service-side support is missing)
- add a test case


src/globus_cli/commands/flows/list.py
```diff
@@ -16,6 +16,20 @@
...
+    "scope_string",
+    # These names are the legacy ones in the service:
+    #       "created_by",
+    #       "administered_by",
+    # These are the new names but are not supported at time of writing:
+    #       "flow_owners",
+    #       "flow_administrators",
+    "created_at",
+    "title",
+    "created_at",
```
kurtmckee, line 30: `created_at` appears twice here.

kurtmckee (changes requested):

ada-globus (commented):
This looks good to me, once @kurtmckee's note is resolved!

src/globus_cli/commands/flows/list.py
sirosen: No it doesn't! What version are you looking at? Must be at your end because [the file looks perfect to me!](https://github.com/globus/globus-cli/pull/839/commits/027f1d70a37384037f9041274250fc276f8998b0)

(Thanks for spotting this. Probably a whoopsie from refactoring this too many times.)

ada-globus (approved):

PR #840, opened 2023-08-11
Check more exhaustively for interactive usage

ada-globus:
image

# Context
The apparent root issue is that starship does not use `PS1`

# Notes
- Only additive, should only catch more cases than before
- Tests appear to fail specifically in Windows, so unsure if this causes a regression there for non-interactive use (in which case we can have this check also assert the shell is not windows) or if this is non-worrisome, in which case perhaps modify the test...

### DIFF
--- a/src/globus_cli/termio/context.py
+++ b/src/globus_cli/termio/context.py
@@ -106,5 +106,5 @@ def term_is_interactive() -> bool:
 
     if sys.stdin.isatty():
         return True
-    
+
     return os.getenv("PS1") is not None
--- a/src/globus_cli/termio/context.py
+++ b/src/globus_cli/termio/context.py
@@ -100,4 +100,8 @@ def term_is_interactive() -> bool:
             return env
     except ValueError:
         pass
+
+    if sys.stdin.isatty():
+        return True
+
     return os.getenv("PS1") is not None
--- a/changelog.d/20230811_075351_ada_check_even_harder_for_interactive.md
+++ b/changelog.d/20230811_075351_ada_check_even_harder_for_interactive.md
@@ -1,0 +1,4 @@
+### Enhancements
+
+* More exhaustively check for interactive usage
+  (detect sessions with alternative prompts).


kurtmckee (approved):

kurtmckee (commented):
Windows tests broke. :cry:

src/globus_cli/termio/context.py
```diff
@@ -94,10 +95,12 @@ def env_interactive() -> bool | None:
...
 def term_is_interactive() -> bool:
-    try:
+    with suppress(ValueError):
         env = env_interactive()
         if env is not None:
             return env
-    except ValueError:
-        pass
+
+    if sys.__stdin__.isatty():
```
sirosen, line 103: This isn't strictly necessary, but I think using the non-dunder name here might better match any expectations we would have when mocking or otherwise testing this.

```suggestion
    if sys.stdin.isatty():
```

Both `__stdin__` and `stdin` are documented, and `sys.stdin is sys.__stdin__` is true unless `stdin` has been rewritten, so it hardly makes a difference...

But I would be curious to see if this fixes the pytest-under-windows issue. They may be rewriting `stdin` but not `__stdin__` in that scenario.

Worst case, we can go to the failing test and add a call to monkeypatch to set `isatty` to `lambda: True` on the necessary object.

src/globus_cli/termio/context.py
```diff
@@ -94,10 +95,12 @@ def env_interactive() -> bool | None:
 
 
 def term_is_interactive() -> bool:
-    try:
+    with suppress(ValueError):
         env = env_interactive()
         if env is not None:
             return env
-    except ValueError:
-        pass
```
sirosen, line 102: If you're alright doing so, would you revert this change? `env_interactive` uses distutils so we need to fix it -- and I plan to have it return `None` rather than raise. I'll do that right now, so leaving it unchanged here will avoid a conflict.

sirosen (commented):
:+1: to using `stdin.isatty()` as an initial check. I have two quick notes though -- one about a related change and one which _might_ fix the CI failure.

src/globus_cli/termio/context.py
ada-globus: Yep, I saw an approved the other PR and prefer the change there 👍

src/globus_cli/termio/context.py
ada-globus: I'm seeing some Windows tests passing, so it looks like your hunch was right! Great call 🚀

sirosen (approved):
GitHub is claiming there's a conflict, but that surprises me a little. I guess the edits are still too close together 😿
Anyway, you could rebase or do a manual merge to resolve. Whichever works.

kurtmckee (approved):

PR #843, opened 2023-08-14
Clean setuptools build artifacts before running tests

kurtmckee:
This resolves test suite failures caused by lowered code coverage numbers. The failure can occur under these circumstances:

* New Python files are introduced in one git branch
* The test suite is run, caching the new files in the `build/` directory
* A new git branch is switched to, which doesn't contain the new Python files
* The test suite is run

Under this circumstance, the cached Python files may persist in the wheel that is built, resulting in dead code that coverage detects as un-run across each test environment.

This is a force multiplier, resulting in (for the file that triggered this bug) 25 lines being multiplied across all six (6) Python test environments.

This PR resolves this problem by running `python setup.py clean --all`, which wipes out the `build/` directory's cached files.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,6 @@
 [tox]
 envlist =
-    cov-clean
+    clean
     py{311,310,39,38,37}
     py37-mindeps
     cov-combine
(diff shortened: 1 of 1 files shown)


sirosen (approved):
I was so busily trying to find a way to make setuptools or tox do this for us, I didn't see this solution which now seems like it's staring us in the face.
Great job cutting through the noise!

ada-globus (approved):
Nice find!!

PR #848, opened 2023-08-17
Align all Timer command help text to use periods

kurtmckee:
Prior to this change, sentences inconsistently end with periods.

This was found while reading the generated HTML documentation.


sirosen (approved):
I wonder if we can write a lint for this. I'm not quite sure how, but it's pretty easy to pull help from all options. I just don't know what we should do to check those strings exactly (ending in `.` seems too strict).

kurtmckee:
_Rebased on `main` to resolve merge conflicts._

PR #849, opened 2023-08-17
default transfer --recursive to None for type auto-detection

aaschaer:
For: https://app.shortcut.com/globus/story/26492/cli-default-transfer-s-recursive-option-to-none-to-properly-use-type-auto-detection

### DIFF
--- a/src/globus_cli/parsing/shared_options/transfer_task_options.py
+++ b/src/globus_cli/parsing/shared_options/transfer_task_options.py
@@ -42,12 +42,13 @@ def sync_level_option(
 
 def transfer_recursive_option(f: C) -> C:
     return click.option(
-        "--recursive",
+        "--recursive/--no-recursive",
         "-r",
         is_flag=True,
         default=None,
-        help="SOURCE_PATH and DEST_PATH are both directories, "
-        "do a recursive directory transfer",
+        help="Specify these paths are for directories and should be transferred "
+        "recursively or for files that must not be transferred recursively. "
+        "Omit this option to use path type auto-detection.",
     )(f)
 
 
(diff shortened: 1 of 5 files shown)


src/globus_cli/parsing/shared_options/transfer_task_options.py
```diff
@@ -42,12 +42,13 @@ def sync_level_option(
...
-        "--recursive",
+        "--recursive/--no-recursive",
         "-r",
         is_flag=True,
         default=None,
-        help="SOURCE_PATH and DEST_PATH are both directories, "
-        "do a recursive directory transfer",
+        help="Specify these paths are for directories and should be transferred "
+        "recursively or for files that must not be transferred recursively. "
+        "Omit this option to use path type auto-detection.",
```
kurtmckee, line 51:
```suggestion
        help=(
            "Specify these paths are for directories and should be transferred "
            "recursively or for files that must not be transferred recursively. "
            "Omit this option to use path type auto-detection."
        ),
```

kurtmckee (changes requested):
This looks great! I have one request, and that's to wrap an implicitly-concatenated string in parentheses so subsequent lines aren't indented at the same level as the function keyword arguments.

sz3:
I patched this onto my local `globus-cli==3.16.0`, and it works (as expected!) against the upcoming transfer release bits.

On that note, it's be nice to have this available next week...

kurtmckee (approved):

sirosen (approved):
I'm late to the party here, but this looks like a good fix.

I'm sure that when Transfer introduced the autodetection behavior, we addressed in the SDK and simply forgot to check the CLI.

sirosen:
I expect that there's only a very narrow set of users who are interested in following the details of this, and those are precisely the users who are already watching the Transfer API and SDK changes in this space (so they're somewhat savvy about the issue).

So I think the current text is okay, but I'm definitely not averse to further changelog tweaking.

PR #853, opened 2023-08-24
Fix a parsing bug when erasing an endpoint subscription ID

kurtmckee:
### Bugfixes

* Fix a bug that prevented running `globus endpoint set-subscription-id ... null`.


sirosen (approved):
There are alternative ways of doing this, but nothing _better_ than this as far as I know. :+1:

PR #856, opened 2023-08-26
Convert param type checker to click-type-test

sirosen:
This is a lift-and-shift of our existing functionality in this repo into a public and open-source alpha project.

The logic being isolated means that globus-cli no longer bears responsibility for testing it internally either, so the tests which were a self-test of the param type checker have been migrated into the new project and eliminated.

---

The `click-type-test` repo is [here](https://github.com/sirosen/click-type-test).
There are some gaps in the documentation, but most of what we're doing with it is covered.

---

I think there's an obvious question of whether or not this is a necessary and useful improvement.
For `globus-cli`, it frankly is not. The out-of-tree code is harder to maintain and subject to wider compatibility concerns if it gets uptake. Managing changes to both sources will require extra thought now that they aren't in a single tree.

However, I have several `click` CLI applications I work on or with both inside and outside of `globus`'s domain.
Exposing the type checker and making it available more broadly opens up improvements to other tools (Globus-maintained and otherwise) and allows those usages to feed back into `click-type-test` to improve it.


tests/unit/test_click_annotations.py
```diff
@@ -44,7 +44,30 @@ def _command_id_fn(val):
...
+
+
+@pytest.mark.skipif(sys.version_info < (3, 10), reason="test requires py3.10+")
+@pytest.mark.parametrize(
+    "param_type", _ALL_CUSTOM_PARAM_TYPES, ids=lambda x: x.__name__
+)
+def test_custom_param_types_are_annotated(param_type):
+    from click_type_test import AnnotatedParamType
+
+    assert isinstance(param_type, AnnotatedParamType)
```
kurtmckee, line 65: I think that testing for a specific Python version is slightly incorrect here.

What matters is that `click_type_test` is importable, not what version of Python the tests are running under. They're linked conditions, but I think it will be better to cut to the heart of the matter: "Is `click_type_test` installed?"

```suggestion
@pytest.mark.parametrize(
    "param_type", _ALL_CUSTOM_PARAM_TYPES, ids=lambda x: x.__name__
)
def test_custom_param_types_are_annotated(param_type):
    click_type_test = pytest.importorskip("click_type_test")

    assert isinstance(param_type, click_type_test.AnnotatedParamType)
```

This feedback applies to the next test function, too.

kurtmckee (changes requested):

tests/unit/test_click_annotations.py
sirosen: I didn't know about importorskip! I'll look into exactly how I'd like to phrase this -- I'd sort of like a module level skip if possible -- but I agree with you about the boundary between checking that click-type-test is available and checking the python version.

kurtmckee: I _think_ that you can use `importorskip` as a module-level check, but I'll watch for the next update!

sirosen: It works quite well at the module-level, though not (IMO) perfectly.
We get a single "skip" reported in the text test report, but there's no specific skipped test output with `-v`. So if someone were trying to find which tests got skipped in a given run, it could get a little confusing.

Still, I like that we get to skip the _whole_ module this way, so if we wanted a future in which `click-type-test` helped us build test parameters, that would be very possible.

kurtmckee (approved):

PR #857, opened 2023-08-28
Introduce `--orderby` for `globus ls`

sirosen:
Using the colon-delimited choice paramtype developed for `flows list --orderby`, it is now relatively trivial to add in `--orderby` support for `ls` and mark it as mutex with `--recursive`.

Tests check that the mutex setting is enforced correctly and that a simple `--orderby` usage is encoded correctly into a query param.


ada-globus (approved):
Really like how this came out!

sirosen:
I think the hard part was already solved in `ColonDelimitedChoiceTuple`!
Well, the hard parts other than coming up with good pithy names. :grin:

PR #860, opened 2023-08-29
Fix some bugs

kurtmckee:
### Bugfixes

*   Make `--no-recursive` and `--batch` mutually exclusive options.

    This affects the `globus transfer` and the `globus timer create transfer` commands.

    Previously, `--no-recursive` was not caught because of its internal representation.
    This is now fixed, so `--no-recursive` cannot be combined with `--batch`.

*   Fix a bug that caused `--batch` input files to default to non-recursive transfers.

    This affects the `globus transfer` and `globus timer create transfer` commands.

    It's now possible for `--batch` input files to use `--recursive` to enable recursive transfers
    or `--no-recursive` to explicitly disable recursive transfers.
    If neither option is specified, path type auto-detection    will determine the correct recursion setting.

### DIFF
--- a/changelog.d/20230828_142533_kurtmckee_fix_some_bugs.md
+++ b/changelog.d/20230828_142533_kurtmckee_fix_some_bugs.md
@@ -1,0 +1,8 @@
+### Bugfixes
+
+*   Make `--no-recursive` and `--batch` mutually exclusive options.
+
+    This affects the `globus transfer` and the `globus timer create transfer` commands.
+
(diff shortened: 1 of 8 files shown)


sirosen (approved):
The src/ tree changes turn out to be quite modest in size, which makes this easy to review and approve.

Thanks for all of the work on the tests! This covers some ground which -- as we discussed -- I was surprised and disappointed to learn we weren't already testing.

PR #869, opened 2023-09-12
Warn/hint on inactive runs when showing logs

sirosen:
When a `globus flows run show-logs` is run, the run will be checked to
see if its status is "INACTIVE".

This allows the CLI to print a useful hint to stderr (interactive mode
only) showing that the run log will not gain more entries until the
run is resumed.

This is an enhancement for interactive usage when a user is trying to
manually poll or "tail" the run-log and expects to see new data.
Because the log itself does not include the transition to "INACTIVE",
this usage leads to a delay between the run going inactive and the
user understanding that the state of the run has changed.

The hint is printed in a non-yellow color, and "bright_blue" was
selected from the options as a neutral color WRT the usual associated
semantics of red/yellow/green.

### DIFF
--- a/src/globus_cli/commands/flows/run/show_logs.py
+++ b/src/globus_cli/commands/flows/run/show_logs.py
@@ -89,8 +89,10 @@ def show_logs_command(
 
     if run_doc["status"] == "INACTIVE":
         print_command_hint(
-            "\nNOTE: This run is INACTIVE. "
-            "No further logs will be added until it is resumed.",
+            (
+                "\nNOTE: This run is INACTIVE. "
+                "No further logs will be added until it is resumed."
+            ),
             color="bright_blue",
         )
 
--- a/src/globus_cli/commands/flows/run/show_logs.py
+++ b/src/globus_cli/commands/flows/run/show_logs.py
@@ -49,6 +49,9 @@ def show_logs_command(
 
     flows_client = login_manager.get_flows_client()
 
+    # get the Flow, to check INACTIVE status below
+    run_doc = flows_client.get_run(run_id)
+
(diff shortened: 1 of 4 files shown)


src/globus_cli/commands/flows/run/show_logs.py
```diff
@@ -84,6 +87,13 @@ def show_logs_command(
             entry_iterator, fields=fields, json_converter=entry_iterator.json_converter
         )
 
+    if run_doc["status"] == "INACTIVE":
+        print_command_hint(
+            "\nNOTE: This run is INACTIVE. "
+            "No further logs will be added until it is resumed.",
```
kurtmckee, line 93: I find the implicit string concatenation confusing in the presence of additional function arguments.

Would it be an option to enclose this in parentheses so the implicit string concatenation isn't indented to match the second function argument that appears on the third line?

kurtmckee (changes requested):

src/globus_cli/commands/flows/run/show_logs.py
sirosen: I just tried out my new alias for putting in a `Co-authored-by` line with this fix. Let me know if you'd rather squash it, but I've made the tweak.

My brain has been sufficiently altered by `black` at this point that if something is not a requirement enforced by the formatter, I'm not sure if I have a stylistic preference. :joy: 
Looking more carefully, I agree that the parens improve readability.

kurtmckee: Crud. Adding me as `Co-authored-by` runs afoul of my 'verify signatures' configuration.

image

I don't think there's a solution for this, but I appreciate the hat-tip!

kurtmckee (approved):

PR #870, opened 2023-09-15
Remove double space in a help message

bryant1410:


derek-globus (approved):
Thanks for the fix, this'll go out in the next cli release.

PR #873, opened 2023-09-26
Add support for pause and resume operations

ada-globus:
image

Adds support for pausing and resuming timers, and (drafting off the prior art in `flows run resume`) inspection of the user's consent forest to determine if required consents are present.

### DIFF
--- a/src/globus_cli/commands/timer/resume.py
+++ b/src/globus_cli/commands/timer/resume.py
@@ -1,0 +1,87 @@
+from __future__ import annotations
+
+import typing as t
+import uuid
+
+import click
+import globus_sdk
+
+from globus_cli.login_manager import LoginManager, read_well_known_config
+from globus_cli.parsing import command
+from globus_cli.termio import TextMode, display
+from globus_cli.utils import CLIAuthRequirementsError
+
+# NB: GARE parsing requires other SDK components and therefore needs to be deferred to
+# avoid the performance impact of non-lazy imports
+if t.TYPE_CHECKING:
+    from globus_sdk.experimental.auth_requirements_error import (
+        GlobusAuthRequirementsError,
+    )
+
+
+@command("resume", short_help="Resume a timer")
+@click.argument("JOB_ID", type=click.UUID)
+@click.option(
+    "--skip-inactive-reason-check",
+    is_flag=True,
+    default=False,
+    help='Skip the check of the timer\'s "inactive reason", which is used to determine '
+    "if additional steps are required to successfully resume the timer.",
+)
+@LoginManager.requires_login("timer")
+def resume_command(
+    login_manager: LoginManager, *, job_id: uuid.UUID, skip_inactive_reason_check: bool
+) -> None:
+    """
+    Resume a timer.
+    """
+    timer_client = login_manager.get_timer_client()
+    job_doc = timer_client.get_job(job_id)
+
+    if not skip_inactive_reason_check:
+        gare = _get_inactive_reason(job_doc)
+        if gare is not None and gare.authorization_parameters.required_scopes:
+            if not _has_required_consent(
+                login_manager, gare.authorization_parameters.required_scopes
+            ):
+                raise CLIAuthRequirementsError(
+                    "This run is missing a necessary consent in order to resume.",
+                    required_scopes=gare.authorization_parameters.required_scopes,
+                )
+
+    resumed = timer_client.resume_job(
+        job_id,
+        update_credentials=(not skip_inactive_reason_check),
+    )
+    display(
+        resumed,
+        text_mode=TextMode.text_raw,
+        simple_text=resumed["message"],
+    )
+
+
+def _get_inactive_reason(
+    job_doc: dict[str, t.Any] | globus_sdk.GlobusHTTPResponse
+) -> GlobusAuthRequirementsError | None:
+    from globus_sdk.experimental.auth_requirements_error import (
+        to_auth_requirements_error,
+    )
+
+    if not job_doc.get("status") == "inactive":
+        return None
+
+    reason = job_doc.get("inactive_reason", {})
+    if reason.get("cause") != "globus_auth_requirements":
+        return None
+
+    return to_auth_requirements_error(reason.get("detail", {}))
+
+
+def _has_required_consent(
+    login_manager: LoginManager, required_scopes: list[str]
+) -> bool:
+    auth_client = login_manager.get_auth_client()
+    user_data = read_well_known_config("auth_user_data", allow_null=False)
+    user_identity_id = user_data["sub"]
+    consents = auth_client.get_consents(user_identity_id)
+    return consents.contains_scopes(required_scopes)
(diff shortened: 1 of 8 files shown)


src/globus_cli/commands/timer/resume.py
```diff
@@ -0,0 +1,87 @@
...
+                login_manager, gare.authorization_parameters.required_scopes
+            ):
+                raise CLIAuthRequirementsError(
+                    "This run is missing a necessary consent in order to resume.",
+                    required_scopes=gare.authorization_parameters.required_scopes,
+                )
+
+    resumed = timer_client.resume_job(
+        job_id,
+        update_credentials=(not skip_inactive_reason_check),
```
sirosen, line 54: Should we perhaps add a dedicated option for this, rather than deducing it from `--skip-inactive-reason-check`? e.g. `--update-credentials/--no-update-credentials`?

I know we want a more nuanced default behavior for the service (in which case we shouldn't ever have to / want to pass this), but I'm trying to reason through what the CLI behavior will look like absent that _and_ once it's introduced.
I get the connection you're drawing here -- if you pass `--skip-inactive-reason-check`, it's very unlikely that you _do_ want us to update creds and skip the check -- but is the situation improved by having a dedicated flag?

Today, `--no/-update-credentials` would give explicit control with a default of omission (meaning "service default"). If the service adds the better default behavior, we naturally inherit it. Yes, it's slightly more awkward to use correctly today, but it means not having to come back to this to "cleanup" when the service changes.

src/globus_cli/commands/timer/resume.py
```diff
@@ -0,0 +1,87 @@
...
+    Resume a timer.
+    """
+    timer_client = login_manager.get_timer_client()
+    job_doc = timer_client.get_job(job_id)
+
+    if not skip_inactive_reason_check:
+        gare = _get_inactive_reason(job_doc)
+        if gare is not None and gare.authorization_parameters.required_scopes:
+            if not _has_required_consent(
+                login_manager, gare.authorization_parameters.required_scopes
```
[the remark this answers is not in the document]
ada-globus: They will! Def don't want this to be the responsibility of clients.

https://github.com/globusonline/globus-timers/blob/b235552f83325b871945c6164740a3c33096fa96/src/globus_timers/db/models.py#L924-L931

src/globus_cli/commands/timer/resume.py
ada-globus: 1) Agree that at _some point_ we may want the explicit credential update flags and to remove the local check entirely in favor of the service telling us directly.
2) I don't have a super strong POV, so I think I can be easily persuaded if you do.
3) I think the combination of the flags could be a little weird. Should `--no-update-credentials` imply `--skip-inactive-reason-check`? Or would you need to specify both if you didn't want that behavior? I think the former maybe should imply the latter from a UX POV, or otherwise we're asking users to understand the internal details of the code path that it's walking. But maybe I am worrying too much and the whole thing is sort of "advanced" behavior anyway that we shouldn't expect most users to think about or use—though in that case we might want to tell them as much.

sirosen: Circling back on this over the weekend, here's what I think we should do, and why:

**What?** Make `update_credentials` match the (desired) future server-default behavior, `true` if there is a GARE and `false` otherwise.
- That means we always do `gare = _get_inactive_reason(job_doc)`, even outside of the inactive reason check
- Users cannot specify `update_credentials` via a CLI option (yet?)
- `skip_inactive_reason_check` is not involved in this process at all -- it is fully orthogonal to this behavior

**Why?** Consistent behavior across future upgrades and across CLI commands.
- When the service adds this as the default behavior (or something close to this), we can drop the parameter from the CLI and the behaviors will remain self-consistent.
- `--skip-inactive-reason-check` therefore only controls one thing, which is whether or not the command will *refuse* to run if there's an unsatisfied GARE detected. That matches the behavior for Flows pretty closely.
- Adding any explicit control for this with `--update-credentials` puts us in a worse position vis-a-vis usability. We already know what the "best" default behavior is within the current paradigm of behavior; this is just pushing it into the client (temporarily) while we get it into the service.

src/globus_cli/commands/timer/resume.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+@command("resume", short_help="Resume a timer")
+@click.argument("JOB_ID", type=click.UUID)
+@click.option(
+    "--skip-inactive-reason-check",
+    is_flag=True,
+    default=False,
+    help='Skip the check of the timer\'s "inactive reason", which is used to determine '
+    "if additional steps are required to successfully resume the timer.",
```
kurtmckee, line 29:
```suggestion
    help=(
        'Skip the check of the timer\'s "inactive reason", which is used to determine '
        'if additional steps are required to successfully resume the timer.'
    ),
```

src/globus_cli/commands/timer/resume.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def _get_inactive_reason(
+    job_doc: dict[str, t.Any] | globus_sdk.GlobusHTTPResponse
+) -> GlobusAuthRequirementsError | None:
+    from globus_sdk.experimental.auth_requirements_error import (
+        to_auth_requirements_error,
+    )
+
+    if not job_doc.get("status") == "inactive":
```
kurtmckee, line 70: I suggest this form, which matches the form used on line 74:

```suggestion
    if job_doc.get("status") != "inactive":
```

Could you also add a test that will trigger this conditional branch? This is the only code that isn't tested and you're right on the cusp of 100% coverage for this file. :crossed_fingers: :grinning:

kurtmckee (changes requested):

src/globus_cli/commands/timer/resume.py
ada-globus: Lol, I _just_ finished going the exact opposite direction. Reading this and digesting.

sirosen: Sorry! :crying_cat_face: 

Still, I think most of your testing work would be portable to this approach as well. i.e. The facade will change if you buy my reasoning, but the guts will stay the same.

kurtmckee (approved):

src/globus_cli/commands/timer/resume.py
ada-globus: All good :)

I backed out all of the changes. `--skip-inactive-reason-check` now merely suppresses the error, and `update_credentials` is explicitly true when a GARE is present. Realizing now that in a slight deviation from your suggestion, when a GARE is not present, I have it pass `None` which will fall back to the service's default of `False`. I modestly prefer it to defer to the service (which would allow the service to impose additional conditions that might update the credentials without needing to change the CLI first) but functionally this doesn't make a material difference afaict today, so I don't care all that much. I can change it to explicitly send `False` if you have a stronger preference than I.

sirosen: Nah, I'm happy with that course of action. I'll double check the changes and probably approve momentarily!

sirosen (approved):

src/globus_cli/commands/timer/resume.py
ada-globus: 🙏

src/globus_cli/commands/timer/resume.py
ada-globus: 🤦‍♀️

PR #878, opened 2023-10-09
Fix the behavior of warnings in tests and resolve the revealed warnings

sirosen:
This is a series of commits which each does a well-defined thing on its own, but in concert they "enable warnings and fix them".

1. Turn on warnings in the tests.

Although we had `filterwarnings = error` set, the CLI assumes control of warnings via `simplefilter` to suppress warnings in the user-facing context.
As a result, `filterwarnings` was ineffective.
The fix I've applied is a wrapper module `globus_cli._warnings` which lets `pytest_configure` reach in and turn off `simplefilter` altogether.
This makes the tests fail because warnings are now being emitted in the tests.

2. Fix timedelta formatting

We wrote the output formatters to be robust to failed data parsing and to fallback to `str(data)` with a warning.
The `globus timer` command tree had a formatter for `datetime.timedelta` which was checking for `isinstance(data, int)` rather than `isinstance(data, (int, float))`.
Good thing our mock data was pulled from real API responses and not synthesized by hand to match our expectations -- we now have a bugfix for timedelta rendering.

This is the only item which I deemed worthy of a changelog entry.

3. GlobusAPIError.raw_text -> GlobusAPIError.text

Seen in `globus api` commands; we deprecated `raw_text` in favor of `text`.
Fix!

4. Fix deprecated `jmespath` literal in tests

This one is a `PendingDeprecationWarning`, so I don't know what the timeline is yet.
`jmespath` used to use or at least accept backticks for strings in expressions.
It now wants single quotes.
I'm concerned that this will impact the validity of some of our CLI examples over time, but for now I'm setting that issue aside as we fix the tests.

5. Handle `oauth2_userinfo` deprecation in SDK `main`

Finally, the driver for this adventure.
`oauth2_userinfo()` is deprecated in SDK `main` in favor of `userinfo()`.
It therefore *should have* warned, and after (1) it did!

There are two easy paths to resolution:
- ignore the warning for now
- patch our customized AuthClient to support `userinfo` early

I chose the second one, since either way involves removing something from this patch later (either compat code or the warning ignore).
This front-loads the work, making the later cleanup simpler.

### DIFF
--- a/changelog.d/20231009_093456_sirosen_fix_test_warnings.md
+++ b/changelog.d/20231009_093456_sirosen_fix_test_warnings.md
@@ -1,0 +1,3 @@
+### Bugfixes
+
+* Fix the rendering of timedeltas in CLI output  for `globus timer` commands
--- a/src/globus_cli/_warnings.py
+++ b/src/globus_cli/_warnings.py
@@ -1,0 +1,14 @@
+import warnings
+
+# this bool turns off all warning controls, handing control of python
+# warnings to the testsuite
+# this ensures that `pytest --filterwarnings error` works
+_TEST_WARNING_CONTROL: bool = False
+
+
+def simplefilter(filterstr: str) -> None:
+    """
+    wrap `warnings.simplefilter` with a check on `_TEST_WARNING_CONTROL`
+    """
+    if not _TEST_WARNING_CONTROL:
+        warnings.simplefilter(filterstr)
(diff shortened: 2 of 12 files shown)


kurtmckee:
Test suite is failing.

sirosen:
Specifically mypy :facepalm: 
I must be getting sloppy as I age -- I always seem to miss something these days. Commit explains what.

kurtmckee:
I think mypy is just young enough to make us look old by comparison.

That whippersnapper.

PR #891, opened 2023-10-31
Add support for GARE-shaped data in timer and flow resume

sirosen:
This updates `globus timer resume` and `globus flows run resume` to handle Globus Auth Requirements Error formatted data in the inactive reason fields for a **timer** or a **run**.

The changes here both improve internal infrastructure and support for GARE data and use that improved infrastructure to handle GAREs which ar encountered in Timers and Flows.

In terms of significant improvements to infrastructure:
- `CLIAuthRequirementsError` now contains a GARE
- The session-error hook is now restructured to mirror the consent-error hook
- `CLIAuthRequirementsError` handling now supports session errors

And the feature changes are the same in each of the resume commands:
- Always check GARE for session errors, not just consent errors
- Only enable the "smooth resume" behavior for consent handling cases; always trigger an error when handling session errors
- When handling session errors, instruct the user (via `CLIAuthRequirementsError.epilog`) to return and run the command with `--skip-inactive-reason-check` in order to bypass the session-reading behavior

---

- Update internal auth errors to carry a GARE
- Allow 'resume's to to raise session errors
- Handle session GAREs in hooks
- Add 'timer resume' easter egg
- Simplify exit code paths in exceptions hooks
- Add an epilog to GARE handling
- Cleanup timer resume tests
- Add tests for 'timer resume' on session errors
- Add tests for `flows run resume` GARE handling
- Add changelog for new GARE handling

### DIFF
--- a/src/globus_cli/commands/flows/run/resume.py
+++ b/src/globus_cli/commands/flows/run/resume.py
@@ -93,7 +93,8 @@ def check_inactive_reason(
     # (consents may be present without session requirements met)
     if unhandled_requirements:
         raise CLIAuthRequirementsError(
-            "This run needs strong authentication in order to resume.",
+            "This run has additional authentication requirements that must be met "
+            "in order to resume.",
             gare=gare,
             epilog=textwrap.dedent(
                 f"""\
--- a/src/globus_cli/commands/flows/run/resume.py
+++ b/src/globus_cli/commands/flows/run/resume.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import textwrap
 import typing as t
 import uuid
 
@@ -43,10 +44,9 @@ def resume_command(
 
     specific_flow_client = login_manager.get_specific_flow_client(flow_id)
 
-    gare: GlobusAuthRequirementsError | None = None
+    gare = _get_inactive_reason(run_doc)
     if not skip_inactive_reason_check:
-        gare = _get_inactive_reason(run_doc)
-        check_inactive_reason(login_manager, gare)
+        check_inactive_reason(login_manager, run_id, gare)
 
     fields = [
         Field("Run ID", "run_id"),
@@ -64,6 +64,7 @@ def resume_command(
 
 def check_inactive_reason(
     login_manager: LoginManager,
+    run_id: uuid.UUID,
     gare: GlobusAuthRequirementsError | None,
 ) -> None:
     if gare is None:
@@ -74,7 +75,7 @@ def check_inactive_reason(
         )
         if consent_required:
             raise CLIAuthRequirementsError(
-                "This timer is missing a necessary consent in order to resume.",
+                "This run is missing a necessary consent in order to resume.",
                 gare=gare,
             )
 
@@ -92,8 +93,15 @@ def check_inactive_reason(
     # (consents may be present without session requirements met)
     if unhandled_requirements:
         raise CLIAuthRequirementsError(
-            "This timer needs strong authentication in order to resume.",
+            "This run needs strong authentication in order to resume.",
             gare=gare,
+            epilog=textwrap.dedent(
+                f"""\
+                After updating your session, resume the timer with
+
+                    globus flows run resume --skip-inactive-reason-check {run_id}
+                """
+            ),
         )
 
 
--- a/src/globus_cli/commands/flows/run/resume.py
+++ b/src/globus_cli/commands/flows/run/resume.py
@@ -51,7 +51,7 @@ def resume_command(
             ):
                 raise CLIAuthRequirementsError(
                     "This run is missing a necessary consent in order to resume.",
-                    required_scopes=gare.authorization_parameters.required_scopes,
+                    gare=gare,
                 )
 
     fields = [
--- a/src/globus_cli/commands/flows/run/resume.py
+++ b/src/globus_cli/commands/flows/run/resume.py
@@ -43,16 +43,10 @@ def resume_command(
 
     specific_flow_client = login_manager.get_specific_flow_client(flow_id)
 
+    gare: GlobusAuthRequirementsError | None = None
     if not skip_inactive_reason_check:
         gare = _get_inactive_reason(run_doc)
-        if gare is not None and gare.authorization_parameters.required_scopes:
-            if not _has_required_consent(
-                login_manager, gare.authorization_parameters.required_scopes
-            ):
-                raise CLIAuthRequirementsError(
-                    "This run is missing a necessary consent in order to resume.",
-                    gare=gare,
-                )
+        check_inactive_reason(login_manager, gare)
 
     fields = [
         Field("Run ID", "run_id"),
@@ -66,6 +60,41 @@ def resume_command(
 
     res = specific_flow_client.resume_run(run_id)
     display(res, fields=fields, text_mode=TextMode.text_record)
+
+
+def check_inactive_reason(
+    login_manager: LoginManager,
+    gare: GlobusAuthRequirementsError | None,
+) -> None:
+    if gare is None:
+        return
+    if gare.authorization_parameters.required_scopes:
+        consent_required = not _has_required_consent(
+            login_manager, gare.authorization_parameters.required_scopes
+        )
+        if consent_required:
+            raise CLIAuthRequirementsError(
+                "This timer is missing a necessary consent in order to resume.",
+                gare=gare,
+            )
+
+    # at this point, the required_scopes may have been checked and satisfied
+    # therefore, we should check if there are additional requirements other than
+    # the scopes/consents
+    unhandled_requirements = set(gare.authorization_parameters.to_dict().keys()) - {
+        "required_scopes",
+        # also remove 'message' -- not a 'requirement'
+        "session_message",
+    }
+    # reraise if anything remains after consent checking
+    # this ensures that we will reraise if we get an error which contains
+    # both required_scopes and additional requirements
+    # (consents may be present without session requirements met)
+    if unhandled_requirements:
+        raise CLIAuthRequirementsError(
+            "This timer needs strong authentication in order to resume.",
+            gare=gare,
+        )
 
 
 def _get_inactive_reason(
--- a/src/globus_cli/commands/flows/run/resume.py
+++ b/src/globus_cli/commands/flows/run/resume.py
@@ -98,7 +98,7 @@ def check_inactive_reason(
             gare=gare,
             epilog=textwrap.dedent(
                 f"""\
-                After updating your session, resume the timer with
+                After updating your session, resume the run with
 
                     globus flows run resume --skip-inactive-reason-check {run_id}
                 """
--- a/src/globus_cli/commands/flows/run/resume.py
+++ b/src/globus_cli/commands/flows/run/resume.py
@@ -82,7 +82,7 @@ def check_inactive_reason(
     # at this point, the required_scopes may have been checked and satisfied
     # therefore, we should check if there are additional requirements other than
     # the scopes/consents
-    unhandled_requirements = set(gare.authorization_parameters.to_dict().keys()) - {
+    unhandled_requirements = set(gare.authorization_parameters.to_dict()) - {
         "required_scopes",
         # also remove 'message' -- not a 'requirement'
         "session_message",
--- a/src/globus_cli/commands/timer/resume.py
+++ b/src/globus_cli/commands/timer/resume.py
@@ -20,41 +20,6 @@ if t.TYPE_CHECKING:
     )
 
 
-def _full_speed_ahead(ctx: click.Context, param: click.Parameter, value: bool) -> None:
-    # this easter egg has no effects other than printing a message
-    # this choice was made to ensure that there are no important behaviors which are
-    # impacted by it, or which *require* its use
-    if value:
-        click.secho(
-            r'''
-   ___                       __   __
-  / _ \ ___ _ __ _   ___    / /_ / /  ___
- / // // _ `//  ' \ / _ \  / __// _ \/ -_)
-/____/ \_,_//_/_/_//_//_/  \__//_//_/\__/
-
-  __                           __
- / /_ ___   ____ ___  ___  ___/ /___  ___  ___
-/ __// _ \ / __// _ \/ -_)/ _  // _ \/ -_)(_-< _
-\__/ \___//_/  / .__/\__/ \_,_/ \___/\__//___/( )
-              /_/                             |/
-   ___       __ __                         __        __                 __ __
-  / _/__ __ / // /  ___  ___  ___  ___  ___/ / ___ _ / /  ___  ___ _ ___/ // /
- / _// // // // /  (_-< / _ \/ -_)/ -_)/ _  / / _ `// _ \/ -_)/ _ `// _  //_/
-/_/  \_,_//_//_/  /___// .__/\__/ \__/ \_,_/  \_,_//_//_/\__/ \_,_/ \_,_/(_)
-                      /_/
-
- .  o ..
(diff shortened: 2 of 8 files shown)


src/globus_cli/commands/flows/run/resume.py
```diff
@@ -43,16 +43,10 @@ def resume_command(
...
         gare = _get_inactive_reason(run_doc)
-        if gare is not None and gare.authorization_parameters.required_scopes:
-            if not _has_required_consent(
-                login_manager, gare.authorization_parameters.required_scopes
-            ):
-                raise CLIAuthRequirementsError(
-                    "This run is missing a necessary consent in order to resume.",
-                    gare=gare,
-                )
+        check_inactive_reason(login_manager, gare)
```
kurtmckee: GitHub is balking at letting me reference a line of code just above this.

It looks like `specific_flow_client` is set up prior to its use in this code (unless calling `login_manager.get_specific_flow_client(flow_id)` modifies `login_manager` as a side effect).

I recommend moving this condition above that unhighlighted line since it isn't needed yet.

src/globus_cli/commands/flows/run/resume.py
```diff
@@ -68,6 +62,41 @@ def resume_command(
...
+        if consent_required:
+            raise CLIAuthRequirementsError(
+                "This timer is missing a necessary consent in order to resume.",
+                gare=gare,
+            )
+
+    # at this point, the required_scopes may have been checked and satisfied
+    # therefore, we should check if there are additional requirements other than
+    # the scopes/consents
+    unhandled_requirements = set(gare.authorization_parameters.to_dict().keys()) - {
```
kurtmckee, line 85: No need for a call to `.keys()` because `set(dict_instance)` returns a set of keys.

```suggestion
    unhandled_requirements = set(gare.authorization_parameters.to_dict()) - {
```

src/globus_cli/commands/flows/run/resume.py
```diff
@@ -68,6 +62,41 @@ def resume_command(
...
+        # also remove 'message' -- not a 'requirement'
+        "session_message",
+    }
+    # reraise if anything remains after consent checking
+    # this ensures that we will reraise if we get an error which contains
+    # both required_scopes and additional requirements
+    # (consents may be present without session requirements met)
+    if unhandled_requirements:
+        raise CLIAuthRequirementsError(
+            "This timer needs strong authentication in order to resume.",
```
kurtmckee: ~timer~ run

Separately, regarding phrasing -- I don't think the line "strong authentication" communicates what the run needs to resume.

src/globus_cli/commands/flows/run/resume.py
```diff
@@ -68,6 +62,41 @@ def resume_command(
...
+) -> None:
+    if gare is None:
+        return
+    if gare.authorization_parameters.required_scopes:
+        consent_required = not _has_required_consent(
+            login_manager, gare.authorization_parameters.required_scopes
+        )
+        if consent_required:
+            raise CLIAuthRequirementsError(
+                "This timer is missing a necessary consent in order to resume.",
```
kurtmckee: ~timer~ run

src/globus_cli/commands/timer/resume.py
```diff
@@ -62,6 +56,41 @@ def resume_command(
...
+        # also remove 'message' -- not a 'requirement'
+        "session_message",
+    }
+    # reraise if anything remains after consent checking
+    # this ensures that we will reraise if we get an error which contains
+    # both required_scopes and additional requirements
+    # (consents may be present without session requirements met)
+    if unhandled_requirements:
+        raise CLIAuthRequirementsError(
+            "This timer needs strong authentication in order to resume.",
```
kurtmckee, line 135: Same concern as above -- "strong authentication" doesn't seem descriptive nor prescriptive for resuming.

src/globus_cli/exception_handling/hooks.py
```diff
@@ -25,17 +25,31 @@ def _pretty_json(data: dict, compact=False) -> str:
...
 def handle_internal_auth_requirements(exception: CLIAuthRequirementsError) -> None:
-    if not (exception.gare and exception.gare.authorization_parameters.required_scopes):
+    gare = exception.gare
+    if not gare:
         click.secho(
             "Fatal Error: Unsupported internal auth requirements error!",
             bold=True,
             fg="red",
         )
         click.get_current_context().exit(255)
```
kurtmckee: Is exiting with 255 significant compared to exiting with some other value?

In the past I've used firmware manipulation tools that had dozens of possible error codes that each conveyed documented, nuanced information; we're not doing anything like that, but if there's a convention to use 255 for fatal errors and, say, 1 for regular ol' "didn't work" errors, I want to know that convention.

Not related to this review but wanted to surface my thoughts since I'm seeing this here.

src/globus_cli/exception_handling/hooks.py
```diff
@@ -94,14 +112,20 @@ def consent_required_hook(exception: globus_sdk.GlobusAPIError) -> None:
...
         exception.message, exception.info.consent_required.required_scopes
     )
+    if ret is not None:
+        click.get_current_context().exit(ret)
 
 
 def _concrete_consent_required_hook(
     message: str | None, required_scopes: list[str]
-) -> None:
+) -> int | None:
```
kurtmckee, line 138: This doesn't return an int -- it returns 255.

src/globus_cli/exception_handling/hooks.py
```diff
@@ -25,17 +25,31 @@ def _pretty_json(data: dict, compact=False) -> str:
...
         click.get_current_context().exit(255)
 
-    required_scopes = exception.gare.authorization_parameters.required_scopes
-
-    _concrete_consent_required_hook(exception.message, required_scopes)
+    required_scopes = gare.authorization_parameters.required_scopes
+    if required_scopes:
+        ret = _concrete_consent_required_hook(exception.message, required_scopes)
+        if ret is not None:
+            click.get_current_context().exit(ret)
```
kurtmckee: This is the reverse of the

```python
if ret is None:
    return None
```

pattern that we've discussed before. 😆 This code would be clearer if it read as:

```suggestion
        if ret is not None:
            click.get_current_context().exit(255)
```

Don't do that! But I've commented below that the return value of `_concrete_consent_required_hook` is not an arbitrary int -- it's 255, always.

src/globus_cli/commands/timer/resume.py
```diff
@@ -30,6 +65,13 @@
         "if additional steps are required to successfully resume the timer."
     ),
 )
+@click.option(
+    "--damn-the-torpedoes",
+    expose_value=False,
+    is_flag=True,
+    hidden=True,
+    callback=_full_speed_ahead,
+)
```
kurtmckee, line 75: This is fun, but I'd like it removed.

src/globus_cli/exception_handling/hooks.py
```diff
@@ -57,23 +57,37 @@ def handle_internal_auth_requirements(exception: CLIAuthRequirementsError) -> No
...
+    )
 
 
 def _concrete_session_hook(
     message: str | None,
     policies: list[str] | None,
     identities: list[str] | None,
     domains: list[str] | None,
-):
+) -> int | None:
```
kurtmckee, line 105: This function only returns `None`.

tests/functional/timer/test_job_resume.py
```diff
@@ -133,50 +225,5 @@ def _register_responses(mock_user_data):
...
-
-def test_resume_job_inactive_gare_consent_missing_but_skip_check(run_line):
-    meta = load_response_set("cli.timer_resume.inactive_gare.consents_missing").metadata
-    job_id = meta["job_id"]
-    run_line(
-        ["globus", "timer", "resume", "--skip-inactive-reason-check", job_id],
-        search_stdout=f"Successfully resumed job {job_id}.",
-    )
+def _urlscope(m: str, s: str) -> str:
+    return f"https://auth.globus.org/scopes/{m}/{s}"
```
kurtmckee, line 275: This is called twice from two adjacent lines in the same function. Would it be beneficial to replace the function calls with a string literal and an f-string?

kurtmckee (changes requested):

src/globus_cli/commands/flows/run/resume.py
sirosen: Good catches on timer/run confusion by me!

What should we call it? It's part of a larger error message which instructs the user that they need to run `globus session update FOO BAR BAZ`.
In that context, it renders like so:
```
The resource you are trying to access requires you to re-authenticate.
This timer needs strong authentication in order to resume.

Please run

    globus session update globus.org

to re-authenticate with the required identities

* * *

After updating your session, resume the timer with

    globus timer resume --skip-inactive-reason-check 35575759-e72a-4be5-be9f-7df73015382b

```

We can adjust any part of the rendering here as seems suitable, within some limits.
But this string is just parametrizing that second line to be specific to the context in which the exception was raised.

src/globus_cli/exception_handling/hooks.py
sirosen: We have some defined and documented exit codes.
1 for "normal failure" (e.g. request failure)
2 for argparsing (typical)
3 for user-facing entity type mismatches (e.g. storage-gateway list invoked on a GCP endpoint)
4 for session errors (note `exit_status=4` in the decorator)

We also reserve the range of exit statuses 50-99 for user-use, via `--map-http-status`. (You are also allowed to `--map-http-status '404=0'`.)

255 is not explicitly documented anywhere, but is used for internal/fatal errors, to ensure that in the event that we encounter a condition like this, we can exit with a status other than the ones which we're already using.

We can add it to the exit code documentation which we render on the docs site (e.g. [one command's exit code doc](https://docs.globus.org/cli/reference/api_auth/#exit_status)). I'm not sure I want it clogging up `--help` output everywhere.

src/globus_cli/exception_handling/hooks.py
sirosen: I would argue it returns any int in the range 0 to 255, inclusive, or None.
That's the interface I'm defining here.
But a range-literal isn't something that the type system provides, so `int` is most appropriate.

src/globus_cli/exception_handling/hooks.py
sirosen: It _happens_ to be `Literal[255] | None` _today_. But the defined interface is not that.

src/globus_cli/exception_handling/hooks.py
```diff
@@ -24,32 +24,47 @@ def _pretty_json(data: dict, compact=False) -> str:
...
             session_identities,
             session_domains,
         )
+        if exit_code is None:
+            exit_code = ret
+        elif ret is not None:
+            # use bitwise-or if we have two codes to handle
+            # it's guaranteed that if either is 255 the result will be 255, and it's
+            # one of the most information-preserving possible handling paths
+            exit_code |= ret
```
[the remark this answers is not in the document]
sirosen: No. I do not want to do that.
The definition here is that there are internal hooks which are allowed to indicate an exit status or None.

The fact that this hook happens not to leverage this right now doesn't change the design.

I could have defined a `Protocol` or wrapped these in some object hierarchy to artificially make the protocol explicit, but that's just so much noise relative to the interface being "returns exit code or None".

tests/functional/timer/test_job_resume.py
sirosen: I sort of think this would be a nice test utility to just have on hand in some central place.
But it's out of scope -- the only reason this line appears as changed is that I reordered file contents.

src/globus_cli/commands/timer/resume.py
```diff
@@ -20,7 +21,7 @@
 
 
 @command("resume", short_help="Resume a timer")
-@click.argument("JOB_ID", type=click.UUID)
+@click.argument("TIMER_ID", type=click.UUID)
```
ada-globus, line 24: ⏲️❤️

src/globus_cli/exception_handling/hooks.py
kurtmckee: No, not something to add to `--help` output everywhere. I just wanted to understand the conventions (and whether they're documented).

Might be good to establish constants in the future so they're used throughout.

sirosen: Yeah, I'm not averse to a follow-up which puts these all into a file like `globus_cli.constants` and then we use `exit(ECODE_FATAL_ERROR)` or something like that. That way the convention is codified in the code without us having to "just know" (the current state of the art).

src/globus_cli/exception_handling/hooks.py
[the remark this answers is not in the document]
kurtmckee: I don't see a way to override this unless the module-level function is swapped out, but this is the CLI, so that's not going to happen. `_concrete_session_hook()` exclusively returns `None` so I'm saying this code can be reverted because `ret` is `None` and this will never run.

Maybe I'm missing something?

ada-globus (approved):

src/globus_cli/exception_handling/hooks.py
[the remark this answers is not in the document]
sirosen: We talked this over to sync; and I'll share the result here:

I was concerned with making sure we had a paradigm in place for handling a series of operations to handle GAREs, with some kind of well-defined cascading failure semantics. Each step executes serially, but we only aggregate-up the results at the end. Everything we mismatched about was in service of having this kind of setup -- which skips over an important but valuable question of whether cascading through to a later failure or failing early is better, etc.

I'm going to adjust the method for handling invalid/internally-inconsistent GARE data (e.g. ConsentRequired with no scopes listed) to evaluate at the start of the error hook whether or not the data appears to be well-formed (to a higher standard than whether or not it can be parsed -- namely, does it match exactly what the CLI expects). After doing so, we'll either error with a clear fail-whale message or proceed with normal (non-erroring) handling.

src/globus_cli/commands/flows/run/resume.py
kurtmckee: Oh! In that broader context it's much clearer what's happening. Nevertheless, "strong authentication" seems non-ideal. Perhaps:

```
This timer has additional authentication requirements that must be met.
```

?

src/globus_cli/commands/flows/run/resume.py
```diff
@@ -92,8 +93,15 @@ def check_inactive_reason(
     # (consents may be present without session requirements met)
     if unhandled_requirements:
         raise CLIAuthRequirementsError(
-            "This timer needs strong authentication in order to resume.",
+            "This run needs strong authentication in order to resume.",
             gare=gare,
+            epilog=textwrap.dedent(
+                f"""\
+                After updating your session, resume the timer with
```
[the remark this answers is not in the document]
kurtmckee: I think two of my comments were resolved by a later commit. This one is still relevant, though.

src/globus_cli/commands/flows/run/resume.py
sirosen: Yeah, let's go with that and I'll plumb it so that it replaces the first line of the messaging rather than supplementing it.

sirosen: I'm going to mark this resolved by the recent changes, which should align well with this discussion.

src/globus_cli/commands/flows/run/resume.py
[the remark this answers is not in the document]
sirosen: I now realize that I missed this one because it's in the `epilog` and I was checking the other messages when I did the fixing. I just came back and got this one done too.

kurtmckee (approved):

PR #893, opened 2023-11-06
Use v2 API for Timer creation

ada-globus:
This supersedes @sirosen's PR—I can't push commits to it directly, so in the meantime, I've pushed his and my combined changes to a branch so that work can continue. His original notes are included below for reference.

As mentioned in the original notes, there is a single assertion that will fail until a new release of the Python SDK containing globus/globus-sdk-python#900 is released.

This should be held for merge until we support displaying these timers in the Web App.

> This is a work-in-progress PR.  Updating Timer creation to use the new v2 API will require that globus/globus-sdk-python#900 is released, but it will also require that all Globus products are ready to display these timers.
>
> ---
> 
> Changes basically consist of translating the current CLI options into the newer API structure. Additionally, because there is a new suite of helpers to use from the SDK, there is some relevant refinement to handling of timezone data included here.
> 
> The work was done by:
> 1. getting it working interactively using `timer_client.post(...)`
> 2. getting it ported to work interactively using `timer_client.create_timer(...)`
> 3. applying any necessary fixes to the tests
> 
> This process is not a perfect/ideal one, but reflects the fact that the first phase of development was carried out *before* the new API was released.  So it's possible that there are improvements which are possible now but which haven't been realized.

### DIFF
--- a/changelog.d/20231106_082938_ada_use_timersv2.md
+++ b/changelog.d/20231106_082938_ada_use_timersv2.md
@@ -1,0 +1,4 @@
+### Other
+
+* The `globus timer create transfer` command now supports the latest
+  version of the Globus Timers API.
--- a/setup.py
+++ b/setup.py
@@ -46,7 +46,7 @@ setup(
     package_dir={"": "src"},
     python_requires=">=3.7",
     install_requires=[
-        "globus-sdk==3.31.0",
+        "globus-sdk==3.32.0",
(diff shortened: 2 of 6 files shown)


kurtmckee (approved):

ada-globus:
Just updated this to globus-sdk 3.32.0, which I expect to clear up the failing test.

> [!Note]
> This is still on hold to sync with Web App support.

PR #897, opened 2023-11-28
Allow copied-and-pasted API routes to the `globus api groups` command

kurtmckee:
### Bugfixes

* Accommodate copied-and-pasted API routes to the `globus api groups` command  that include the `/v2` route.


sirosen (approved):
I wonder if this is a safe way for us to extend the SDK implementation as well...?

Let's add it here for now, but perhaps the right way to get rid of the base path notion is for us to implicitly handle it in this way in the SDK.

PR #898, opened 2023-11-28
Implement mapped collection create

sirosen:
This is a rebase of the prior work which I had on this, with some minimal cleanup to get things into a reasonable state. I'm opening it as a draft to improve visibility. It conflicts with the guest collection creation command and will need a rebase after that merges (so please ignore the conflicts).

I see a few things which need additional attention in here, and I especially note that there were no tests when I dusted this off. I think I had done some measure of manual testing which at this point is not very relevant.

Some areas of interest:
- This shares parameters with `collection update`, and it will share some with guest creation as well -- we should move those to common helpers where possible.
- Handling of the `policies` sub-object is tricky. The current implementation has a small collection of mutually exclusive options which can be used. (These could be declared as mutex, but I believe mutex checking of `multiple=True` options requires some care.) I would like it better if we could declare these usages separately somehow, but I think the current solution is likely to be a good pragmatic solution for the near-to-medium term.
- The commit history notes an interesting rationale for naming the option `--storage-gateway` rather than `--storage-gateway-id`. The expectation baked into this is that a typical endpoint will have only one or two gateways attached, and that typically when there are two (or more), there will only be one of each type. It is therefore a planned enhancement to support `--storage-gateway posix` as a method for selecting the only POSIX gateway on the Endpoint. I think this branch of work should be updated to either implement that logic OR switch to `--storage-gateway-id`. I'm not yet sure which.

### DIFF
--- a/src/globus_cli/commands/collection/create/__init__.py
+++ b/src/globus_cli/commands/collection/create/__init__.py
@@ -5,6 +5,7 @@ from globus_cli.parsing import group
     "create",
     lazy_subcommands={
         "guest": (".guest", "collection_create_guest"),
+        "mapped": (".mapped", "collection_create_mapped"),
     },
 )
 def collection_create() -> None:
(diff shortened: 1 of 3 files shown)


sirosen:
Superseded by #919

PR #906, opened 2023-12-07
globus api --scope-string (for clients)

derek-globus:
## What?
* The ``globus api <service>`` command now supports a ``--scope-string`` parameter.

  * If supplied, the CLI will enforce that any specified scope strings are included
    in consent requirements *in addition to* standard service scope requirements.

  * This parameter may be supplied multiple times to specify multiple scope strings.

  * This parameter is only supported in the context of Client Credentials-based authentication.
    ([Client Credentials with GLOBUS_CLI_CLIENT_ID](https://docs.globus.org/cli/environment_variables/#client_credentials_with_globus_cli_client_id))

## Why?
Provide users a path to interact with APIs not currently supported in the CLI which require specific scope consents (ie. auth's project management apis) 

## Testing

```zsh
> export GLOBUS_CLI_CLIENT_ID="...redacted..."
> export GLOBUS_CLI_CLIENT_SECRET="...redacted..."
> globus api auth GET /v2/api/projects --scope-string 'urn:globus:auth:scope:auth.globus.org:manage_projects'
```

### DIFF
--- a/src/globus_cli/commands/api.py
+++ b/src/globus_cli/commands/api.py
@@ -9,7 +9,7 @@ import click
 import globus_sdk
 
 from globus_cli import termio, version
-from globus_cli.login_manager import LoginManager
+from globus_cli.login_manager import LoginManager, is_client_login
 from globus_cli.login_manager.scopes import CLI_SCOPE_REQUIREMENTS
 from globus_cli.parsing import command, group, mutex_option_group
 from globus_cli.termio import display
@@ -255,13 +255,14 @@ sends a 'GET' request to '{_get_url(service_name)}foo/bar'
     )
     @click.option("--no-retry", is_flag=True, help="Disable built-in request retries")
     @click.option(
-        "--scope",
+        "--scope-string",
         type=str,
         multiple=True,
         help=(
-            "A scope string to specify as required for this request. "
-            "Use this option multiple times to define multiple scope requirements."
-        )
+            "A scope string to consent to for this command (only supported for "
+            "confidential client usage). "
+            "Pass this option multiple times to specify multiple scopes."
+        ),
     )
     @mutex_option_group("--body", "--body-file")
     def service_command(
@@ -277,7 +278,7 @@ sends a 'GET' request to '{_get_url(service_name)}foo/bar'
         allow_errors: bool,
         allow_redirects: bool,
         no_retry: bool,
-        scope: tuple[str, ...] | None,
+        scope_string: tuple[str, ...],
     ) -> None:
         # the overall flow of this command will be as follows:
         # - prepare a client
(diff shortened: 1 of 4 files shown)


src/globus_cli/commands/api.py
```diff
@@ -240,6 +254,16 @@ def build_command(service_name: ServiceNameLiteral) -> click.Command:
...
     )
     @click.option("--no-retry", is_flag=True, help="Disable built-in request retries")
+    @click.option(
+        "--scope-string",
+        type=str,
+        multiple=True,
+        help=(
+            "A scope string to consent to for this command (only supported for "
+            "confidential client usage). "
+            "Pass this option multiple times to specify multiple scopes."
```
kurtmckee, line 264: Recommending removal of "for this command". I've rewrapped the lines as well.

```suggestion
            "A scope string to consent to "
            "(only supported for confidential client usage). "
            "Pass this option multiple times to specify multiple scopes."
```

tests/functional/test_api.py
```diff
@@ -105,3 +106,29 @@ def test_api_command_query_params_multiple_become_list(run_line):
...
+
+
+def test_api_command_with_scope_strings(monkeypatch, client_login, run_line):
+    load_response("cli.api.transfer_stub")
+    load_response("auth.oauth2_client_credentials_tokens")
+
+    run_line("globus api transfer get /foo --scope-string foobarjohn")
+
+    token_grant = [
+        kall for kall in responses.calls if kall.request.url.endswith("/token")
```
sirosen, line 118: Micro comment:
I'm knot shore what two kall my reactshun to this spelleng. But it fits with my liking for `klass` as a `class` alternative, so :shrug:

kurtmckee (approved):

src/globus_cli/login_manager/manager.py
```diff
@@ -118,6 +118,13 @@ def has_login(self, resource_server: str) -> bool:
 
         return self._validate_token(tokens["refresh_token"])
 
+    def _tokens_meet_auth_requirements(
+        self, resource_server: str, tokens: dict[str, t.Any]
+    ) -> bool:
+        return self._tokens_meet_static_requirements(
+            resource_server, tokens
+        ) and self._tokens_meet_nonstatic_requirements(resource_server, tokens)
```
sirosen, line 126: Should this now be used in the implementation of `has_login`?

sirosen (approved):
One minor question/refactoring option, but I like how tightly this is scoped.
Because it passes through the existing login manager mechanisms, I have fewer concerns about impact, upon review, than I voiced in earlier discussions.

tests/functional/test_api.py
derek-globus: Hmm, yeah this is definitely a practice that's in the same place in my head as `klass`. I'm not sure why though, I thought `call` (lowercased) was a namespaced word by `pytest` or`requests` or some standard testing library.

Looking for it now though I can't find the term that I'd be avoiding collisions this way...

src/globus_cli/commands/api.py
kurtmckee: Only about readability -- "to consent to for" caught my eye.

kurtmckee (approved):

PR #907, opened 2023-12-07
Replace all references to the tutorial endpoints

ada-globus:
image


kurtmckee (approved):
I approve, but part of me is thinking that it would be nice to make these start with `11111111-` and `22222222-` so that it's a little more guessable that these are bogus.

ada-globus:
> I approve, but part of me is thinking that it would be nice to make these start with `11111111-` and `22222222-` so that it's a little more guessable that these are bogus.

But isn't it easy to see at a glance to determine that these are simply `str(uuid.uuid5(uuid.UUID(int=0), "foo"))` 😉

PR #911, opened 2023-12-08
Remove README badges

kurtmckee:
The badges don't contribute meaningful information to me, just a wall of colors. I'd like to remove them.

## PyPI rendering

image

# GitHub rendering

image

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -10,26 +10,6 @@
     :alt: Supported Python Versions
     :target: https://img.shields.io/pypi/pyversions/globus-cli.svg
 
-.. image:: https://img.shields.io/badge/License-Apache%202.0-blue.svg
-    :alt: License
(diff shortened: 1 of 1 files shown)


sirosen:
I disagree regarding _some_ of the badges.
The `isort` one, for example, is useless and frivolous (I added it).
The python version badge is useful, not just for consumers, but also for us to keep an eye on the classifiers we're publishing.

Seeking a middle ground, can we keep, for now:
- build
- pypi version
- pypi pythons/classifiers

and ditch the rest?

I'm on the fence about the build badge providing value, but the other two are things I would want for a library.

kurtmckee:
Sure thing! I added those three back and force-pushed.

sirosen (approved):

PR #913, opened 2023-12-08
Add `globus api gcs` command for direct GCS Manager API usage

sirosen:
The addition of this command requires some large-diff-but-small-logic changes to the `globus api` command definition module.

Firstly, the major logical components of these commands are separated into three helpers:
- a decorator for the common parameters
- an "execute service command" helper
- a "handle scope string" helper (more on why this is needed below)

The command callback passes all of its parameters verbatim to the execution helper, so there's very little change to those parts of the command. The execution helper consumes any object of type `globus_sdk.BaseClient`, meaning it can be passed any service client, already authenticated/configured, and it uses that object equivalently.

The "handle scope string" component needs to be treated differently because it executes *before* there is a client object in hand, and alters how that client object is authenticated for later use. It differs subtly between the known services (handled by name) and GCS (where `resource_server == endpoint_id`).

The GCS variant uses a positional arg for ENDPOINT_ID, so it is declared separately from the other commands. Effectively, it's a different interface wrapping the same functionality.

---

One thing I noticed while testing was that we're currently doing some name-shadowing (purely accidentally) of `service_name`, which gets defined in module-scope for use in a loop. In my first draft, this caused a bug as `service_name` was defined, but didn't have the value I expected. I'd like to fix this in any of a number of ways, but I've kept it out of scope for this PR.

### DIFF
--- a/src/globus_cli/commands/api.py
+++ b/src/globus_cli/commands/api.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import json
 import sys
 import typing as t
+import uuid
 from collections import defaultdict
 
 import click
@@ -11,7 +12,7 @@ import globus_sdk
 from globus_cli import termio, version
 from globus_cli.login_manager import LoginManager, is_client_login
 from globus_cli.login_manager.scopes import CLI_SCOPE_REQUIREMENTS
-from globus_cli.parsing import command, group, mutex_option_group
+from globus_cli.parsing import command, endpoint_id_arg, group, mutex_option_group
 from globus_cli.termio import display
 from globus_cli.types import ServiceNameLiteral
 
@@ -19,6 +20,8 @@ if sys.version_info >= (3, 8):
     from typing import Literal
 else:
     from typing_extensions import Literal
+
+C = t.TypeVar("C", bound=t.Union[t.Callable, click.Command])
 
 
 class QueryParamType(click.ParamType):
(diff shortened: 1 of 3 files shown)


src/globus_cli/commands/api.py
```diff
@@ -263,102 +247,212 @@ def build_command(service_name: ServiceNameLiteral) -> click.Command:
...
-            # we're in the allow-errors case, so print the HTTP response
-            print_error_or_response(e)
-        else:
-            print_error_or_response(res)
 
     return t.cast(click.Command, service_command)
 
 
 for service_name in CLI_SCOPE_REQUIREMENTS:
     api_command.add_command(build_command(service_name))
```
kurtmckee, line 456: Is this the troublemaker? I'd make this very much in-scope for this PR.

```suggestion
for service_name_ in CLI_SCOPE_REQUIREMENTS:
    api_command.add_command(build_command(service_name_))
```

kurtmckee (approved):
This looks fine to me, but is there anything additional that needs to be considered that would make it infeasible to just rename the module-level `service_name`?

src/globus_cli/commands/api.py
sirosen: This is the troublemaker, indeed! We can make this change (I'll commit this suggestion after commenting), but I don't love that we still end up binding `service_name_` in module scope.

I was considering replacing this loop with the non-looping version with all of the names hardcoded in series. That's what I was considering out of scope for this PR.

kurtmckee: Oh! I think you can `del service_name_` after the loop. :+1:

sirosen: Oh, good call! Let me try after the merge.
That sounds best but I want this merged for other work to go smoothly.

PR #916, opened 2023-12-12
Rename Timers "jobs" to simply "timers"

kurtmckee:
### Other

* User timers are now referred to as "timers" rather than as "jobs".

This results in output at the CLI that shows the term "Timer ID" instead of "Job ID", for example.

### DIFF
--- a/src/globus_cli/commands/timer/_common.py
+++ b/src/globus_cli/commands/timer/_common.py
@@ -76,7 +76,7 @@ class ScheduleFormatter(formatters.FieldFormatter[t.Dict[str, t.Any]]):
 
 
 _COMMON_FIELDS = [
-    Field("Job ID", "job_id"),
+    Field("Timer ID", "job_id"),
     Field("Name", "name"),
     Field("Type", "callback_url", formatter=CallbackActionTypeFormatter()),
     Field("Submitted At", "submitted_at", formatter=formatters.Date),
@@ -85,7 +85,7 @@ _COMMON_FIELDS = [
 ]
 
 
-JOB_FORMAT_FIELDS = _COMMON_FIELDS + [
+TIMER_FORMAT_FIELDS = _COMMON_FIELDS + [
     Field("Status", "status"),
     Field("Last Run", "last_ran_at", formatter=formatters.Date),
     Field("Next Run", "next_run", formatter=formatters.Date),
@@ -95,7 +95,7 @@ JOB_FORMAT_FIELDS = _COMMON_FIELDS + [
     Field("Number of Timer Errors", "n_errors"),
 ]
 
-DELETED_JOB_FORMAT_FIELDS = _COMMON_FIELDS + [
+DELETED_TIMER_FORMAT_FIELDS = _COMMON_FIELDS + [
     Field("Status", "status"),
     Field("Stop After Date", "stop_after.date"),
     Field("Stop After Number of Runs", "stop_after.n_runs"),
--- a/src/globus_cli/commands/timer/resume.py
+++ b/src/globus_cli/commands/timer/resume.py
@@ -43,9 +43,9 @@ def resume_command(
     Resume a timer.
     """
     timer_client = login_manager.get_timer_client()
-    job_doc = timer_client.get_job(timer_id)
+    timer_doc = timer_client.get_job(timer_id)
 
-    gare = _get_inactive_reason(job_doc)
+    gare = _get_inactive_reason(timer_doc)
     if not skip_inactive_reason_check:
         check_inactive_reason(login_manager, timer_id, gare)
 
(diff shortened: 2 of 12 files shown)


src/globus_cli/commands/timer/resume.py
```diff
@@ -43,21 +43,14 @@ def resume_command(
...
-        timer_id,
-        update_credentials=(gare is not None),
-    )
-    display(
-        resumed,
-        text_mode=TextMode.text_raw,
-        simple_text=resumed["message"],
-    )
+    resumed = timer_client.resume_job(timer_id, update_credentials=(gare is not None))
+    display(resumed, text_mode=TextMode.text_raw, simple_text=resumed["message"])
```
sirosen, line 53: I would (actually _quite strongly_) prefer that we not remove trailing commas in cases like this, which causes `black` to format onto a single line.
Leaving trailing commas -- and in many cases I intentionally add them for this -- makes `black` split arguments over multiple lines. The result is one of the many benefits of `black`: minimal diffs.

If a keyword argument is added here, I should only have to review that addition, not the potential refolding of text caused by extending past the 88 char limit.

Let's please revert this part of the change and any other updates like this which aren't part of renaming.

tests/functional/timer/test_job_operations.py
```diff
@@ -5,15 +5,15 @@
 import pytest
 from globus_sdk._testing import RegisteredResponse, load_response, load_response_set
 
-from globus_cli.commands.timer._common import JOB_FORMAT_FIELDS
+from globus_cli.commands.timer._common import TIMER_FORMAT_FIELDS
 
-# NOTE: this is not quite the same as the "normal" job responses from
+# NOTE: this is not quite the same as the "normal" timer responses from
```
ada-globus, line 10: Nit: I might not change this note until we've updated the APIs, since technically it _is_ a job response today (and hopefully once we make it a proper timer response, we can return and remove the note entirely!)

src/globus_cli/commands/timer/_common.py
```diff
@@ -76,7 +76,7 @@ def render(self, value: dict[str, t.Any]) -> str:
 
 
 _COMMON_FIELDS = [
-    Field("Job ID", "job_id"),
+    Field("Timer ID", "job_id"),
```
sirosen, line 79: I want to voice a general concern I have about making changes like this, although in this case I think it's 99-100% fine.
Users can (and, in the past at least, often did) parse the text output using unix tools. e.g.,
```bash
timer_id="$(globus timer create transfer ... | \
  awk -F':' '$1=="Job ID"{print $2}')" | tr -d ' '
```

There are three reasons I think this particular one is not likely to be an issue:
- this mode of usage was much more common before `--jmespath` was added to the CLI and we started actively advocating that users leverage that flag to parse response data
- Timers is still a relatively young service
- this is the primary object ID, so there is no utility in parsing it out of `globus timer show` output

However, I want us to acknowledge that even though we don't _recommend_ parsing the text output, changing it always carries the risk/possibility that it breaks some downstream user script. I think we should include this in the changelog. It doesn't need to be a separate bullet point -- my thought is to append something like "In text output, the ID of a Timer is now referred to as `Timer ID`."

sirosen (changes requested):

ada-globus (approved):
I've got a _slight_ uneasiness about a little confusion between jobs and timers from an internal POV (i.e., internal naming) since things like "job_id" are referring to the specific key in the response and/or route they're interacting with. But I am super enthusiastic about the changes to the customer-facing messaging, and am not going to let my very modest reservation hold up what is overall a super positive change!

src/globus_cli/commands/timer/_common.py
kurtmckee: I can add some mention of that.

Having parsed text output of commands for over a decade, I can tell you that parsing text instead of the JSON output that's readily available is wrong, and we should recommend parsing the JSON directly.

src/globus_cli/commands/timer/resume.py
kurtmckee: To be fair, I didn't simply remove the trailing comma -- I manually reformatted it myself because it fit on a single line. I didn't realize you added trailing commas specifically to force things onto multiple lines, and I'll try to respect that.

I'll undo this change.

tests/functional/timer/test_job_operations.py
kurtmckee: Is this something you want changed?

src/globus_cli/commands/timer/_common.py
sirosen: I do not disagree in the slightest. But that has never stopped anyone from writing perl. :stuck_out_tongue_winking_eye:

tests/functional/timer/test_job_operations.py
ada-globus: I didn't request a change b/c I didn't feel strongly enough (or that the impact was significant enough) to insist on it, only that it registered on my radar. I leave it to your judgment, but will not be put out if you leave it as is.

src/globus_cli/commands/timer/resume.py
kurtmckee: This should be undone (simply kicked the change out of the previous commit and force-pushed).

src/globus_cli/commands/timer/_common.py
kurtmckee: I've updated the changelog fragment to give an example that output shows "Timer ID" instead of "Job ID".

sirosen (approved):

PR #918, opened 2023-12-13
Cleanups

kurtmckee:
I stumbled across these while investigating the login code.

These can be summarized as:

* Remove unnecessary parentheses
* Use `all()`'s shortcircuit logic when passed a generator


sirosen (approved):
Yep! Good improvements!

I've worked up a fixer for `slyp` which would auto-remove those parens, but because fixers have a slightly different interface from linters, it's not smooth to integrate it. So that work is a little stalled until I have more time to figure out how to rejigger some of the plumbing there.

PR #919, opened 2023-12-13
Implement GCS mapped collection creation

sirosen:
Introduce a new command, `globus gcs collection create mapped` (also available under `globus collection`).

In addition to the new command and tests, the following related improvements have been introduced here:

- `--public/--private` is now implemented once in the whole CLI, and shared between the multiple endpoint and collection variants as appropriate.
- In service of the above, passing legacy parameters for `globus endpoint` commands has been slightly refactored.
- `--identity-id` (also used by `globus gcs collection create guest`) now uses a common option definition

Some particular notes on issues from past variants of this:

- An earlier draft used `--storage-gateway`, with some justification that this would be future compatible with various ways of selecting and specifying gateways.

Given how the Guest creation command has evolved in a similar space (user credentials), I don't think this is likely to be our future path.
The option is now named `--storage-gateway-id` to harmonize it with the API's field name.

- Further consolidation of options may be possible.

After doing `--public/--private` and `--identity-id` and seeing the level of impact those changes carry, I've decided not to make any more changes in this scope of work.

- `mutex_option_group` application makes policy options "cleanly" mutex, but doesn't solve the larger interface complexity.

Because different connectors carry different expectations and constraints, `--google-project-id` is an option to mapped collection creation, but "obviously" incompatible with a posix storage gateway.
Similarly, the POSIX options aren't sensible on a Google Cloud Storage or S3 backed collection.

This is a minor interface problem for a CLI -- mapped creation may or may not need various options, and they need to harmonize with the backend requirements, which means that users cannot use the CLI presentation of options to learn what they need to provide.
For now, we're following the lead of the GCS team in the prior art provided by the gcs-cli: mapped collection creation simply allows these various options as siblings and treats them as mutex.

---

The full commit history here reveals the fact that this is an older body of work which was rebased.

- Port 'collection create mapped' from gcs-cli
- Improve defaults for 'collection create mapped'
- Convert GCP mapped create to common params
- Update 'collection create mapped' after rebase
- Add testing of 'collection create mapped'
- Consolidate the `--public/--private` option
- Update mapped creation to use --storage-gateway-id
- Make `--identity-id` a common option definition
- Mapped collection mutex policies use proper mutex
- Add changelog for mapped collection create

### DIFF
--- a/changelog.d/20231213_172721_sirosen_mapped_collection_create.md
+++ b/changelog.d/20231213_172721_sirosen_mapped_collection_create.md
@@ -1,0 +1,4 @@
+### Enhancements
+
+* Add `globus gcs collection create mapped` as a new command for creating
+  mapped collections
--- a/src/globus_cli/commands/collection/_common.py
+++ b/src/globus_cli/commands/collection/_common.py
@@ -1,9 +1,43 @@
 from __future__ import annotations
 
+import click
 import globus_sdk
 
+from globus_cli.login_manager.utils import get_current_identity_id
+from globus_cli.parsing import AnnotatedOption
 from globus_cli.termio import Field, formatters
 from globus_cli.types import DATA_CONTAINER_T
+
+
+class LazyCurrentIdentity:
+    def __init__(self, value: str | None) -> None:
+        self._value = value
+
+    @property
+    def value(self) -> str:
+        if self._value is None:
+            self._value = get_current_identity_id()
+        return str(self._value)
+
+
+def _identity_id_callback(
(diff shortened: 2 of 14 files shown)


ada-globus (approved):
I only skimmed the tests (lmk if you'd like me to give those a close inspection), but I've walked through everything else in detail. Looks GTG to me!

sirosen:
I feel relatively confident in the work and see no particular reason to stand on ceremony when our users may be needing this feature in the here-and-now.

PR #921, opened 2023-12-15
GCS/GCP Subscription ID Management

derek-globus:
[sc-28587](https://app.shortcut.com/globus/story/28587)

## What?
* Added new `globus gcs endpoint set-subscription-id` command which allows subscription managers and endpoint admins to modify the subscription ID for a gcs endpoint.
* Added `globus gcp set-subscription-id` which aliases existing `globus endpoint set-subscription-id` command, allowing subscription managers & collection admins to modify the subscription ID for a gcp collection.

## Testing
* Stood up a gcs endpoint in sandbox, tested setting/removing the subscription for:
   1. A user which is an endpoint admin but only subscription member
   2. A user which is a subscription manager but not an endpoint admin

* Stood up a gcp endpoint in sandbox, tested setting/removing the subscription for:
   1. A user which is a collection admin but only subscription member
   2. A user which is a subscription manager but not a collection admin

### DIFF
--- a/changelog.d/20231215_140223_derek_gcs_subscription_id_sc_28587.md
+++ b/changelog.d/20231215_140223_derek_gcs_subscription_id_sc_28587.md
@@ -1,0 +1,11 @@
+
+### Enhancements
+
+* `[sc-28587] <https://app.shortcut.com/globus/story/28587>`_
+  Added a new command `globus gcs endpoint set-subscription-id` which allows
+  subscription managers and endpoint admins to modify the subscription ID for a
+  gcs endpoint.
+
+* `[sc-28587] <https://app.shortcut.com/globus/story/28587>`_
+  Added a new command `globus gcp set-subscription-id` which allows subscription
+  managers and collection admins to modify the subscription ID for a gcp collection.
--- a/changelog.d/20231215_140223_derek_gcs_subscription_id_sc_28587.md
+++ b/changelog.d/20231215_140223_derek_gcs_subscription_id_sc_28587.md
@@ -4,7 +4,7 @@
 * `[sc-28587] <https://app.shortcut.com/globus/story/28587>`_
   Added a new command `globus gcs endpoint set-subscription-id` which allows
   subscription managers and endpoint admins to modify the subscription ID for a
-  gcs endpoint.
+  GCS endpoint.
 
 * `[sc-28587] <https://app.shortcut.com/globus/story/28587>`_
   Added a new command `globus gcp set-subscription-id` which allows subscription
--- a/changelog.d/20231215_140223_derek_gcs_subscription_id_sc_28587.md
+++ b/changelog.d/20231215_140223_derek_gcs_subscription_id_sc_28587.md
@@ -8,4 +8,4 @@
 
 * `[sc-28587] <https://app.shortcut.com/globus/story/28587>`_
   Added a new command `globus gcp set-subscription-id` which allows subscription
-  managers and collection admins to modify the subscription ID for a gcp collection.
+  managers and collection admins to modify the subscription ID for a GCP collection.
--- a/src/globus_cli/commands/gcp/__init__.py
+++ b/src/globus_cli/commands/gcp/__init__.py
@@ -6,10 +6,7 @@ from globus_cli.parsing import group
     lazy_subcommands={
         "create": (".create", "create_command"),
         "update": (".update", "update_command"),
-        "set-subscription-id": (
-            "endpoint.set_subscription_id",
-            "set_endpoint_subscription_id",
-        ),
+        "set-subscription-id": (".set_subscription_id", "set_endpoint_subscription_id"),
     },
 )
 def gcp_command() -> None:
--- a/src/globus_cli/commands/gcp/__init__.py
+++ b/src/globus_cli/commands/gcp/__init__.py
@@ -6,6 +6,10 @@ from globus_cli.parsing import group
     lazy_subcommands={
         "create": (".create", "create_command"),
         "update": (".update", "update_command"),
+        "set-subscription-id": (
+            "endpoint.set_subscription_id",
+            "set_endpoint_subscription_id",
+        ),
     },
 )
 def gcp_command() -> None:
--- a/src/globus_cli/commands/gcp/set_subscription_id.py
+++ b/src/globus_cli/commands/gcp/set_subscription_id.py
@@ -1,0 +1,68 @@
+from __future__ import annotations
+
+import typing as t
+import uuid
+
+import click
+
+from globus_cli.constants import EXPLICIT_NULL, ExplicitNullType
+from globus_cli.endpointish import Endpointish, EntityType
+from globus_cli.login_manager import LoginManager
+from globus_cli.parsing import command, endpoint_id_arg
+from globus_cli.termio import TextMode, display
+
+
+class GCPSubscriptionIdType(click.ParamType):
+    def get_type_annotation(self, _: click.Parameter) -> type:
+        return t.cast(type, uuid.UUID | ExplicitNullType)
+
+    def convert(
+        self, value: str, param: click.Parameter | None, ctx: click.Context | None
+    ) -> t.Any:
+        if ctx and ctx.resilient_parsing:
+            return None
+
+        if value.lower() == "null":
+            return EXPLICIT_NULL
+
+        try:
+            return uuid.UUID(value)
+        except ValueError:
+            msg = (
+                f"{value} is invalid. Expected either a UUID or the special value "
+                '"null"'
+            )
+            self.fail(msg, param, ctx)
+
+
+@command("set-subscription-id", short_help="Update a GCP endpoint's subscription")
+@endpoint_id_arg
+@click.argument("SUBSCRIPTION_ID", type=GCPSubscriptionIdType())
+@LoginManager.requires_login("transfer")
+def set_endpoint_subscription_id(
+    login_manager: LoginManager,
+    *,
+    endpoint_id: uuid.UUID,
+    subscription_id: uuid.UUID | ExplicitNullType,
+) -> None:
+    """
+    Update a GCP endpoint's subscription.
+
+    This operation does not require you to be an admin of the endpoint. It is useful in
+    cases where you are a subscription manager applying a subscription to an endpoint
+    administered by someone else.
+
+    SUBSCRIPTION_ID must be one of: (1) A valid subscription ID (UUID) or (2) the value
+    "null" (clears the endpoint's subscription).
+    """
+    transfer_client = login_manager.get_transfer_client()
+    epish = Endpointish(endpoint_id, transfer_client=transfer_client)
+    epish.assert_entity_type(expect_types=EntityType.GCP_MAPPED)
+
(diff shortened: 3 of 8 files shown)


changelog.d/20231215_140223_derek_gcs_subscription_id_sc_28587.md
```diff
@@ -0,0 +1,11 @@
+
+### Enhancements
+
+* `[sc-28587] <https://app.shortcut.com/globus/story/28587>`_
+  Added a new command `globus gcs endpoint set-subscription-id` which allows
+  subscription managers and endpoint admins to modify the subscription ID for a
+  gcs endpoint.
```
kurtmckee, line 7:
```suggestion
  GCS endpoint.
```

changelog.d/20231215_140223_derek_gcs_subscription_id_sc_28587.md
```diff
@@ -0,0 +1,11 @@
...
+
+* `[sc-28587] <https://app.shortcut.com/globus/story/28587>`_
+  Added a new command `globus gcp set-subscription-id` which allows subscription
+  managers and collection admins to modify the subscription ID for a gcp collection.
```
kurtmckee, line 11:
```suggestion
  managers and collection admins to modify the subscription ID for a GCP collection.
```

src/globus_cli/commands/gcs/endpoint/set_subscription_id.py
```diff
@@ -0,0 +1,72 @@
...
+
+class GCSSubscriptionIdType(click.ParamType):
+    def get_type_annotation(self, _: click.Parameter) -> type:
+        return t.cast(type, uuid.UUID | t.Literal["DEFAULT"] | ExplicitNullType)
+
+    def convert(
+        self, value: str, param: click.Parameter | None, ctx: click.Context | None
+    ) -> t.Any:
+        if value is None or (ctx and ctx.resilient_parsing):
+            return None
```
kurtmckee, line 22: Can `value` be `None`? It isn't reflected in the function annotations.

src/globus_cli/commands/gcs/endpoint/set_subscription_id.py
```diff
@@ -0,0 +1,72 @@
...
+            return value
+        except ValueError:
+            msg = (
+                f"{value} is invalid. Expected either a UUID or the special "
+                'values "DEFAULT" or "null"'
+            )
+            self.fail(msg, param, ctx)
+
+
+@command("set-subscription-id", short_help="Update an endpoint's subscription")
```
kurtmckee, line 38: Short help differs from first line of docstring:

```suggestion
@command("set-subscription-id", short_help="Update an endpoint's subscription ID")
```

src/globus_cli/commands/gcs/endpoint/set_subscription_id.py
```diff
@@ -0,0 +1,72 @@
...
+) -> None:
+    """
+    Update an endpoint's subscription ID.
+
+    SUBSCRIPTION_ID must be one of: (1) A valid subscription ID (UUID), (2) the value
+    "DEFAULT" (requires that you manage exactly one subscription & assigns the endpoint
+    to that subscription), or (3) the value "null" (clears the endpoint's subscription)
+
+    Setting a subscription requires that you are a subscription manager for the
+    subscription group being assigned.
```
kurtmckee, line 56: The phrase "subscription group" catches my eye. Just want to confirm that this terminology is correct.

tests/functional/gcs/endpoint/test_set_subscription_id.py
```diff
@@ -0,0 +1,64 @@
...
+    )
+
+    result = run_line(
+        f"globus gcs endpoint set-subscription-id {endpoint_id} {subscription_id}"
+    )
+
+    assert f"Updated Endpoint {endpoint_id}" in result.stdout
+
+
+def test_gcs_endpoint_set_subscription_id__when_not_subscription_manager(
```
kurtmckee, line 34:
```suggestion
def test_gcs_endpoint_set_subscription_id_when_not_subscription_manager(
```

tests/functional/gcs/endpoint/test_set_subscription_id.py
kurtmckee, line 1: These tests don't _appear_ to exercise the case where `value` is `None` (see the note for the `.convert()` function).

kurtmckee (changes requested):

tests/functional/gcs/endpoint/test_set_subscription_id.py
derek-globus: This extra _ was intentional.

It's a test naming convention of
```
def test_operation[__when_condition][__then_result]
```

kurtmckee: I wondered if it was intentional. Thanks for sharing that convention!

src/globus_cli/commands/gcs/endpoint/set_subscription_id.py
derek-globus: Looking at the super type, it cannot actually be None; will update in new rev.

src/globus_cli/commands/gcs/endpoint/set_subscription_id.py
derek-globus: Syncing in the opposite direction (updating the first line of the docstring.

src/globus_cli/commands/gcs/endpoint/set_subscription_id.py
derek-globus: Posted this Q to auth for a more authoritative answer:

https://globus.slack.com/archives/C03N0A8J7/p1703102225633449

tests/functional/gcs/endpoint/test_set_subscription_id.py
derek-globus: Yep, one of those little naming tricks that helps frame the way you think about writing a test.

tests/functional/gcs/endpoint/test_set_subscription_id.py
derek-globus: Dropping the `None` check should alleviate this concern.

src/globus_cli/commands/gcs/endpoint/set_subscription_id.py
derek-globus: Updated to ```Setting a subscription requires that you are a subscription manager for the subscription being assigned.``` based on Josh's response.

kurtmckee (approved):

src/globus_cli/commands/gcp/__init__.py
```diff
@@ -6,6 +6,10 @@
     lazy_subcommands={
         "create": (".create", "create_command"),
         "update": (".update", "update_command"),
+        "set-subscription-id": (
+            "endpoint.set_subscription_id",
+            "set_endpoint_subscription_id",
+        ),
```
sirosen, line 12: I'm okay with this as a short-term position, but I'd like us to write a new command for this, rather than reusing the existing one.

My reason is primarily helptext and parameter names. If you look at the existing command, you'll see it states
```
Unlike the '--managed' flag for 'globus endpoint update' ...
```

Which is not something we want to retain into the future.

I think it's also worth having the new command check to confirm that it's operating on a `GCP_mapped_collection` rather than a `GCP_guest_collection`. That way, we more clearly explain to users that it's only possible to use the relevant command on a GCP host, not the shares underneath that host.

I'm perfectly happy for us to pursue that change as a fast-follow to this PR though.

sirosen (approved):

src/globus_cli/commands/gcp/__init__.py
derek-globus: I initially did write a new command for GCP but erased it when it ended up being identical to this one.

---

If we do add a duplicate code path, it's worth noting that introduces the risk of drift (one path gets updated but not the other).

If it's just a helptext concern, I don't think that's worth a copy/paste. If it's adding additional validation of `mapped` vs `guest` collection types; I could buy it. It's a fairly small copy/paste.

---

I'll include that in a new rev of this PR.

sirosen: I'm not that concerned about the paths being duplicated, since I think they are actually _different_. The question IMO is only how much it's worth working hard to avoid code duplication.

I want the following:
- GCP-specific checks in the GCP command
- legacy-specific deprecations in the legacy command
- different helptext for the two

The worry about drift would be more pronounced if this didn't take the form of a deprecation. However, I don't think we want to keep `globus endpoint` (the whole tree), so we should be making moves to deprecate its various pieces.

src/globus_cli/commands/gcp/set_subscription_id.py
```diff
@@ -0,0 +1,68 @@
...
+from globus_cli.termio import TextMode, display
+
+
+class GCPSubscriptionIdType(click.ParamType):
+    def get_type_annotation(self, _: click.Parameter) -> type:
+        return t.cast(type, uuid.UUID | ExplicitNullType)
+
+    def convert(
+        self, value: str, param: click.Parameter | None, ctx: click.Context | None
+    ) -> t.Any:
```
sirosen, line 21: It's not super-closely related to this PR, but just because I think it's cool, I wanted to mention that under the latest `click-type-test`, you can write `convert(...) -> uuid.UUID | ExplicitNullType` and omit `get_type_annotation`.

src/globus_cli/commands/gcp/set_subscription_id.py
```diff
@@ -0,0 +1,68 @@
...
+    "null" (clears the endpoint's subscription).
+    """
+    transfer_client = login_manager.get_transfer_client()
+    epish = Endpointish(endpoint_id, transfer_client=transfer_client)
+    epish.assert_entity_type(expect_types=EntityType.GCP_MAPPED)
+
+    sub_id = None if subscription_id is EXPLICIT_NULL else str(subscription_id)
+    res = transfer_client.put(
+        f"/endpoint/{endpoint_id}/subscription",
+        data={"subscription_id": sub_id},
```
sirosen, line 65: I want to note, because it's not necessarily obvious at two levels, that you could write as follows:
```suggestion
    res = transfer_client.put(
        f"/endpoint/{endpoint_id}/subscription",
        data={"subscription_id": ExplicitNullType.nullify(subscription_id)},
```

The first thing to note is that `nullify()` is a classmethod which turns `EXPLICIT_NULL` to `None`.
The second, less obvious, thing is that the SDK now supports automatically stringifying UUIDs.

I don't know if it's _better_ -- both seem equally readable to me. But it's _different_ and I wanted to offer it up as an option.

sirosen (changes requested):
This is the change I was hoping to see! There's a a docstring typo to fix, and an optional suggestion I shared.

src/globus_cli/commands/gcp/set_subscription_id.py
sirosen: With #923 merged, you could now apply this change here (after a rebase).

sirosen (approved):

PR #923, opened 2023-12-21
Update to click-type-test v0.0.7, using new return type annotation inspection

sirosen:
click-type-test ~v0.0.6~ v0.0.7 is here, and provides the ability to deduce the type of a parameter from the return type of a `callback` or `convert` method.

Overriding with `AnnotatedParameter` and `AnnotatedParamType` is still fully supported, so this change is almost entirely additive.
For fuller detail on how types are determined, see:
  https://github.com/sirosen/click-type-test#type-determination-logic

Most of these changes are completely innocuous and center only on the types and simplifying things where possible, but there are some minor corrections.
Most notably, this caught some improper handling of the timeout exit code control flag.


sirosen:
This conflicts with #927 , so I'm marking it as a draft for now (and rebasing it onto that PR to serialize these two).

PR #929, opened 2024-01-04
Handle unrecognized errors in guest create

### LINKED ISSUE #928: Error with command: globus gcs collection create guest
Command generates error without much additional info. 

Command:
globus gcs collection create guest --user-credential-id <credential ID> <collection ID> /path/to/shared_dir test

Error:
AttributeError: 'str' object has no attribute 'get'

sirosen:
`globus gcs collection create guest` previously assumed that the shape
of an error from GCS met its expectations that `detail` would be dict
data. However, GCS docs do not guarantee this -- `detail` can be of
any type. Therefore, the inspection should explicitly check for this
and fail if the field is present with an unexpected type.
Failure to do so resulted in a `get()` call on a non-dict object.

Resolves #928


derek-globus (approved):
Oh noez!

PR #935, opened 2024-01-09
Remove support for Python 3.7

sirosen:
3.7 reached EOL in June, 2023, so it has been over 6 months since EOL.
As requisite maintenance, we need to update the lower bounds for what we support.

Most of the changes here are isolated into the second commit, which touches pyupgrade config and then fixes up the results.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -60,7 +60,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
         # we do not want a large number of windows and macos builds, so
         # enumerate them explicitly
         include:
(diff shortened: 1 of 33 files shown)


.github/workflows/build.yaml
```diff
@@ -283,15 +283,15 @@ jobs:
 
   # use the oldest python version we support for this build
   test-ancient-virtualenv:
-    name: "test on py3.7, using old virtualenv"
```
kurtmckee, line 286: Is this test still relevant? [virtualenv 16.7.12's trove classifiers](https://pypi.org/project/virtualenv/16.7.12/) only claim to support Python 3.7 and lower (though its actual constraints are Python 2.7 and 3.4+).

My gut is that this job should be re-evaluated.

sirosen: It definitely needs to be re-evaluated.
I need to look at the commit history and check why we have this test. I remember that it was something to do with the `site` module being broken on old venvs, but it might not be relevant anymore.

I was going to check and maybe remove it after this PR, but we could fold it in here.

kurtmckee: It doesn't have to be part of this PR, but I wanted to get initial feedback for this comment before marking this as "LGTM". :grinning:

kurtmckee (approved):
"LGTM"!

PR #937, opened 2024-01-24
Improve handling of clock skew when decoding a JWT

sirosen:
- increase leeway from 64s to 300s
- add a warning about clock skew vs the server on failure

Regarding the increase from 64s to 300s, this represents a refinement of the reasoning around what our tolerance should be, based on the same relevant data + the experience of users seeing errors even in the presence of this higher tolerance. As noted in the comments, the new conclusion drawn is that the tolerance value selected by Kerberos, an existing authn framework, is a reasonable default for us to use as well. That is chosen instead of the slightly arbitrary 64s from before.

The warning is emitted if the "Date" header in the response can be parsed to show us a delta between the local system clock and the server time. It is a simple echo to stderr, to guarantee that it appears in this failure mode.

These changes have been tested manually.

### DIFF
--- a/changelog.d/20240129_181132_sirosen_improve_jwt_clock_skew_tolerance.md
+++ b/changelog.d/20240129_181132_sirosen_improve_jwt_clock_skew_tolerance.md
@@ -1,0 +1,5 @@
+### Other
+
+* `globus login` and related commands are now more tolerant of clock drift, and
+  will emit a clearer error message when clock drift is severe enough to cause
+  errors during authentication attempts.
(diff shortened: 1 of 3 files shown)


kurtmckee (changes requested):
Please add a test that exercises the new clock-skew-detection code.

This is particularly important for catching potential deprecation warnings from within the Python standard library.

sirosen:
It was really, really good for you to call for the extra discipline of adding some tests (which I skipped in favor of my manual tests).

It revealed a bug in how invalid dates are handled. More testing revealed a very subtle stdlib contract change in `email.utils` which needed a second fix. This is all good stuff to be catching now, rather than later.

---

I'm satisfied with the testing path which I took, which is basically to unit test this function for these behaviors and not worry about trying to develop a functional test for now.

kurtmckee (approved):

PR #938, opened 2024-01-25
Fix endpoint permission create principal flag value bug

derek-globus:
## What?
Fixes a bug identified in user ticket [zd-381646](https://globusonline.zendesk.com/agent/tickets/381646) where `globus endpoint permission create` failed to correctly parse the flag input values for `--anonymous` and `--all-authenticated` options.

## Testing
Added unit tests which invoke `globus endpoint permission create` with the 4 standard principal types & verifies they are passed through as expected to transfer.

### DIFF
--- a/src/globus_cli/parsing/shared_options/__init__.py
+++ b/src/globus_cli/parsing/shared_options/__init__.py
@@ -313,6 +313,7 @@ def security_principal_opts(
             f = click.option(
                 "--anonymous",
                 "principal",
+                type=click.Tuple([str, str]),
                 flag_value=("anonymous", ""),
                 help=(
                     "Allow anyone access, even without logging in "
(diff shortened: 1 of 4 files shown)


kurtmckee (approved):

src/globus_cli/parsing/shared_options/__init__.py
```diff
@@ -313,6 +313,7 @@ def decorate(f: C) -> C:
             f = click.option(
                 "--anonymous",
                 "principal",
+                type=click.Tuple([str, str]),
```
derek-globus, line 316: Note: these are necessary as without them, the value gets injected as a string rather than a tuple:
```python
principal="('anonymous', '')"
```

PR #940, opened 2024-01-25
Fix incorrect parameter value specified in changelog

derek-globus:
## What?
Update the changelog to accurately specify correct flag value.

The value passed through to transfer is `all-authenticated-users` but the flag in the cli is `--all-authenticated` ([ref](https://github.com/derek-globus/globus-cli/blob/439793725fb74c794bb34852fed475c8094ad922/src/globus_cli/parsing/shared_options/__init__.py#L324-L332))

### DIFF
--- a/changelog.adoc
+++ b/changelog.adoc
@@ -6,8 +6,8 @@
 
 Bugfixes
 
-* Fixes a bug which would not allow users to utilize ``--anonymous`` or
-  ``--all-authenticated`` when creating an endpoint permission.
+* Fixes a bug which would not allow users to utilize `--anonymous` or
+  `--all-authenticated` when creating an endpoint permission.
 
 Other
 
--- a/changelog.adoc
+++ b/changelog.adoc
@@ -6,8 +6,8 @@
 
 Bugfixes
 
-* Fixes a bug which would not allow users to utilize "--anonymous" or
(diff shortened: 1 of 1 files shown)


changelog.adoc
```diff
@@ -6,8 +6,8 @@
 
 Bugfixes
 
-* Fixes a bug which would not allow users to utilize "--anonymous" or
-  "--all-authenticated-users" when creating an endpoint permission.
+* Fixes a bug which would not allow users to utilize ``--anonymous`` or
+  ``--all-authenticated`` when creating an endpoint permission.
```
sirosen, line 10:
```suggestion
* Fixes a bug which would not allow users to utilize `--anonymous` or
  `--all-authenticated` when creating an endpoint permission.
```

I _believe_ that single backtick and double backtick are both accepted by asciidoc/docbook, but I'd rather play it safe. As far as I know, ReST is the only markdown language with the double-backtick syntax.

kurtmckee (approved):
Syntax feedback aside, this reads fine to me.

changelog.adoc
derek-globus: Sure, I don't have a good mapping in my head of what syntax is valid where.

I just know that Kurt has frequently corrected by single backticks with double backticks in changelog fragments so I started to default to that in changelogs.

PR #946, opened 2024-02-06
Added globus gcs endpoint role commands

derek-globus:
## What?
* Added gcs endpoint role management commands:
    * `globus gcs endpoint role create`
    * `globus gcs endpoint role list`
    * `globus gcs endpoint role show`
    * `globus gcs endpoint role delete`

* Introduced principal urn resolution as a common util function
   * This pulls urn resolution up from search index role creation into a common functionality to mitigate wheel reinvention.

## Usage
```
# Create a role for a user by username
> globus gcs endpoint role create $EP activity_manager derk@uchicago.edu
ID:        1464ecf6-c53e-11ee-9d93-0b316eac9044
Role:      activity_manager
Principal: derk@uchicago.edu

# List my roles
> globus gcs endpoint role list $EP
ID                                   | Role          | Principal       
------------------------------------ | ------------- | ----------------
8e1f2c75-140d-442e-8133-6d64cdba6bb0 | owner         | derek@globus.org
d6496578-9b6c-11ee-b77e-714fd30d1108 | administrator | derek@globus.org

# List all roles
> globus gcs endpoint role list $EP --all-roles
ID                                   | Role             | Principal        
------------------------------------ | ---------------- | -----------------
1464ecf6-c53e-11ee-9d93-0b316eac9044 | activity_manager | derk@uchicago.edu
8e1f2c75-140d-442e-8133-6d64cdba6bb0 | owner            | derek@globus.org 
d6496578-9b6c-11ee-b77e-714fd30d1108 | administrator    | derek@globus.org 

# Show a particular role
> globus gcs endpoint role show $EP 1464ecf6-c53e-11ee-9d93-0b316eac9044 
ID:        1464ecf6-c53e-11ee-9d93-0b316eac9044
Role:      activity_manager
Principal: derk@uchicago.edu

# Create a role by identity UUID
> globus gcs endpoint role create $EP activity_monitor 97dcf40b-836a-49fa-9541-1c2f74ee0cc2                       
ID:        e5e3be56-c53e-11ee-9d93-0b316eac9044
Role:      activity_monitor
Principal: derk@uchicago.edu

# Create a role by group UUID
> globus gcs endpoint role create $EP activity_manager 9dc4c59b-0b3a-11ec-9a67-afcad34eca9b --principal-type group 
ID:        96da8fb0-c53e-11ee-9d93-0b316eac9044
Role:      activity_manager
Principal: Globus Group (9dc4c59b-0b3a-11ec-9a67-afcad34eca9b)

# Attempt to create a role for a non-existent identity
> globus gcs endpoint role create $EP activity_monitor d3c543b2-8119-4ed2-bf18-bbd1e005e435
Globus CLI Error: A Globus API Error Occurred.
HTTP status:      400
code:             bad_request
message:          Role principal is not a valid Globus account

# Delete a role
> globus gcs endpoint role delete $EP 1464ecf6-c53e-11ee-9d93-0b316eac9044 
Deleted role 1464ecf6-c53e-11ee-9d93-0b316eac9044
```

### DIFF
--- a/src/globus_cli/commands/gcs/endpoint/role/create.py
+++ b/src/globus_cli/commands/gcs/endpoint/role/create.py
@@ -1,0 +1,59 @@
+import typing as t
+import uuid
+
+import click
+import globus_sdk
+
+from globus_cli.commands.gcs.endpoint.role._common import role_fields
+from globus_cli.login_manager import LoginManager
+from globus_cli.parsing import command, endpoint_id_arg
+from globus_cli.termio import TextMode, display
+from globus_cli.utils import resolve_principal_urn
+
+_VALID_ROLES = ["administrator", "activity_manager", "activity_monitor"]
+
+
+@command(
+    "show",
+)
+@endpoint_id_arg
+@click.argument("ROLE", type=click.Choice(_VALID_ROLES), metavar="ROLE")
+@click.argument("PRINCIPAL", type=str)
+@click.option(
+    "--principal-type",
+    type=click.Choice(["identity", "group"]),
+    help="Qualifier to specify what type of principal (identity or group) is provided.",
+)
+@LoginManager.requires_login("transfer")
+def create_command(
+    login_manager: LoginManager,
+    *,
+    endpoint_id: uuid.UUID,
+    role: str,
+    principal: str,
+    principal_type: t.Literal["identity", "group"] | None,
+) -> None:
+    """
+    Create a role on a GCS Endpoint.
+
+    ROLE may be one of: ["administrator", "activity_manager", "activity_monitor"].
+
+    PRINCIPAL may be either a username or a UUID/URN for an identity or group.
+
+       * If a UUID, you must also provide --principal-type.
+    """
+    gcs_client = login_manager.get_gcs_client(endpoint_id=endpoint_id)
+    auth_client = login_manager.get_auth_client()
+
+    # Format the principal into a URN
+    principal_urn = resolve_principal_urn(
+        auth_client=auth_client,
+        principal_type=principal_type,
+        principal=principal,
+    )
+
+    res = gcs_client.create_role(
+        globus_sdk.GCSRoleDocument(role=role, principal=principal_urn)
+    )
+
+    display(res, text_mode=TextMode.text_record, fields=role_fields(auth_client))
--- a/src/globus_cli/commands/gcs/endpoint/role/create.py
+++ b/src/globus_cli/commands/gcs/endpoint/role/create.py
@@ -29,7 +29,7 @@ def create_command(
     login_manager: LoginManager,
     *,
     endpoint_id: uuid.UUID,
-    role: t.Literal["administrator", "activity_manager", "activity_monitor"],
+    role: _VALID_ROLES,
     principal: str,
     principal_type: t.Literal["identity", "group"] | None,
 ) -> None:
--- a/src/globus_cli/commands/gcs/endpoint/role/create.py
+++ b/src/globus_cli/commands/gcs/endpoint/role/create.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import typing as t
 import uuid
 
@@ -29,7 +31,7 @@ def create_command(
     login_manager: LoginManager,
     *,
     endpoint_id: uuid.UUID,
-    role: str,
+    role: t.Literal["administrator", "activity_manager", "activity_monitor"],
     principal: str,
     principal_type: t.Literal["identity", "group"] | None,
 ) -> None:
--- a/src/globus_cli/commands/gcs/endpoint/role/create.py
+++ b/src/globus_cli/commands/gcs/endpoint/role/create.py
@@ -12,14 +12,12 @@ from globus_cli.parsing import command, endpoint_id_arg
 from globus_cli.termio import TextMode, display
 from globus_cli.utils import resolve_principal_urn
 
-_VALID_ROLES = ["administrator", "activity_manager", "activity_monitor"]
+_VALID_ROLES = t.Literal["administrator", "activity_manager", "activity_monitor"]
 
 
-@command(
-    "show",
-)
+@command("create")
 @endpoint_id_arg
-@click.argument("ROLE", type=click.Choice(_VALID_ROLES), metavar="ROLE")
+@click.argument("ROLE", type=click.Choice(t.get_args(_VALID_ROLES)), metavar="ROLE")
 @click.argument("PRINCIPAL", type=str)
 @click.option(
     "--principal-type",
@@ -38,11 +36,11 @@ def create_command(
     """
     Create a role on a GCS Endpoint.
 
-    ROLE may be one of: ["administrator", "activity_manager", "activity_monitor"].
+    ROLE must be one of "administrator", "activity_manager", or "activity_monitor".
 
-    PRINCIPAL may be either a username or a UUID/URN for an identity or group.
-
-       * If a UUID, you must also provide --principal-type.
+    PRINCIPAL must be a username, UUID, or URN associated with a globus identity or
+    group.
+    If UUID, use `--principal-type` to specify the type (defaults to "identity").
     """
     gcs_client = login_manager.get_gcs_client(endpoint_id=endpoint_id)
     auth_client = login_manager.get_auth_client()
--- a/src/globus_cli/commands/gcs/endpoint/role/delete.py
+++ b/src/globus_cli/commands/gcs/endpoint/role/delete.py
@@ -7,9 +7,7 @@ from globus_cli.parsing import command, endpoint_id_arg
 from globus_cli.termio import TextMode, display
 
 
-@command(
-    "show",
-)
+@command("delete")
 @endpoint_id_arg
 @click.argument("ROLE_ID", type=click.UUID)
 @LoginManager.requires_login("transfer")
--- a/src/globus_cli/commands/gcs/endpoint/role/delete.py
+++ b/src/globus_cli/commands/gcs/endpoint/role/delete.py
@@ -1,0 +1,27 @@
+import uuid
+
+import click
+
+from globus_cli.login_manager import LoginManager
+from globus_cli.parsing import command, endpoint_id_arg
+from globus_cli.termio import TextMode, display
+
+
+@command(
+    "show",
+)
+@endpoint_id_arg
+@click.argument("ROLE_ID", type=click.UUID)
+@LoginManager.requires_login("transfer")
+def delete_command(
+    login_manager: LoginManager,
+    *,
+    endpoint_id: uuid.UUID,
+    role_id: uuid.UUID,
+) -> None:
+    """Delete a role from a GCS Endpoint."""
+    gcs_client = login_manager.get_gcs_client(endpoint_id=endpoint_id)
+
+    res = gcs_client.delete_role(role_id)
+
+    display(res, text_mode=TextMode.text_raw, response_key="message")
--- a/src/globus_cli/commands/gcs/endpoint/role/list.py
+++ b/src/globus_cli/commands/gcs/endpoint/role/list.py
@@ -6,14 +6,11 @@ from globus_sdk.paging import Paginator
 from globus_cli.commands.gcs.endpoint.role._common import role_fields
 from globus_cli.login_manager import LoginManager
 from globus_cli.parsing import command, endpoint_id_arg
-from globus_cli.services.transfer import iterable_response_to_dict
 from globus_cli.termio import display
 from globus_cli.utils import PagingWrapper
 
 
-@command(
-    "show",
-)
+@command("list")
 @endpoint_id_arg
 @click.option(
     "--all-roles", is_flag=True, help="Show all roles, not just yours.", default=False
@@ -30,13 +27,11 @@ def list_command(
     auth_client = login_manager.get_auth_client()
 
     paginator = Paginator.wrap(gcs_client.get_role_list)
-    if all_roles:
-        res = PagingWrapper(paginator(include="all_roles").items())
-    else:
-        res = PagingWrapper(paginator().items())
+    paginated_call = paginator(include="all_roles") if all_roles else paginator()
+    paging_wrapper = PagingWrapper(paginated_call.items(), json_conversion_key="DATA")
 
     display(
-        res,
+        paging_wrapper,
         fields=role_fields(auth_client),
-        json_converter=iterable_response_to_dict,
+        json_converter=paging_wrapper.json_converter,
     )
--- a/src/globus_cli/commands/gcs/endpoint/role/list.py
+++ b/src/globus_cli/commands/gcs/endpoint/role/list.py
@@ -1,0 +1,42 @@
+import uuid
+
+import click
+from globus_sdk.paging import Paginator
+
+from globus_cli.commands.gcs.endpoint.role._common import role_fields
+from globus_cli.login_manager import LoginManager
+from globus_cli.parsing import command, endpoint_id_arg
+from globus_cli.services.transfer import iterable_response_to_dict
+from globus_cli.termio import display
+from globus_cli.utils import PagingWrapper
+
+
+@command(
+    "show",
+)
+@endpoint_id_arg
+@click.option(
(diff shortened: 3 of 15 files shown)


src/globus_cli/commands/gcs/endpoint/role/list.py
```diff
@@ -0,0 +1,42 @@
+import uuid
+
+import click
+from globus_sdk.paging import Paginator
+
+from globus_cli.commands.gcs.endpoint.role._common import role_fields
+from globus_cli.login_manager import LoginManager
+from globus_cli.parsing import command, endpoint_id_arg
+from globus_cli.services.transfer import iterable_response_to_dict
```
sirosen, line 9: It feels slightly awkward / boundary-breaking to me that we'd have a non-Transfer-API command which pulls in a data handler from Transfer tools. On the other hand, I can understand why you'd prefer not to make things different unnecessarily, and the helper doesn't _really_ have anything to do with Transfer (other than aping its payload format a little).

Since we're already putting some supporting tools into `utils`, maybe we should do one more in this PR? Here's a slightly revamped version of this helper which might be nice:
```python
def iterable_response_to_dict(iterator: t.Iterable[t.Any], *, key: str = "DATA") -> dict[str, list[t.Any]]:
    output_dict: dict[str, list[t.Any]] = {key: []}
    for item in iterator:
        try:
            value = item.data
        except AttributeError:
            value = item
        output_dict[key].append(value)
    return output_dict
```

What do you think?

src/globus_cli/commands/gcs/endpoint/role/list.py
```diff
@@ -0,0 +1,42 @@
...
+) -> None:
+    """List all roles on a GCS Endpoint associated with you."""
+    gcs_client = login_manager.get_gcs_client(endpoint_id=endpoint_id)
+    auth_client = login_manager.get_auth_client()
+
+    paginator = Paginator.wrap(gcs_client.get_role_list)
+    if all_roles:
+        res = PagingWrapper(paginator(include="all_roles").items())
+    else:
+        res = PagingWrapper(paginator().items())
```
sirosen, line 36: I want to note and acknowledge -- but explicitly not try to address here -- that it would be nice to have the `json_converter` behavior built into `PagingWrapper` more deeply. These pieces were added incrementally over time, and could do with a bit of a clean redesign.
(Also setting aside the idea of changing how CLI pagination should work: I mean that even if we're keeping exactly the same behavior, we could probably have a tidier implementation.)

I'm glad that the tools we have continue to work to put together new commands, but there's nothing like seeing it in code review to drive home that these pieces don't quite click together the way I'd like.

src/globus_cli/commands/gcs/endpoint/role/list.py
```diff
@@ -0,0 +1,42 @@
...
+from globus_cli.commands.gcs.endpoint.role._common import role_fields
+from globus_cli.login_manager import LoginManager
+from globus_cli.parsing import command, endpoint_id_arg
+from globus_cli.services.transfer import iterable_response_to_dict
+from globus_cli.termio import display
+from globus_cli.utils import PagingWrapper
+
+
+@command(
+    "show",
```
sirosen, line 15:
```suggestion
    "list",
```
typop

src/globus_cli/commands/gcs/endpoint/role/delete.py
```diff
@@ -0,0 +1,27 @@
...
+
+import click
+
+from globus_cli.login_manager import LoginManager
+from globus_cli.parsing import command, endpoint_id_arg
+from globus_cli.termio import TextMode, display
+
+
+@command(
+    "show",
```
sirosen, line 11:
```suggestion
    "delete",
```

typoo

src/globus_cli/commands/gcs/endpoint/role/create.py
```diff
@@ -0,0 +1,61 @@
...
+    principal_type: t.Literal["identity", "group"] | None,
+) -> None:
+    """
+    Create a role on a GCS Endpoint.
+
+    ROLE may be one of: ["administrator", "activity_manager", "activity_monitor"].
+
+    PRINCIPAL may be either a username or a UUID/URN for an identity or group.
+
+       * If a UUID, you must also provide --principal-type.
```
sirosen, line 45: I would make some pretty minor adjustments to this text on three points:
- "may" -> "must", "is"
- unwrap that list of roles by removing the brackets and adding `and`
- correct `--principal-type` to note the default

```suggestion
    ROLE must be one "administrator", "activity_manager", or "activity_monitor".

    PRINCIPAL is either a username or a UUID/URN for an identity or group.
    If it is a UUID, use `--principal-type` to provide the type (defaults to "identity").
```

That third point is the most important. Regarding the other changes, I'm really just polishing things a little. I think we already throw a lot of string syntaxes at users (e.g. `{foo,bar}` vs `[foo|bar]` vs ...) and I want to try to reduce the degree of that.

src/globus_cli/utils.py
```diff
@@ -216,3 +220,90 @@ def __init__(
...
+        if principal.startswith("urn:globus:auth:identity:"):
+            return principal
+
+        if not principal.startswith("urn:"):
+            resolved = auth_client.maybe_lookup_identity_id(principal)
+            if resolved:
+                return f"urn:globus:auth:identity:{resolved}"
+
+        raise click.UsageError(
+            f"{principal_type_key} identity but '{principal}' is not a valid "
```
sirosen, line 281: Should we quote the leading bit?

```suggestion
            f"'{principal_type_key} identity' but '{principal}' is not a valid "
```

src/globus_cli/utils.py
```diff
@@ -216,3 +220,90 @@ def __init__(
...
+    elif principal_type == "group":
+        if principal.startswith("urn:globus:groups:id:"):
+            return principal
+
+        resolved = principal if _is_uuid(principal) else None
+        if resolved:
+            return f"urn:globus:groups:id:{resolved}"
+
+        raise click.UsageError(
+            f"{principal_type_key} group but '{principal}' is not a valid group UUID or"
```
sirosen, line 295: As above, maybe?

```suggestion
            f"'{principal_type_key} group' but '{principal}' is not a valid group UUID or"
```

src/globus_cli/commands/gcs/endpoint/role/create.py
```diff
@@ -0,0 +1,61 @@
...
+import click
+import globus_sdk
+
+from globus_cli.commands.gcs.endpoint.role._common import role_fields
+from globus_cli.login_manager import LoginManager
+from globus_cli.parsing import command, endpoint_id_arg
+from globus_cli.termio import TextMode, display
+from globus_cli.utils import resolve_principal_urn
+
+_VALID_ROLES = ["administrator", "activity_manager", "activity_monitor"]
```
sirosen, line 15: Just as a cute idea, how do you like this?

```suggestion
_VALID_ROLES_T = t.Literal["administrator", "activity_manager", "activity_monitor"]
_VALID_ROLES: tuple[str, ...] = t.get_args(_VALID_ROLES_T)
```

sirosen (changes requested):
Per usual, this assignment is getting a 9/10 at least from the grader. :wink: 

I think there are some minor improvements we can make, noted, and I want to review the tests. I didn't quite get to those. But the implementation looks solid and I think that moving the existing capability for handling this into a simple helper is a very good "next adjacent move" for us. We may continue to refine this, but IMO it's a step in the right direction.

src/globus_cli/commands/gcs/endpoint/role/create.py
derek-globus: Sure I can make these. You may be able to tell but I wrote this docstring while UUIDs didn't assume that principal type was identity & apparently forgot to come back and update it.

src/globus_cli/utils.py
derek-globus: Sure

src/globus_cli/commands/gcs/endpoint/role/create.py
derek-globus: Sure

src/globus_cli/commands/gcs/endpoint/role/list.py
derek-globus: Can you guess which command I wrote first 😬

src/globus_cli/commands/gcs/endpoint/role/list.py
derek-globus: Yeah, I considered peeking under the hood to iterate on this design (no pun intended) but had already let a bit of scope creep into this change with the principal type resolution & didn't want to add more.

src/globus_cli/commands/gcs/endpoint/role/list.py
derek-globus: Sure, I'm happy with the rewrite.

One disagreement, I don't like using a generic `utils.py`. I did it in this PR because I couldn't think of a more specific place to put a common function like `resolve_principal_urn`.

This function however does have a specific usage pattern, which is termio formatting (there are 7 usage sites across the CLI, all of which are calls to `display(..., json_converter=iterable_response_to_dict)`), so I'd prefer to put it in the `termio` package.

Things that grow together go together.

derek-globus: Okay, I didn't rewrite but I removed the transfer import.

While moving this stuff around, I noticed that somebody (2022-sirosen) had added [basically duplicate functionality](https://github.com/globus/globus-cli/blob/main/src/globus_cli/utils.py#L153-L164) in the PagingWrapper itself so I used that instead.

The only notable distinction is that transfer's `iterable_response_to_dict` attempts to extract data from a `data` attribute from the item falling back to the item itself if a `data` attribute doesn't exist. For this usage, a `data` attribute doesn't exist so it's identical.

sirosen: > One disagreement, I don't like using a generic `utils.py` ...

I disagree with your disagreement! We actually agree! :upside_down_face: 

I view `utils.py` (and worse, `utils/`) as a code-smell in many cases. It too often represents "I've given up on namespacing this thing."
Waaay back in time I used to be a bit more of a maniac about it and try to actually eliminate a `utils.py` module from codebases. I've since grown more lenient -- especially for things like "functions I wish were in the stdlib or builtins in the context of this codebase." But there's a danger of growing too lax and starting to just dump your whole application into there. So, yeah, if we want output munging tools, `termio` is a nice place for them.


> I noticed that somebody (2022-sirosen)

I never did like that guy much. :stuck_out_tongue_winking_eye: 

I'm good with the way you rewrote it. We can, as we agree, iterate further in the future.

tests/unit/test_utils.py
```diff
@@ -94,3 +96,146 @@ def foo(bar):
...
+        assert (
+            urn == "urn:globus:auth:identity:64831427-e501-420b-93da-9e4efee3dd51"
+            or urn == "urn:globus:groups:id:449718d5-32b2-48ba-bd46-e045deab5430"
+        )
+    except click.UsageError as e:
+        assert expect_error
+        assert e.message == (
+            f"'--principal-type' was unspecified and '{principal}' was not resolvable "
+            f"to a globus identity."
+        )
```
sirosen, line 145: I'm finding this pattern in the tests, of trying, then having assertions in the `try`-block, and otherwise branching to the `except`-block, to be hard to read. I think the main thing is that the part which is common to both paths (`resolve_principal_urn(...)`) is grouped with assertions that are only relevant in half of the cases.

So, with that motivation in mind, I'm going to suggest this change, to move the `assert`s down into an else block:
```suggestion
    except click.UsageError as e:
        assert expect_error
        assert e.message == (
            f"'--principal-type' was unspecified and '{principal}' was not resolvable "
            f"to a globus identity."
        )
    else:
        assert not expect_error
        assert (
            urn == "urn:globus:auth:identity:64831427-e501-420b-93da-9e4efee3dd51"
            or urn == "urn:globus:groups:id:449718d5-32b2-48ba-bd46-e045deab5430"
        )
```

How would you feel about such a change across this whole test file?

I'm trying to respond to the fact that I was confused a bit on my first read, but I'll also note that I _did_ adjust and was able to understand all of the tests, so it's not critical.

derek-globus: I find the proposed structure a lot harder to read actually. It's pretty really unintituitve to me that I'd look for the good path after the error declaration in a different nested block (a reason I'd only use try/except/else in very extreme contexts).

Would explicit comments chunking out the sections make the reading experience better for you with the existing structure? I want to be very sympathetic to the plight of a code reader.

Something like 
```python
    #  === Setup ===
    auth_client = unittest.mock.Mock(spec=CustomAuthClient)
    resolved_id = "64831427-e501-420b-93da-9e4efee3dd51" if resolve_id else None
    auth_client.maybe_lookup_identity_id.return_value = resolved_id

    try:
        # === Act ===
        urn = resolve_principal_urn(
            auth_client=auth_client,
            principal=principal,
            principal_type=None,
        )

       # === Verify Expected Success ===
       assert not expect_error
       assert (
            urn == "urn:globus:auth:identity:64831427-e501-420b-93da-9e4efee3dd51"
            or urn == "urn:globus:groups:id:449718d5-32b2-48ba-bd46-e045deab5430"
       )
    except click.UsageError as e:
        # === Verify Expected Failure ===
        assert expect_error
        assert e.message == (
            f"'--principal-type' was unspecified and '{principal}' was not resolvable "
            f"to a globus identity."
        )
```

sirosen: I think the comments help a little but don't really resolve the readability quirk -- TBH, if that's the only option, I'd just as soon leave it all alone.

For me, the disorientation comes from the fact that `try-except` is actually encoding branching here -- which is not what it does as flow-control, but is what it's doing vis-a-vis test logic. `try-except-else` is a branching construct, so it logically fits what the test is doing much better (whether or not you find it readable is a separate matter from how the language defines these constructs).

Perhaps this is over-valuing terseness and avoiding repetition which would be an improvement to readability. The following reconstruction seems like the most straightforward way to encode this:
```python
    if expect_error:
        with pytest.raises(
            click.UsageError, 
            match=(
                f"'--principal-type' was unspecified and '{principal}' was not resolvable "
                "to a globus identity."
            ),
        ):
            resolve_principal_urn(
                auth_client=auth_client,
                principal=principal,
                principal_type=None,
            )
    else:
        urn = resolve_principal_urn(
            auth_client=auth_client,
            principal=principal,
            principal_type=None,
        )
        assert (
             urn == "urn:globus:auth:identity:64831427-e501-420b-93da-9e4efee3dd51"
             or urn == "urn:globus:groups:id:449718d5-32b2-48ba-bd46-e045deab5430"
        )
```

---

> It's pretty unintituitve to me to look for the happy path in a different nested block after the error declaration (a reason I'd only use try/except/else in very extreme contexts).

We could probably have a much richer/fuller conversation about `try-except-else-finally` sometime, but I think we should avoid drawing conclusions from test code regarding these constructs which we'd carry into production code.

If this were _not_ a test, I would be much more vigorously requesting that we scope down the `try` block to only the action which has some known chance of failing -- the considerations are different for the different contexts.

derek-globus: I don't think we're gonna agree on this one.

I'm fairly familiar with the branching logic of `try-except-else-finally`.
In this case, I think that the most formally accurate approach (`try-except-else`) is both:
 1. quite a bit more unintuitive to read
 2. only more formally accurate in an trivially unimpactful way (if of the two assertions somehow raises a click UsageError we'd apply error-expectation semantics which would fail with a slightly more inaccurate error path)

---

I think we're spending too long on this part of the PR though so I'll update to your `try/except/else` semantics as that's the most palatable of the options you're happy with for me & it's minor enough that I'd rather not spend more cycles on it.

New rev coming.

sirosen: All a very fair assessment. We might be too far apart on this to agree quickly, and it's too minor of a topic for us to devote too much time to it.

I would have also been okay with us agreeing to close this as "we don't agree and therefore we'll leave it as it was", but I appreciate you biasing towards my input as a reviewer.
Readability is very subjective, so it's not something I like to push too hard, but I am always pretty eager to discuss it. Rather than continuing to belabor this topic though... Approval incoming. :wink:

sirosen (approved):

PR #947, opened 2024-02-08
Add `--subscription-id` to `flows create`

sirosen:
This requires the latest SDK `main` version (doesn't work against the last release).

It adds the option, passes it through appropriately, and adds minimal new testing.

----

Marked as a draft until an SDK release + version bump here.


ada-globus (approved):
Looks like what I'd expect! Given that SDK released w/ this last week and the deadline is fast approaching, any chance we could get this merged and released today?

PR #949, opened 2024-02-15
Update pinned dependencies

sirosen:
- update the updater script (a pre-dependabot tool for updating our various pinned dependencies; still useful for a manual update) to handle changes to where dependencies are declared
- pin the `pre-commit` version in `tox.ini` -- more reproducible runs and valid as a target of the updater
- run the updater

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -59,7 +59,8 @@ commands_pre = coverage html --fail-under=0
 commands = coverage report
 
 [testenv:lint]
-deps = pre-commit==3.6.1
+deps = pre-commit
+recreate = true
 skip_install = true
 commands = pre-commit run --all-files
 
--- a/tox.ini
+++ b/tox.ini
@@ -59,7 +59,7 @@ commands_pre = coverage html --fail-under=0
 commands = coverage report
 
 [testenv:lint]
-deps = pre-commit~=2.9.2
+deps = pre-commit==3.6.1
 skip_install = true
 commands = pre-commit run --all-files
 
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -59,7 +59,7 @@ commands_pre = coverage html --fail-under=0
 commands = coverage report
 
 [testenv:lint]
-deps = pre-commit~=2.9.2
```
kurtmckee, line 62: I'm not yet convinced this should be pinned; I would anticipate that pinning the hook versions, not pre-commit, is acceptable here.

kurtmckee (approved):

tox.ini
sirosen: Yeah, I'm not sure. I almost wrote `<4` as the version bound. Do we want to go with that for now?

We sometimes need a lower bound (which can be implicit; we'll assume a relatively recent one if unspecified) to make sure newer features work, like `types_or`. I think we want an explicit upper bound in case there are new format changes (v3 was, I think, when the old config format was removed?) or other breaking changes which I can't foresee.

Give me an indication of which you'd prefer: pinned or `<4`. I can't think of alternatives, but if there is one that's compelling, that would be cool too.

kurtmckee: I think my gut here is "linting isn't a default environment, and pre-commit.ci is enabled for this repo".

Therefore I'd set `recreate = true` in the tox environment and then remove the version specifier completely. Tox will always recreate the environment and install the latest version, which will almost certainly identical behavior to what we'll be getting from pre-commit.ci anyway.

sirosen: I just circled back to this and made the change. I'm sort of on the fence now that I look at it -- maybe we should just remove `tox r -e lint`... but for now unpinned with `recreate=true` will do the trick!

kurtmckee (approved):

PR #950, opened 2024-02-15
Make `--format=JSON` output on errors "raw"

sirosen:
When JSON output is requested, error responses are no longer reformulated by the globus-cli's error presentation code if the error data is JSON-formatted. Instead, in these cases the JSON output of the CLI is the complete error output.

Also simplify a slightly overcomplicated path by which `write_error_info` becomes a simple `click.echo` wrapper.

---

This is a relatively significant change in outputs, but only under error scenarios.
Manual testing demonstrates that this works well for complex error outputs (e.g. Flow validation failures, Search errors, etc).

### DIFF
--- a/changelog.d/20240214_225218_sirosen_fix_json_error_presentation.md
+++ b/changelog.d/20240214_225218_sirosen_fix_json_error_presentation.md
@@ -1,0 +1,5 @@
+### Enhancements
+
+* `--format=JSON` output now offers greater detail when API errors are
+  encountered with JSON bodies. Rather than re-encoding error details, the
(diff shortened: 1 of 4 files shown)


ada-globus (approved):
Just to confirm my understanding: because it is declared first, the `json_error_handler` will have the highest precedence, right?

sirosen:
Yep, under the hood it's a module-scoped list and we just "append" to register hooks, and "iterate until we find a match" to resolve an error to a hook. So order matters and first match wins. :slightly_smiling_face: 

I'll sort out that mypy error.

PR #951, opened 2024-02-16
Added globus gcs endpoint show and update

derek-globus:
[sc-28586](https://app.shortcut.com/globus/story/28586/cli-implement-globus-gcs-endpoint-update)

## Testing
 1. Unit tested
 2. Verified lots of manual update/show situations against sandbox-hosted GCSv5 endpoint. 

## Visual

Here's what it looks like to "show" an endpoint (response structure is identical to update)

```
globus-cli % globus gcs endpoint update $EP --network-use aggressive
Endpoint ID:                  379f7156-c996-4de5-a228-db2a4d872f89
Display Name:                 Dereks Cool GCSv5.4 Sandbox Endpoint
Allow UDT:                    False
Contact Email:                derek@globus.org
Contact Info:                 401 North Michigan Ave
Department:                   Automate
Description:                  This is my endpoint. There are many like it but this one is mine.
Earliest Last Access:         2023-12-15
GCS Manager URL:              https://0629c.5898.sandbox2.zones.dnsteam.globuscs.info
GridFTP Control Channel Port: 443
Info Link:                    None
Keywords:                     Derek,Derk,MyCoolEndpoint
Network Use:                  aggressive
Network Use (Concurrency):    aggressive
Network Use (Parallelism):    aggressive
Organization:                 Globus
Public:                       True
Subscription ID:              4fa609fc-14f6-4c62-acfd-680873d3b6fe
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ CLI_VERSION=$(shell grep '^__version__' src/globus_cli/version.py | cut -d '"' -
 	virtualenv --python=$(PYTHON_VERSION) .venv
 	.venv/bin/pip install -U pip setuptools
 	.venv/bin/pip install -e '.[development]'
+	.venv/bin/pip install -e '.[test]'
 
 .PHONY: localdev
 localdev: .venv
--- a/changelog.d/20240216_122016_derek_gcs_endpoint_update_sc_28586.md
+++ b/changelog.d/20240216_122016_derek_gcs_endpoint_update_sc_28586.md
@@ -1,0 +1,6 @@
+
+### Enhancements
+
+* [sc-28586](https://app.shortcut.com/globus/story/28586/cli-implement-globus-gcs-endpoint-update) Added support for GCSv5 endpoint displaying & updating:
+  * `globus gcs endpoint show ENDPOINT_ID`
+  * `globus gcs endpoint update ENDPOINT_ID`
--- a/changelog.d/20240216_122016_derek_gcs_endpoint_update_sc_28586.md
+++ b/changelog.d/20240216_122016_derek_gcs_endpoint_update_sc_28586.md
@@ -1,6 +1,6 @@
 
 ### Enhancements
 
-* [sc-28586](https://app.shortcut.com/globus/story/28586/cli-implement-globus-gcs-endpoint-update) Added support for GCSv5 endpoint displaying & updating:
+* Added support for GCSv5 endpoint displaying & updating:
   * `globus gcs endpoint show ENDPOINT_ID`
   * `globus gcs endpoint update ENDPOINT_ID`
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ classifiers = [
 ]
 requires-python = ">=3.8"
 dependencies = [
-    "globus-sdk==3.35.0",
+    "globus-sdk==3.37.0",
     "click>=8.1.4,<9",
     "jmespath==1.0.1",
     "packaging>=17.0",
--- a/src/globus_cli/commands/gcs/endpoint/_common.py
+++ b/src/globus_cli/commands/gcs/endpoint/_common.py
@@ -1,0 +1,25 @@
+from globus_cli.termio import Field, formatters
+
+# https://docs.globus.org/globus-connect-server/v5.4/api/schemas/Endpoint_1_2_0_schema/
+GCS_ENDPOINT_FIELDS = [
+    Field("Endpoint ID", "id"),
+    Field("Display Name", "display_name"),
+    Field("Allow UDT", "allow_udt", formatter=formatters.FuzzyBool),
+    Field("Contact Email", "contact_email"),
+    Field("Contact Info", "contact_info"),
+    Field("Department", "department"),
+    Field("Description", "description"),
+    Field("Earliest Last Access", "earliest_last_access", formatter=formatters.Date),
+    Field("GCS Manager URL", "gcs_manager_url"),
+    Field("GridFTP Control Channel Port", "gridftp_control_channel_port"),
+    Field("Info Link", "info_link"),
+    Field("Keywords", "keywords"),
+    Field("Max Concurrency", "max_concurrency"),
+    Field("Max Parallelism", "max_parallelism"),
+    Field("Network Use", "network_use"),
+    Field("Organization", "organization"),
+    Field("Preferred Concurrency", "preferred_concurrency"),
+    Field("Preferred Parallelism", "preferred_parallelism"),
+    Field("Public", "public", formatter=formatters.FuzzyBool),
+    Field("Subscription ID", "subscription_id"),
+]
--- a/src/globus_cli/commands/gcs/endpoint/_common.py
+++ b/src/globus_cli/commands/gcs/endpoint/_common.py
@@ -1,4 +1,41 @@
+from __future__ import annotations
+
+import typing as t
+
 from globus_cli.termio import Field, formatters
+from globus_cli.termio.formatters import FieldFormatter
+
+
+class NetworkUseFormatter(FieldFormatter[t.Union[str, t.Tuple[int, int]]]):
+    """
+    Custom Formatter to make network use associated fields better grouped.
+
+    Data is expected to be passed as a list of three elements:
+      [network_use, preferred, max]
+
+    Examples:
+      ("custom", 1, 2) -> "Preferred: 1, Max: 2"
+      ("aggressive", None, None) -> "aggressive"
+      ("normal", None, 3) -> "normal"
+    """
+
+    def parse(self, value: t.Any) -> str | tuple[int, int]:
+        if isinstance(value, list) and len(value) == 3:
+            if value[0] == "custom":
+                if isinstance(value[1], int) and isinstance(value[2], int):
+                    return value[1], value[2]
+            elif isinstance(value[0], str):
+                return value[0]
+
+        raise ValueError(
+            f"Invalid network use data shape. Expected [str, int, int]; found {value}."
+        )
+
+    def render(self, value: str | tuple[int, int]) -> str:
+        if isinstance(value, tuple):
+            return f"Preferred: {value[0]}, Max: {value[1]}"
+        return value
+
 
 # https://docs.globus.org/globus-connect-server/v5.4/api/schemas/Endpoint_1_2_0_schema/
 GCS_ENDPOINT_FIELDS = [
@@ -14,12 +51,18 @@ GCS_ENDPOINT_FIELDS = [
     Field("GridFTP Control Channel Port", "gridftp_control_channel_port"),
     Field("Info Link", "info_link"),
     Field("Keywords", "keywords", formatter=formatters.Array),
-    Field("Max Concurrency", "max_concurrency"),
-    Field("Max Parallelism", "max_parallelism"),
     Field("Network Use", "network_use"),
+    Field(
+        "Network Use (Concurrency)",
+        "[network_use, preferred_concurrency, max_concurrency]",
+        formatter=NetworkUseFormatter(),
+    ),
+    Field(
+        "Network Use (Parallelism)",
+        "[network_use, preferred_parallelism, max_parallelism]",
+        formatter=NetworkUseFormatter(),
+    ),
     Field("Organization", "organization"),
-    Field("Preferred Concurrency", "preferred_concurrency"),
-    Field("Preferred Parallelism", "preferred_parallelism"),
     Field("Public", "public", formatter=formatters.FuzzyBool),
     Field("Subscription ID", "subscription_id"),
 ]
--- a/src/globus_cli/commands/gcs/endpoint/_common.py
+++ b/src/globus_cli/commands/gcs/endpoint/_common.py
@@ -13,7 +13,7 @@ GCS_ENDPOINT_FIELDS = [
     Field("GCS Manager URL", "gcs_manager_url"),
     Field("GridFTP Control Channel Port", "gridftp_control_channel_port"),
     Field("Info Link", "info_link"),
-    Field("Keywords", "keywords"),
+    Field("Keywords", "keywords", formatter=formatters.Array),
     Field("Max Concurrency", "max_concurrency"),
     Field("Max Parallelism", "max_parallelism"),
     Field("Network Use", "network_use"),
--- a/src/globus_cli/commands/gcs/endpoint/set_subscription_id.py
+++ b/src/globus_cli/commands/gcs/endpoint/set_subscription_id.py
@@ -35,7 +35,7 @@ class GCSSubscriptionIdType(click.ParamType):
             self.fail(msg, param, ctx)
 
 
-@command("set-subscription-id", short_help="Set a GCS Endpoint's subscription")
+@command("set-subscription-id", short_help="Set a GCS Endpoint's subscription.")
 @endpoint_id_arg
 @click.argument("SUBSCRIPTION_ID", type=GCSSubscriptionIdType())
 @LoginManager.requires_login("transfer")
--- a/src/globus_cli/commands/gcs/endpoint/set_subscription_id.py
+++ b/src/globus_cli/commands/gcs/endpoint/set_subscription_id.py
@@ -35,7 +35,7 @@ class GCSSubscriptionIdType(click.ParamType):
             self.fail(msg, param, ctx)
 
 
-@command("set-subscription-id", short_help="Set an GCS Endpoint's subscription")
+@command("set-subscription-id", short_help="Set a GCS Endpoint's subscription")
 @endpoint_id_arg
 @click.argument("SUBSCRIPTION_ID", type=GCSSubscriptionIdType())
 @LoginManager.requires_login("transfer")
--- a/src/globus_cli/commands/gcs/endpoint/set_subscription_id.py
+++ b/src/globus_cli/commands/gcs/endpoint/set_subscription_id.py
@@ -35,7 +35,7 @@ class GCSSubscriptionIdType(click.ParamType):
             self.fail(msg, param, ctx)
 
 
-@command("set-subscription-id", short_help="Update an endpoint's subscription")
+@command("set-subscription-id", short_help="Set an GCS Endpoint's subscription")
 @endpoint_id_arg
 @click.argument("SUBSCRIPTION_ID", type=GCSSubscriptionIdType())
 @LoginManager.requires_login("transfer")
--- a/src/globus_cli/commands/gcs/endpoint/show.py
+++ b/src/globus_cli/commands/gcs/endpoint/show.py
@@ -1,0 +1,22 @@
+import uuid
+
+from globus_cli.commands.gcs.endpoint._common import GCS_ENDPOINT_FIELDS
+from globus_cli.login_manager import LoginManager
+from globus_cli.parsing import command, endpoint_id_arg
+from globus_cli.termio import TextMode, display
+
+
+@command("show")
+@endpoint_id_arg
+@LoginManager.requires_login("transfer")
+def show_command(
+    login_manager: LoginManager,
+    *,
+    endpoint_id: uuid.UUID,
+) -> None:
+    """Describe a GCS Endpoint"""
+    gcs_client = login_manager.get_gcs_client(endpoint_id=endpoint_id)
+
+    res = gcs_client.get_endpoint()
+
+    display(res, text_mode=TextMode.text_record, fields=GCS_ENDPOINT_FIELDS)
--- a/src/globus_cli/commands/gcs/endpoint/show.py
+++ b/src/globus_cli/commands/gcs/endpoint/show.py
@@ -14,7 +14,7 @@ def show_command(
     *,
     endpoint_id: uuid.UUID,
 ) -> None:
-    """Describe a GCS Endpoint"""
+    """Display information about a particular GCS Endpoint."""
     gcs_client = login_manager.get_gcs_client(endpoint_id=endpoint_id)
 
     res = gcs_client.get_endpoint()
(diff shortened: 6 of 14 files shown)


Makefile
```diff
@@ -5,6 +5,7 @@ CLI_VERSION=$(shell grep '^__version__' src/globus_cli/version.py | cut -d '"' -
 	virtualenv --python=$(PYTHON_VERSION) .venv
 	.venv/bin/pip install -U pip setuptools
 	.venv/bin/pip install -e '.[development]'
+	.venv/bin/pip install -e '.[test]'
```
sirosen, line 8: I forgot this part of the Makefile even existed.
I may want to suggest removing it later.

changelog.d/20240216_122016_derek_gcs_endpoint_update_sc_28586.md
```diff
@@ -0,0 +1,6 @@
+
+### Enhancements
+
+* [sc-28586](https://app.shortcut.com/globus/story/28586/cli-implement-globus-gcs-endpoint-update) Added support for GCSv5 endpoint displaying & updating:
```
sirosen, line 4:
```suggestion
* Added support for GCSv5 endpoint displaying & updating:
```

Because the CLI changelog is end-user facing, we should not include links to our internal tracking here. It invites users to click on a link they can't follow.

pyproject.toml
```diff
@@ -28,7 +28,7 @@ classifiers = [
 ]
 requires-python = ">=3.8"
 dependencies = [
-    "globus-sdk==3.35.0",
+    "globus-sdk==3.37.0",
```
sirosen, line 31: I just pushed through a general dependency update (today? yesterday? it's a blur), so this will probably be an annoying conflict. It's the same change for this line though, so :crossed_fingers: `git` could handle it gracefully.

src/globus_cli/commands/gcs/endpoint/_common.py
```diff
@@ -0,0 +1,25 @@
...
+    Field("Department", "department"),
+    Field("Description", "description"),
+    Field("Earliest Last Access", "earliest_last_access", formatter=formatters.Date),
+    Field("GCS Manager URL", "gcs_manager_url"),
+    Field("GridFTP Control Channel Port", "gridftp_control_channel_port"),
+    Field("Info Link", "info_link"),
+    Field("Keywords", "keywords"),
+    Field("Max Concurrency", "max_concurrency"),
+    Field("Max Parallelism", "max_parallelism"),
+    Field("Network Use", "network_use"),
```
sirosen, line 19: Can we do some custom format work for the network use parameters, since concurrency and parallelism are only meaningful if `network_use='custom'`?

We have four other parameters here, so I'm imagining something like...
```
Network Use:               custom
Network Use (concurrency): prefer=8 max=32
Network Use (parallelism): prefer=16 max=16
```
as well as
```
Network Use:               aggressive
Network Use (concurrency): aggressive
Network Use (parallelism): aggressive
```

src/globus_cli/commands/gcs/endpoint/set_subscription_id.py
```diff
@@ -35,7 +35,7 @@ def convert(
             self.fail(msg, param, ctx)
 
 
-@command("set-subscription-id", short_help="Update an endpoint's subscription")
+@command("set-subscription-id", short_help="Set an GCS Endpoint's subscription")
```
kurtmckee, line 38:
```suggestion
@command("set-subscription-id", short_help="Set a GCS Endpoint's subscription")
```

src/globus_cli/commands/gcs/endpoint/show.py
```diff
@@ -0,0 +1,22 @@
...
+
+@command("show")
+@endpoint_id_arg
+@LoginManager.requires_login("transfer")
+def show_command(
+    login_manager: LoginManager,
+    *,
+    endpoint_id: uuid.UUID,
+) -> None:
+    """Describe a GCS Endpoint"""
```
kurtmckee, line 17: I anticipated something akin to "Show information about a GCS Endpoint". Does the "show"-command/"describe"-text match other art?

src/globus_cli/commands/gcs/endpoint/_common.py
```diff
@@ -0,0 +1,25 @@
...
+    Field("Allow UDT", "allow_udt", formatter=formatters.FuzzyBool),
+    Field("Contact Email", "contact_email"),
+    Field("Contact Info", "contact_info"),
+    Field("Department", "department"),
+    Field("Description", "description"),
+    Field("Earliest Last Access", "earliest_last_access", formatter=formatters.Date),
+    Field("GCS Manager URL", "gcs_manager_url"),
+    Field("GridFTP Control Channel Port", "gridftp_control_channel_port"),
+    Field("Info Link", "info_link"),
+    Field("Keywords", "keywords"),
```
sirosen, line 16: I think that GCS returns keywords as a list. I don't fully remember how we made formatting work. Do we need to specify an array formatter for this?

src/globus_cli/commands/gcs/endpoint/update.py
```diff
@@ -0,0 +1,244 @@
...
+from globus_cli.constants import EXPLICIT_NULL, ExplicitNullType
+from globus_cli.login_manager import LoginManager
+from globus_cli.parsing import CommaDelimitedList, command, endpoint_id_arg
+from globus_cli.parsing.param_types.nullable import IntOrNull
+from globus_cli.termio import TextMode, display
+
+
+class SubscriptionIdType(click.ParamType):
+    def get_type_annotation(self, _: click.Parameter) -> type:
+        return t.cast(type, str | ExplicitNullType)
```
kurtmckee, line 20: The cast here, together with a generic type, surprises me but may reflect a misunderstanding on my part.

Is it possible to avoid the cast by writing the annotation as `-> type[str | ExplicitNullType]` and then returning that? Or `t.Type[str | ExplicitNullType]`?

src/globus_cli/commands/gcs/endpoint/update.py
```diff
@@ -0,0 +1,244 @@
...
+    login_manager: LoginManager,
+    *,
+    endpoint_id: uuid.UUID,
+    allow_udt: bool | None,
+    contact_email: str | None,
+    contact_info: str | None,
+    department: str | None,
+    description: str | None,
+    display_name: str | None,
+    gridftp_control_channel_port: str | None,
```
sirosen, line 204: I'm reading the options correctly for this, yeah?

```suggestion
    gridftp_control_channel_port: int | ExplicitNullType | None,
```

src/globus_cli/commands/gcs/endpoint/update.py
```diff
@@ -0,0 +1,244 @@
...
+@click.option(
+    "--gridftp-control-channel-port",
+    type=IntOrNull(),
+    help="The TCP port which the Globus control channel should listen on.",
+)
+@click.option(
+    "--info-link",
+    type=str,
+    help=(
+        "An end-user-facing url for a webpage with more information about the endpoint."
```
kurtmckee, line 117:
```suggestion
        "An end-user-facing URL for a webpage with more information about the endpoint."
```

src/globus_cli/termio/formatters/primitive.py
```diff
@@ -14,16 +14,22 @@ def render(self, value: str) -> str:
         return value
 
 
-class DateFormatter(FieldFormatter[datetime.datetime]):
-    def parse(self, value: t.Any) -> datetime.datetime:
+class DateFormatter(FieldFormatter[datetime.datetime | datetime.date]):
+    def parse(self, value: t.Any) -> datetime.datetime | datetime.date:
```
sirosen, line 18: To ask a provocative question, should we have one type which handles both, or should we rename
```
DateFormatter -> DatetimeFormatter
```
and add a new `DateFormatter`?

I recognize that the rename is lots-of-lines-of-change, but it's probably pretty simple to do. So it's more about what we want to have in our library of formatters.

sirosen (changes requested):
Tests and the core of this look great, but I have some minor changes requested.
The fields and formatting in particular caught my eye this time.

kurtmckee (changes requested):

src/globus_cli/termio/formatters/primitive.py
kurtmckee: A good idea that I would not want to see pulled into this well-scoped PR. :pray:

src/globus_cli/commands/gcs/endpoint/update.py
sirosen: This may be copied from my prior art elsewhere. I can offer some relevant info: `types.UnionType` is not a `type`.

I think this derives in some way from the difference between a `UnionType` at runtime and any "normal" `type`. For example, you can't instantiate a `UnionType`, as in `(int | str)("1")`.

It may be that `type` is the wrong return-type for this, and `click-type-test` might need some tinkering to make it all work nicely. But once the typing folks start talking about "special forms" I generally get scared and want to go back to thinking about the language I know (the runtime one). :wink:

src/globus_cli/termio/formatters/primitive.py
sirosen: Good call! Let's consider this scoped as "future improvements" unless Derek offers a differing opinion.

src/globus_cli/commands/gcs/endpoint/update.py
kurtmckee: > the difference between a `UnionType` at runtime and any "normal" `type`.

_That's_ exactly the kind of misunderstanding that I was expecting. Thanks for clarifying!

Makefile
derek-globus: What alternative path would you suggest for contributers setting up their dev environment?

changelog.d/20240216_122016_derek_gcs_endpoint_update_sc_28586.md
derek-globus: Sure.

pyproject.toml
derek-globus: Sure.

src/globus_cli/commands/gcs/endpoint/_common.py
derek-globus: I'm not sure what I think about that suggestion.

It definitely looks better than the current version. Is the improved view simplicity sufficiently worth munging the response structure like that?

src/globus_cli/commands/gcs/endpoint/_common.py
derek-globus: It currently renders like this:

image

---

With `Field("Keywords", "keywords", formatter=formatters.Array)` it renders like this

image

src/globus_cli/commands/gcs/endpoint/update.py
derek-globus: Ah correct, good catch. New Rev.

src/globus_cli/termio/formatters/primitive.py
derek-globus: ¯\_(ツ)_/¯

src/globus_cli/commands/gcs/endpoint/set_subscription_id.py
derek-globus: Sure.

Makefile
sirosen: It's mostly OT from this PR, but it depends on what flavor of "dev environment" you mean. For a runnable local copy of the CLI as you work on it, I'd suggest
```shell
python -m venv .venv
.venv/bin/pip install -e .
. .venv/bin/activate
```
as probably all you need.

For test dependencies, I'd typically tell people to rely on `tox`, but of course that fails to acknowledge workflow tools and IDEs which want a test environment like VSCode and PyCharm.

Maybe this is the right place for this to live? But then it needs to be maintained and we should get rid of `.[development]` which is now wrong (and which I think `pip` is just polite about ignoring).

The fact that nothing in CI is using this means it's very easy for it to drift out of date or get broken. If it's supported, even just as a dev workflow, I think we should think about testing it in CI.

src/globus_cli/commands/gcs/endpoint/show.py
derek-globus: I think it just depends on who wrote it:

**sirosen**
* [Display information about an index](https://github.com/globus/globus-cli/blob/main/src/globus_cli/commands/search/index/show.py#L14)
* [Print information detailing the status and other info about a task. The task may be pending, completed, or in progress.](https://github.com/globus/globus-cli/blob/main/src/globus_cli/commands/task/show.py#L184C5-L186C56)
* [Show a flow](https://github.com/globus/globus-cli/blob/main/src/globus_cli/commands/flows/show.py#L15)

**Derek**
* [Describe a particular role on a GCS Endpoint.](https://github.com/globus/globus-cli/blob/main/src/globus_cli/commands/gcs/endpoint/role/show.py#L21)

**Kurt**
* [Display information about a particular timer.](https://github.com/globus/globus-cli/blob/main/src/globus_cli/commands/timer/show.py#L17)

**Aaron**
* [Show a group definition](https://github.com/globus/globus-cli/blob/main/src/globus_cli/commands/group/show.py#L14)

src/globus_cli/commands/gcs/endpoint/update.py
derek-globus: Sure.

pyproject.toml
derek-globus: Sorry, I misread this as you asking me to update to 3.37.0...
But then I went to update to 3.37.0 and I was already on it!

Looks like no merge conflicts between this and upstream-main?

src/globus_cli/commands/gcs/endpoint/_common.py
derek-globus: Updating rendering to `formatters.Array`. New Rev

src/globus_cli/commands/gcs/endpoint/_common.py
sirosen: > Is the improved view simplicity sufficiently worth munging the response structure like that?

I'm not sure if you mean "is the improvement worth it because we're supposed to reflect API response data" or "is it worth the effort/complexity for us"?

I don't think there is or should be any expectation that CLI text-mode outputs directly reflect API response data. We're already doing "work" to responses in a lot of cases, and I think this one isn't particularly much different from many others (though I can see an argument about this).

As for whether or not it's worth the complexity in terms of effort, I don't think it's that hard, if we get a little clever.
`Field` does its key lookup using `jmespath`, so I think we need a custom format and then the fields become something like
```python
Field(
    "Network Use (concurrency)",
    "[prefer_concurrency, max_concurrency]",
    formatter=FancyFormatter()
),
Field(
    "Network Use (parallelism)",
    "[prefer_parallelism, max_parallelism]",
    formatter=FancyFormatter()
),
```

Modulo whatever parts of that I got subtly wrong.

src/globus_cli/commands/gcs/endpoint/update.py
derek-globus: Looks like no.

I think this has to do with mypy's evaluation of Unions in particular but I'll be the first to admit that I'm simply [following the traditions of my ancestors](https://github.com/globus/globus-cli/blob/main/src/globus_cli/parsing/param_types/json_strorfile.py#L62).

---

**Code**
```python
    def get_type_annotation(self, _: click.Parameter) -> type[str | ExplicitNullType]:
        return str | ExplicitNullType
```

**Mypy Error**
```
error: Incompatible return value type (got "UnionType", expected "type[str] | type[ExplicitNullType]")  [return-value]
```

src/globus_cli/commands/gcs/endpoint/show.py
sirosen: I am in favor of this new helptext because it apparently doesn't quite match any of the existing helptext. :upside_down_face: 

I would not want to block this change on any concern about things drifting even further apart, but if Derek (as the current PR author) wants to pick a direction and lead us all towards more uniform shorthelp for our commands, I'm more than happy to follow.

Makefile
derek-globus: Sure, considering it non-blocking.

I'm open to testing it in CI but I'm pretty opposed to the idea of taking away a contributor's one-click "get your environment set up to start contributing."

And yeah, I added `[test]` in this situation specifically because my IDE wants an environment that includes our test dependencies if I'm gonna add unit tests. I'm open to driving that virtualenv creation through tox or whatever tool makes the most sense but I'm not open to removing it.

src/globus_cli/commands/gcs/endpoint/update.py
kurtmckee: I think `str | ExplicitNullType` indicates an instance of either type will be returned. Perhaps:

```python
def get_type_annotation(self, _: click.Parameter) -> type[str | ExplicitNullType]:
    return type[str | ExplicitNullType]
```

But I'm not trying to get stuck on this, just understand what's possible / specific.

derek-globus: **Code**
```python
    def get_type_annotation(self, _: click.Parameter) -> type[str | ExplicitNullType]:
        return type[str | ExplicitNullType]
```

**Mypy Errors**
```
error: Returning Any from function declared to return "type[str] | type[ExplicitNullType]"  [no-any-return]
```

```
error: The type "type[type]" is not generic and not indexable  [misc]
```

src/globus_cli/commands/gcs/endpoint/show.py
derek-globus: I'm not gonna be the one to spearhead helptext wording cause I don't think I'm very good at that.

But if there's a common pattern (ideally enforced as much as possible by linter) I can follow it!

src/globus_cli/commands/gcs/endpoint/update.py
sirosen: I would not want to change the runtime value which is returned here without a more deliberate process about getting that new style working in the click type checking. `click-type-test` is expecting the `types.UnionType` value to be returned, and isn't expecting an instance of `type.__class_getitem__` (does that produce a `typing.Type`?).

`click-type-test` would need to be updated with whatever change and potentially use `get_origin` or `get_args` to traverse the value being returned. It's something which could possibly get cleaned up, but I think that's a bigger conversation to be had.

And it may turn out that when we go to cases like this one to cleanup, we find that we can remove `get_type_annotation` hook entirely. The [v0.0.6](https://github.com/sirosen/click-type-test/releases/tag/0.0.6) features should cover most of our cases (which is why this hook is gone from most of our custom types).

src/globus_cli/commands/gcs/endpoint/show.py
derek-globus: Having said that though, I like Kurt's wording better than mine so I updated it to `Display information about a particular GCS Endpoint.`

src/globus_cli/commands/gcs/endpoint/_common.py
sirosen: I also love writing code! :grin: 

I think this is within bounds for us transforming outputs. I think that there are so many fields in the GCS output that the majority of users will benefit from us compacting things a bit, and we always have the crutch of `--format JSON` for folks who want the "real data".

derek-globus: Sure, I agree more than I disagree.

Plus now I get to write more of that code stuff!

derek-globus: image

kurtmckee (approved):

src/globus_cli/commands/gcs/endpoint/_common.py
derek-globus: Alright, here's what it looks like

```
globus-cli % globus gcs endpoint show $EP
Endpoint ID:                  379f7156-c996-4de5-a228-db2a4d872f89
Display Name:                 Dereks Cool GCSv5.4 Sandbox Endpoint
Allow UDT:                    False
Contact Email:                derek@globus.org
Contact Info:                 401 North Michigan Ave
Department:                   Automate
Description:                  This is my endpoint. There are many like it but this one is mine.
Earliest Last Access:         2023-12-15
GCS Manager URL:              https://0629c.5898.sandbox2.zones.dnsteam.globuscs.info
GridFTP Control Channel Port: 443
Info Link:                    None
Keywords:                     Derek,Derk,MyCoolEndpoint
Network Use:                  custom
Network Use (Concurrency):    Preferred: 3, Max: 5
Network Use (Parallelism):    Preferred: 2, Max: 8
Organization:                 Globus
Public:                       True
Subscription ID:              4fa609fc-14f6-4c62-acfd-680873d3b6fe
```

```
globus-cli % globus gcs endpoint update $EP --network-use aggressive
Endpoint ID:                  379f7156-c996-4de5-a228-db2a4d872f89
Display Name:                 Dereks Cool GCSv5.4 Sandbox Endpoint
Allow UDT:                    False
Contact Email:                derek@globus.org
Contact Info:                 401 North Michigan Ave
Department:                   Automate
Description:                  This is my endpoint. There are many like it but this one is mine.
Earliest Last Access:         2023-12-15
GCS Manager URL:              https://0629c.5898.sandbox2.zones.dnsteam.globuscs.info
GridFTP Control Channel Port: 443
Info Link:                    None
Keywords:                     Derek,Derk,MyCoolEndpoint
Network Use:                  aggressive
Network Use (Concurrency):    aggressive
Network Use (Parallelism):    aggressive
Organization:                 Globus
Public:                       True
Subscription ID:              4fa609fc-14f6-4c62-acfd-680873d3b6fe

```

kurtmckee: > Keywords:                     Derek,Derk,MyCoolEndpoint

This perfectly mirrors my own endpoints.

sirosen (approved):
I love how the NetworkUseFormatter implementation came out. I wouldn't have expected it to work so nicely -- kudos to our past selves for really working hard on the field and formatter design!

PR #956, opened 2024-02-26
Add 'group get-by-subscription' command

sirosen:
The behavior of this command is specced out currently as follows:
- first, it will fetch the subscription_info document
- next, it will check: has text output been requested?
  - only in this case, fetch the relevant group
  - display the group info
- if fetching the group failed OR we weren't asked for text output, display the subscription_info document
  - if fetching failed, show a hint to stderr that the group is not visible

Tests attempt to exercise all of the relevant cases.

Some important nuances:
- we detect "could not fetch the group" via a 403 -- all other errors, including 404s, are treated as "true errors"
- make sure we don't print the command hint when someone runs with `--format=json` and we didn't even try a lookup
- this behavior ensures that JSON output will work even if fetching the group would fail hard
- for initial display, we include no connector info (it can always be added later)

For now this is implemented without using an underlying SDK method for getting subscription_info (as that is not yet released), but this can be updated as soon as support is available.

### DIFF
--- a/changelog.d/20240226_144957_sirosen_get_group_by_subscription_id.md
+++ b/changelog.d/20240226_144957_sirosen_get_group_by_subscription_id.md
@@ -1,0 +1,4 @@
+### Enhancements
+
+* Add `globus group get-by-subscription` for looking up Subscription Groups
+  based on their subscription IDs.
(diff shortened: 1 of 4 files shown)


kurtmckee:
I think the SDK support for getting a group by subscription ID been released. If that's so, I'd like to see that used here.

kurtmckee (approved):

PR #961, opened 2024-03-01
Remove autoactivate calls from the CLI

sirosen:
These calls to Transfer don't do anything in the post-GCSv4 universe. Remove the helper which implemented them, and all of its call sites.

One test for handling of service-401s relied on the fact that this was the first call to Transfer for `ls`. As a result, it had to be rewritten -- in the course of doing so, rearrange the relevant test data into a more self-contained test module for authentication failures.


kurtmckee (commented):
I don't currently agree with this change -- it doesn't look like we're removing the CLI options to skip the activation check, for example.

It may be that we need to simply push out a new CLI major version with activation code fully and properly removed.

aaschaer (approved):

sirosen:
> I don't currently agree with this change -- it doesn't look like we're removing the CLI options to skip the activation check, for example.

The `skip_autoactivation_check` is actually an option to the Transfer API which we're passing through as well.
I'd like to remove it too, but I am _very_ concerned about what will happen as we remove activation. I want to move slowly and methodically, only making one change at a time and meticulously but gradually removing support.

Activation has been an expected and essential behavior for over a decade, and by removing behaviors I am worried that we will uncover corner cases in which it still is necessary (e.g. on GCP hosts; shared endpoints; something else unexpected). This change is therefore scoped to only remove the implicit autoactivate calls emitted by the CLI and not remove any of the numerous bits of support like `globus endpoint is-activated`, `globus endpoint activate`, etc. (those are hidden but still present)

PR #963, opened 2024-03-04
add stat command

aaschaer:
Shortcut: https://app.shortcut.com/globus/story/30451/sdk-cli-add-support-for-transfer-stat-operation
SDK: https://github.com/globus/globus-sdk-python/pull/961

Opening as draft since tests will fail until SDK support is live and being used here.

### DIFF
--- a/src/globus_cli/commands/stat.py
+++ b/src/globus_cli/commands/stat.py
@@ -39,7 +39,7 @@ $ globus stat $col_id:/home/share/godata/file1.txt
 def stat_command(
     login_manager: LoginManager,
     *,
-    endpoint_plus_path: tuple[uuid.UUID, str | None],
+    endpoint_plus_path: tuple[uuid.UUID, str],
     local_user: str | None,
 ) -> None:
     """
--- a/src/globus_cli/commands/stat.py
+++ b/src/globus_cli/commands/stat.py
@@ -1,0 +1,63 @@
+from __future__ import annotations
+
+import uuid
+
+import click
+import globus_sdk
+
+from globus_cli.login_manager import LoginManager
+from globus_cli.parsing import ENDPOINT_PLUS_REQPATH, command, local_user_option
+from globus_cli.termio import Field, TextMode, display
+
+STAT_FIELDS = [
+    Field("Name", "name"),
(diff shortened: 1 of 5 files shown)


tests/functional/test_stat.py
```diff
@@ -0,0 +1,28 @@
...
+Size:          4
+Permissions:   0644
+User:          tutorial
+Group:         tutorial
+"""
+    assert result.output == expected
+
+    req = get_last_request()
+    parsed_qs = urllib.parse.parse_qs(urllib.parse.urlparse(req.url).query)
+    assert parsed_qs == {"path": ["foo/"], "local_user": ["bar"]}
```
sirosen, line 28: Could you add two tests to this, to exercise the error paths in the command?
- 404 not-found case
- 5xx :boom: case

aaschaer: Added!

I used a 403 instead of a 500 since those are probably more common in the wild.

I also added the fixtures to the sdk here: https://github.com/globus/globus-sdk-python/pull/962 since they seemed like they might be generally useful

src/globus_cli/commands/stat.py
```diff
@@ -0,0 +1,63 @@
...
+
+""",
+)
+@click.argument("endpoint_plus_path", type=ENDPOINT_PLUS_REQPATH)
+@local_user_option
+@LoginManager.requires_login("transfer")
+def stat_command(
+    login_manager: LoginManager,
+    *,
+    endpoint_plus_path: tuple[uuid.UUID, str | None],
```
sirosen, line 42: I tried out the tests on this with the latest SDK release, and that allowed the annotation tests to run, which flagged this:
```suggestion
    endpoint_plus_path: tuple[uuid.UUID, str],
```

(Because the path is required, the type is `str`, not `str | None`.)

kurtmckee (commented):
The team's been discussing whether this should go into a top-level `globus stat` command or get tucked into a subcommand like `globus transfer-service` so that we're not continuing to put Transfer commands into the top-level namespace.

I'll contact you out-of-band to summarize the discussion.

sirosen (approved):

PR #965, opened 2024-03-05
Improve the internal output display API

sirosen:
This is a pair of changes aimed at cleanup and simplification.

---

- Combine interfaces for `display()` and `TextMode`

The interface change in `display` is best summarized by example:

```python
from globus_cli.termio import display, TextMode

display(foo, text_mode=TextMode.text_record)
```

is converted to

```python
from globus_cli.termio import display

display(foo, text_mode=display.RECORD)
```

In addition to the changes to `display`'s interface, the following additional adjustments are made:
- `display` is "well-typed" under mypy, meaning its usages can be type-checked
- `TextMode`, although still used internally, is no longer exported by `globus_cli.termio`

---

- Remove CLIStubResponse

This is an old remnant of an early version of the output formatting code. It's now just a weak dict wrapper, and safe to remove.

### DIFF
--- a/src/globus_cli/commands/cli_profile_list.py
+++ b/src/globus_cli/commands/cli_profile_list.py
@@ -7,7 +7,7 @@ import click
 
 from globus_cli.login_manager import is_client_login, token_storage_adapter
 from globus_cli.parsing import command
-from globus_cli.termio import Field, TextMode, display, formatters
+from globus_cli.termio import Field, display, formatters
 
 
 def _profilestr_to_datadict(s: str) -> dict[str, t.Any] | None:
@@ -97,7 +97,7 @@ def cli_profile_list(*, all: bool) -> None:
         ]
         if all:
             fields.append(Field("GLOBUS_SDK_ENVIRONMENT", "env"))
-        display(user_profiles, text_mode=TextMode.text_table, fields=fields)
+        display(user_profiles, text_mode=display.TABLE, fields=fields)
     if client_profiles:
         click.echo("")
         fields = [
@@ -106,4 +106,4 @@ def cli_profile_list(*, all: bool) -> None:
         ]
         if all:
             fields.append(Field("GLOBUS_SDK_ENVIRONMENT", "env"))
-        display(client_profiles, text_mode=TextMode.text_table, fields=fields)
+        display(client_profiles, text_mode=display.TABLE, fields=fields)
--- a/src/globus_cli/commands/_common.py
+++ b/src/globus_cli/commands/_common.py
@@ -6,7 +6,7 @@ import uuid
 
 import click
 
-from globus_cli.termio import TextMode, display
+from globus_cli.termio import display
 
 if t.TYPE_CHECKING:
     from ..services.transfer import CustomTransferClient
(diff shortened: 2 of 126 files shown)


src/globus_cli/commands/cli_profile_list.py
```diff
@@ -97,7 +97,7 @@ def cli_profile_list(*, all: bool) -> None:
         ]
         if all:
             fields.append(Field("GLOBUS_SDK_ENVIRONMENT", "env"))
-        display(user_profiles, text_mode=TextMode.text_table, fields=fields)
+        display(user_profiles, text_mode=display.TABLE, fields=fields)
```
kurtmckee, line 100: The default is to display a table. Does every call site specify the `text_mode` to use? If so, I suggest removing the default parameter value.

sirosen: I was thinking about this when I made the change. It may require a few touches, but I can't imagine many commands are relying on the default. Let's get rid of that and make it a required keyword-only param.

sirosen: Actually, it seems there are a few, but this is raising a separate interface improvement -- I think we can eliminate `simple_text` with a little bit of extra elbow-grease. Given that it complicates this interface, I'm going to give that cleanup a shot this evening.

sirosen: To explain how it relates, here's how it came up when I was looking at `globus whoami`:
```diff
@@ -81,11 +83,7 @@ def whoami_command(login_manager: LoginManager, *, linked_identities: bool) -> N
                     Field("ID", "sub"),
                     Field("Email", "email"),
                 ],
-                simple_text=(
-                    None
-                    if is_verbose()
-                    else "\n".join([x["username"] for x in res["identity_set"]])
-                ),
+                text_mode=display.TABLE if is_verbose() else _render_linked_identities,
             )
         except KeyError:
             click.echo(
@@ -99,12 +97,20 @@ def whoami_command(login_manager: LoginManager, *, linked_identities: bool) -> N
     else:
         display(
             res,
-            text_mode=display.RECORD,
+            text_mode=display.RECORD if is_verbose() else _render_preferred_username,
             fields=[
                 Field("Username", "preferred_username"),
                 Field("Name", "name"),
                 Field("ID", "sub"),
                 Field("Email", "email"),
             ],
-            simple_text=(None if is_verbose() else res["preferred_username"]),
         )
+
+
+def _render_linked_identities(data: t.Any) -> None:
+    for x in data["identity_set"]:
+        click.echo(x["username"])
+
+
+def _render_preferred_username(data: t.Any) -> None:
+    click.echo(data["preferred_username"])
```

sirosen:
I'm going to revert the last commit. It took me a while to finish and I wanted to circle back to look at it with a clear eye.
I don't "feel uncomfortable with it" -- I dislike it. It's more verbose and less clear.

It does leave me in the position of turning down a change to make `text_mode` a required field, at least for now.

src/globus_cli/commands/cli_profile_list.py
sirosen: Per my last comment, the attempted change, and the revert, are we okay punting this problem out of the scope of this PR?

I don't want to make it a required param because then usages with `simple_text` will have to specify it _and_ override it. I haven't noodled out the right interface fix for all of this yet.

For now, I want to stick with reducing the number of symbols being passed around (`display, TextMode` -> `display`) and making `display` into a callable object which opens up some interesting options for giving it properties or otherwise being a "smarter" implementation.

kurtmckee (approved):

sirosen:
I've just done a rebase, so I'm hoping we'll see clean CI and then I'll merge.

PR #971, opened 2024-04-08
Allow subscriptions to be assigned to existing flows

kurtmckee:
### Enhancements

* Allow subscriptions to be assigned to existing flows
  using the `globus flows update --subscription-id` option.

### DIFF
--- a/src/globus_cli/commands/flows/_common.py
+++ b/src/globus_cli/commands/flows/_common.py
@@ -1,3 +1,8 @@
+from __future__ import annotations
+
+import typing as t
+import uuid
+
 import click
 
 from globus_cli.parsing import JSONStringOrFile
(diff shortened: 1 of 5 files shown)


src/globus_cli/commands/flows/_common.py
```diff
@@ -102,3 +107,30 @@
...
+            return uuid.UUID(value)
+        except ValueError:
+            self.fail(f"{value} must be either a UUID or 'DEFAULT'", param, ctx)
+
+
+subscription_id_option = click.option(
+    "--subscription-id",
+    "subscription_id",
+    type=SubscriptionIdType(),
+    multiple=False,
```
sirosen, line 130: I would typically not specify `multiple=False`, as it's the default, but this definitely isn't harmful.

sirosen (approved):

PR #976, opened 2024-05-01
Add `globus flows validate` command

ada-globus:
Some affordances are provided for anticipated future output. In particular, the non-error output includes a way to render multiple sections corresponding to different analyses that were performed during the `/flows/validate` call. Importantly, `display()` is only called once, insuring proper behavior when formats (like JSON) are specifically requested.

## Expected Output (User scopes)

```
Discovered Scopes
=================
Identity | Scope
-------- | --------------------------------------------------------------------------------------------
User     | https://auth.globus.org/scopes/5fac2e64-c734-4e6b-90ea-ff12ddbf9653/transfer_collection_info
User     | https://auth.globus.org/scopes/5fac2e64-c734-4e6b-90ea-ff12ddbf9653/transfer_mkdir
User     | https://auth.globus.org/scopes/5fac2e64-c734-4e6b-90ea-ff12ddbf9653/transfer_ls
User     | https://auth.globus.org/scopes/actions.globus.org/transfer/delete
User     | https://auth.globus.org/scopes/actions.globus.org/transfer/transfer
```

## Expected Output (No scopes)

```
Discovered Scopes
=================
No scopes discovered
```

### DIFF
--- a/src/globus_cli/commands/flows/validate.py
+++ b/src/globus_cli/commands/flows/validate.py
@@ -1,0 +1,92 @@
+from __future__ import annotations
+
+import click
+import globus_sdk
+
+from globus_cli.commands.flows._common import input_schema_option
+from globus_cli.login_manager import LoginManager
+from globus_cli.parsing import JSONStringOrFile, ParsedJSONData, command
+from globus_cli.termio import Field, display
+
+
+@command("validate", short_help="Validate a flow definition")
+@click.argument(
+    "definition",
+    type=JSONStringOrFile(),
+    metavar="DEFINITION",
+)
+@input_schema_option
+@LoginManager.requires_login("flows")
+def validate_command(
+    login_manager: LoginManager,
+    *,
+    definition: ParsedJSONData,
+    input_schema: ParsedJSONData | None,
+) -> None:
+    """
+    Validate a flow definition (BETA).
+
+    DEFINITION is the JSON document that defines the flow's instructions.
+    The definition document may be specified inline, or it may be
+    a path to a JSON file.
+
+        Example: Inline JSON:
+
+        \b
+            globus flows validate \\
+            '{{"StartAt": "a", "States": {{"a": {{"Type": "Pass", "End": true}}}}}}'
+
+        Example: Path to JSON file:
+
+        \b
+            globus flows validate definition.json
+    """
+    payload = {}
+
+    # Ensure that the definition is a JSON object
+    if not isinstance(definition.data, dict):
+        raise click.UsageError("Flow definition must be a JSON object")
+    payload["definition"] = definition.data
+
+    # Ensure the input schema, if provided, is a JSON object
+    if input_schema is not None:
+        if not isinstance(input_schema.data, dict):
+            raise click.UsageError("--input-schema must be a JSON object")
+        payload["input_schema"] = input_schema.data
+
+    # Configure client
+    flows_client = login_manager.get_flows_client()
+
+    res = flows_client.validate_flow(**payload)
+
+    display(
+        res,
+        text_mode=_validate_flow_output_handler,
+    )
+
+
+def _validate_flow_output_handler(result: globus_sdk.GlobusHTTPResponse) -> None:
+    # Discovered scopes output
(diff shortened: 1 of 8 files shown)


src/globus_cli/commands/flows/validate.py
```diff
@@ -0,0 +1,92 @@
...
+
+    # Ensure that the definition is a JSON object
+    if not isinstance(definition.data, dict):
+        raise click.UsageError("Flow definition must be a JSON object")
+    payload["definition"] = definition.data
+
+    # Ensure the input schema, if provided, is a JSON object
+    if input_schema is not None:
+        if not isinstance(input_schema.data, dict):
+            raise click.UsageError("--input-schema must be a JSON object")
```
sirosen, line 54: As a not-right-now enhancement, let's make `JSONStringOrFile` capable of doing this for us, as in...
```python
JSONStringOrFile(require_dict=True)
```

src/globus_cli/termio/printer.py
```diff
@@ -200,6 +199,9 @@ def __call__(
             sort_json_keys=sort_json_keys,
         )
 
+    def render_table(self, iterable, fields, print_headers=True):
+        print_table(iterable, fields, print_headers)
```
sirosen, line 203: This is more or less what we discussed and therefore :+1: from me at its core.

I would like us to put an annotation on it (even if the iterable is `Iterable[Any]`), since this is a method we're exposing to the rest of the CLI code. Can we do that in this PR? (If it proves problematic, I can treat it as a fast-follow.)

tests/functional/flows/test_validate_flow.py
```diff
@@ -0,0 +1,186 @@
...
+        # Parse the headers and rows
+        headers = [header.strip() for header in lines[divider_index - 1].split(" | ")]
+        rows = lines[divider_index + 1 :]
+        # Turn the rows into a table data as a list of lists
+        table_data = [
+            [cell.strip() for cell in row.split(" | ")] for row in rows if row
+        ]
+        tables.append((headers, table_data))
+
+    return tables
```
sirosen, line 186: I have nothing against having another way of parsing our own output on hand, for testing, but I think we have a couple of these kicking around already. It would be nice to do a cleanup sweep and produce something like a centralized fixture.

e.g.
```python
@pytest.fixture
def output_parser():
    return GlobusCLIOutputParser()


class GlobusCLIOutputParser:
    def parse_table(data: str) -> ParsedTable: ...
    def iter_all_tables(data: str) -> Iterable[ParsedTable]: ...
    def parse_record(data: str) -> ParsedRecord: ...
    def really_cool_stuff_here(data: str) -> YourNameOnThisType: ...
```

src/globus_cli/commands/flows/validate.py
```diff
@@ -0,0 +1,92 @@
...
+    # Discovered scopes output
+    scopes_response = result.get("scopes")
+    # Beta API: Defend against the case where 'scopes' is no longer returned
+    if scopes_response is not None:
+        _validate_flow_scope_output_handler(scopes_response)
+
+
+def _validate_flow_scope_output_handler(scopes_response: dict[str, str]) -> None:
+    # Always include User and Flow scopes
+    scopes_fields = [Field("Identity", "identity"), Field("Scope", "scope")]
```
sirosen, line 78: Rather than "Identity", I think we should call this field "RunAs", for the connection with the Flows terminology.

"Identity" will look to people like it should be an Auth identity, which it isn't, and I think could create confusion.

sirosen (changes requested):
Mostly positive feedback. My only major reservation is `Identity` as the column name.

tests/functional/flows/test_validate_flow.py
ada-globus: Yeah, I didn't see one as I was writing. Agree—would prefer there are shared utilities for working with this.

src/globus_cli/termio/printer.py
ada-globus: Added `-> None`

sirosen (approved):

sirosen (approved):

PR #979, opened 2024-05-02
Set no-untyped-defs on termio.errors

sirosen:
Progress towards "actually strict" mypy settings.

The type for 'value' on an error field is *very* nebulous. In terms of usage today, `str | int | None` is sufficient. We can expand or refine this in the future -- potentially we want special handling for `None` anyway.

### DIFF
--- a/src/globus_cli/termio/errors.py
+++ b/src/globus_cli/termio/errors.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import json
 
 import click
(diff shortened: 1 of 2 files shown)


src/globus_cli/termio/errors.py
```diff
@@ -13,17 +15,17 @@ class PrintableErrorField:
 
     TEXT_PREFIX = "Globus CLI Error:"
 
-    def __init__(self, name, value, multiline=False):
+    def __init__(self, name: str, value: str | int | None, multiline: bool = False):
```
kurtmckee, line 18: I thought that mypy required that even `__init__()` methods needed a trailing `-> None` annotation. Maybe that was a past version, or just an incorrect memory...

kurtmckee (approved):

src/globus_cli/termio/errors.py
sirosen: No, you're correct! That's definitely a requirement of strict mode. Let me add it really quickly. I'm sure something in our settings (which are not so simple as `--strict`) is interfering, so we'd just have to come back to this later if I don't.

sirosen: I would have thought that `disallow-incomplete-defs = true` would catch this... Maybe a bug triggered by our config?
It's fixed now anyway.

PR #994, opened 2024-06-14
Add a '--quiet' option as an inverse of '-v/--verbose'

sirosen:
This introduces `--quiet` and refactors the verbosity control code to ensure good behavior.
Commit messages have full detail.

`-q` is not supported because it would conflict with some commands today (at least with `globus search query`).

This change is preparatory -- we have a feature request open for `--quiet` mode on `globus login` having specific effects.
I intend to satisfy that by making that behavior driven by `verbosity < 0`.

### DIFF
--- a/src/globus_cli/parsing/command_state.py
+++ b/src/globus_cli/parsing/command_state.py
@@ -66,11 +66,19 @@ class CommandState:
     def is_verbose(self) -> bool:
         return self.verbosity > 0
 
+    def is_quiet(self) -> bool:
+        return self.verbosity < 0
+
     def set_verbosity(self, value: int) -> None:
+        # short-circuit if verbosity is already below 0 -- this makes `--quiet` higher
+        # precedence than `--verbose` regardless of the order of application
+        if self.verbosity < 0:
+            return
+
         self.verbosity = value
 
-        # verbosity level 0: never warn, never log normal events
-        # (also covers negative/quiet modes, e.g. `--quiet`)
+        # min verbosity level: never warn, never log normal events
+        # (covers quiet modes, e.g. `--quiet`)
         if value <= 0:
             _warnings.simplefilter("ignore")
             _setup_logging(level="CRITICAL")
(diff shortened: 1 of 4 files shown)


src/globus_cli/parsing/command_state.py
```diff
@@ -140,48 +158,35 @@ def callback(ctx: click.Context, param: click.Parameter, value: t.Any) -> None:
 
 
 def verbose_option(f: F) -> F:
-    def callback(ctx: click.Context, param: click.Parameter, value: t.Any) -> None:
+    def callback(ctx: click.Context, param: click.Parameter, value: int) -> None:
         # set state verbosity value from option
         state = ctx.ensure_object(CommandState)
-        state.verbosity = value
+        state.set_verbosity(state.verbosity + value)
```
kurtmckee, line 164: I don't think this matches my expectations.

```
globus ... -vvv --quiet --quiet --quiet
```

appears to increment verbosity by 3, then decrement it by 3, resulting in a net sum of zero.

What was the original request associated with a `--quiet` option? I don't remember what the requestor was wanting to accomplish.

derek-globus: If I were a user, I think I'd be confused by this behavior too.

My expectation would be:
* `--verbose` puts it in "verbose mode"
* `--quiet` pus it in "quiet mode"
* `--verbose --quiet`; not sure if this is logical, maybe errors.

I'm not entirely thrown by multiple verboses increasing the verbosity even more, but I don't think we should be exposing verbose or quiet as mapping onto a + or - background verbosity level.

kurtmckee: I think this aligns with my expectations. There's a `-vv` stacking option, but not a `--quiet --quiet` possibility, and `-v --quiet` would be an error.

sirosen: Fishing around the various utilities I have on hand -- TBH I did this expecting that the counting mode would be the "norm" -- I find a variety of behaviors and no particularly clear/consistent expectation about how `--quiet` and `--verbose` interact. `ssh -vvv -q` exemplifies the kind of behavior described here, in which `-q` takes precedence and suppresses all debug output. `curl -s -v` is interesting in that the silent-mode and verbosity are entirely separate controls.

Regarding all of what's been said above, there's only one point where I strongly disagree:

> `--verbose --quiet`; not sure if this is logical, maybe errors.

> `-v --quiet` would be an error.

This seems like an un-ergonomic choice to me.

It's pretty natural if you have another command built in a var or script with passthrough, and then you're invoking it but want to adjust verbosity. i.e.
```shell
./my-cli-wrapper.sh --quiet
# or, similarly
$cmd --quiet
```

It's pretty common to have scripted tool operations and then augment or override the options provided in a later context.

---

My intent is to have `--quiet` set verbosity to `-1` and for `--verbose` and `--debug` to check if verbosity is `< 0` and be no-ops on verbosity if `--quiet` was set.
That makes `--quiet` "stronger" regardless of the order of option callback evaluation. It makes `--quiet --quiet`, `-v --quiet`, `-vvv --quiet` all equivalent to a single `--quiet`.

sirosen: I've implemented the above, but would like to wait until I get some signoff before marking this thread resolved in case my perspective here is contentious.

kurtmckee (approved):
I'm okay with this.

PR #997, opened 2024-06-21
Add silent mode for `globus login --quiet`

sirosen:
When `globus login --quiet` is run and the user is already logged in,
no output will be shown. Effectively, this "enables quiet mode" for
logins.


kurtmckee (approved):
I ❤️ it when a plan comes together!

PR #1002, opened 2024-07-03
Paginate `globus gcs collection list` command

ada-globus:
Additionally, add a `--limit` option.

"Tutorial Collection 1" now has enough guest collections to exceed the GCS default of 16.

For posterity:
```
for i in $(seq 1 15); do
  globus collection create guest 6c54cade-bde5-45c1-bdea-f4bd71dba2cc /home/u_5ovepnp5hbhaxflmbikcgve27e/ "Ada Guest Test $i";
done
```

Signed-off-by: Ada <ada@globus.org>

### DIFF
--- a/tests/functional/collection/test_collection_list.py
+++ b/tests/functional/collection/test_collection_list.py
@@ -30,6 +30,16 @@ def test_collection_list(run_line, add_gcs_login, base_command):
         assert name in result.stdout
 
 
+def test_collection_list_limit(run_line, add_gcs_login, base_command):
+    meta = load_response_set("cli.collection_operations").metadata
+    epid = meta["endpoint_id"]
+    add_gcs_login(epid)
+    result = run_line(f"{base_command} {epid} --limit 1")
+    lines = result.stdout.splitlines()
+    assert len(lines) == 3  # header + 1 collection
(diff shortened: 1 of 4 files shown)


tests/functional/collection/test_collection_list.py
```diff
@@ -30,6 +30,16 @@ def test_collection_list(run_line, add_gcs_login, base_command):
         assert name in result.stdout
 
 
+def test_collection_list_limit(run_line, add_gcs_login, base_command):
```
kurtmckee, line 33: This tests the limit, but is there a test that exercises the pagination across a larger limit than what's returned?

For example, if the first call returns 16 but the requested limit is 25, is there a test that exercises the pagination and guarantees that the next call, also returning 16, is clipped to only 9 entries?

ada-globus: There isn't, at least not directly. I actually didn't see tests exercising similar scenario elsewhere this paradigm is used in the CLI, so hadn't included it here and instead opted to manually exercise it. I'll look into it.

ada-globus: Added!

kurtmckee (approved):

PR #1007, opened 2024-07-23
Add flake8-comprehensions to pre-commit

sirosen:
This is added to our configuration to match globus-sdk-python and help ensure consistent styling.

The changes were primarily applied via editor macros, as a large portion were conversions of `dict(x=1)` calls to `{"x": 1}` literals.
A scan for any accidental semantic changes is warranted, though the changes are somewhat large in area.

The first two commits handle the `.pre-commit-config.yaml` config and the `src/` tree, which is an ideal area for focused review.
The rest of the changes apply to tests.


kurtmckee (approved):
I'm mostly okay with this, but I'd prefer if this was modified by a code formatter, not complained about by a linter.

sirosen:
As I mentioned in chat, I do want to try to cover some of these rules with `slyp`. That said, if there's a formatter which does this job for us, I'm all for it!

I would probably still want this in place after such an addition though (at least, via the `slyp` path). I think it will be easy to add autofixing for a lot of the trivial cases, but not for some of the more elaborate ones. e.g. I think we can autofix the `dict()` to dict-literal rewrites, but I'm not sure how to write a safe fixer for
```diff
- k: v for (k, v) in {"x": y, ...}.items()
+ k: v for (k, v) in (("x", y), ...)
```

So I expect `flake8-comprehensions` to continue to deliver value, even if we add fixers.

PR #1009, opened 2024-07-27
Add more flake8 plugin rules to enforce short_help consistency

sirosen:
## What

Expand the CLI's flake8 plugin (`globus-cli-flake8`) with the following new rules aimed at command helptext:

- CLI003 single-line function docstring did not end in '.'
- CLI004 short_help string did not end in '.'
- CLI005 command function short_help too long
- CLI006 command function implicit short_help does not end in '.'
- CLI007 command function missing expected docstring

## Why

We regularly note inconsistent styling across CLI command help as an area for improvement.
These inconsistencies are usually trivial to resolve but hard to handle systematically.
`<insert long and boring discussion of why linters and style guides are useful here/>`

These new lint rules are specifically aimed at `short_help`, not the full `help` text attached to each command.
`short_help` is used to render lists of subcommands.

The goal here is to begin enforcing consistent styles.

### Example improvement

We consider the following example output substandard:
```
$ globus toon
Usage: globus toon [OPTIONS] COMMAND [ARGS]...

  Frob the snorks! Demuddle the smurfs!

Commands:
  gargamel  Demuddle gargamel.
  smurf     Demuddle the smurf of your choice, allowing for...
  occy      Frob occy
  snork     Frob the snork of your choice.
```

The preamble is fine, but the style of the command short_help for each subcommand is not consistent.
`globus toon gargamel` and `globus toon snork` are correct.
Each is a short phrase ending in a period.
`globus toon smurf` is too long, and truncates.
`globus toon occy` is missing the trailing period.

Better command listing would be:
```
Commands:
  gargamel  Demuddle gargamel.
  smurf     Demuddle the smurf of your choice.
  occy      Frob occy.
  snork     Frob the snork of your choice.
```

The rules obviously can't enforce this exact text, but they will guide us towards it.

### Why Not Exclude `tests/`, utilities, etc

At first glance, some of the changes here seem entirely unnecessary.
For example, docstrings in helper modules have been updated, even though these have nothing to do with helptext.

A balance needs to be struck between complexity for the flake8 plugin and its ability to filter out "irrelevant" content.
Even simple rules -- e.g., "don't descend into functions named `test_*`" -- increase implementation complexity and in many cases may prove to be error prone.

Because the plugin is not a major area of focus for our team, I decided to keep the implementation as simple as possible.
Suffering through a very small number of false-positives is a tradeoff we're making here because the cost of obeying these lint rules is very low.

## Review Notes

In order to keep `flake8` passing, we must apply these changes in the same PR which updates the plugin.
This is a very large number of simple, rote changes, which were mostly applied with editor macros.
My intent is to squash-merge the result, so we don't need to worry about the exact commit history.

The history here, in this PR, demonstrates part of how this change was executed.
CLI003 and CLI004 are added at the start of the process, and refined when CLI005, CLI006, and CLI007 were added later.
Chunks of the src/ and tests/ trees were modified on an ongoing basis, largely by means of piped `flake8` output.
e.g., `flake8 src/globus_cli/commands/bookmark/ | cut -d':' -f1 | uniq | xargs -o vim`
This should help give us confidence in the process used, that no extraneous changes were introduced.

### Notable exception: group help

After working up this PR (and opening it initially), I found that one of the targets for it failed.
[aea93d5](https://github.com/globus/globus-cli/pull/1009/commits/aea93d529d3ecb61480ab17ca06ffc370b72d576) fixes the command, which passes linting but demonstrates the ellipsis behavior which we're trying to avoid.

What happens in this case is that the short_help is 58 chars long (below the 60 limit used here), but the command name is long. Because the gap this reveals is caused not only by the associated command-name but also from _sibling command names_, I have decided not to try to address it here.

This counterexample undercuts some of the value of this new linting, but there are still many demonstrated cases in which it _does_ help us fix our helptext.

### DIFF
--- a/src/globus_cli/commands/endpoint/user_credential/create/s3.py
+++ b/src/globus_cli/commands/endpoint/user_credential/create/s3.py
@@ -12,7 +12,7 @@ from globus_cli.termio import display
 from .._common import user_credential_create_and_update_params
 
 
-@command("s3", short_help="Create a User Credential for an S3 Storage Gateway")
+@command("s3")
 @endpoint_id_arg
 @user_credential_create_and_update_params(create=True)
 @click.argument("s3_key_id")
@@ -30,7 +30,7 @@ def s3(
     display_name: str | None,
 ) -> None:
     """
-    Create a User Credential for an S3 Storage Gateway
+    Create a User Credential for an S3 Storage Gateway.
     """
     gcs_client = login_manager.get_gcs_client(endpoint_id=endpoint_id)
     auth_client = login_manager.get_auth_client()
--- a/src/globus_cli/commands/endpoint/user_credential/delete.py
+++ b/src/globus_cli/commands/endpoint/user_credential/delete.py
@@ -7,7 +7,7 @@ from globus_cli.termio import display
 from ._common import user_credential_id_arg
 
 
-@command("delete", short_help="Delete a specific User Credential on an Endpoint")
+@command("delete", short_help="Delete a specific User Credential on an Endpoint.")
 @endpoint_id_arg
 @user_credential_id_arg()
 @LoginManager.requires_login("auth", "transfer")
@@ -18,7 +18,7 @@ def user_credential_delete(
     user_credential_id: uuid.UUID,
 ) -> None:
     """
-    Delete a specific User Credential on a given Globus Connect Server v5 Endpoint
+    Delete a specific User Credential on a given Globus Connect Server v5 Endpoint.
     """
     gcs_client = login_manager.get_gcs_client(endpoint_id=endpoint_id)
 
--- a/src/globus_cli/commands/group/get_by_subscription.py
+++ b/src/globus_cli/commands/group/get_by_subscription.py
@@ -24,7 +24,7 @@ def group_get_by_subscription(
     """
     Show the group which provides a specific subscription.
 
-    If the group is not visible to the current user, only the Group ID will be shown.
+    If the group is not visible to the current user, only the group ID will be shown.
     """
     groups_client = login_manager.get_groups_client()
 
--- a/src/globus_cli/commands/group/get_by_subscription.py
+++ b/src/globus_cli/commands/group/get_by_subscription.py
@@ -13,14 +13,18 @@ from ._common import GROUP_FIELDS_W_SUBSCRIPTION, SUBSCRIPTION_FIELDS
 
 
 @click.argument("subscription_id", type=click.UUID)
-@command("get-by-subscription")
+@command(
+    "get-by-subscription",
+    short_help="Show the group for a specific subscription.",
+)
 @LoginManager.requires_login("groups")
 def group_get_by_subscription(
     login_manager: LoginManager, *, subscription_id: uuid.UUID
 ) -> None:
-    """Show the Group which provides a specific Subscription.
+    """
+    Show the group which provides a specific subscription.
 
-    If the Group is not visible to the current user, only the Group ID will be shown.
+    If the group is not visible to the current user, only the Group ID will be shown.
     """
     groups_client = login_manager.get_groups_client()
 
--- a/src/globus_cli/commands/group/get_by_subscription.py
+++ b/src/globus_cli/commands/group/get_by_subscription.py
@@ -61,7 +61,7 @@ def group_get_by_subscription(
 def try_resolve_group(
     groups_client: globus_sdk.GroupsClient, group_id: str
 ) -> globus_sdk.GlobusHTTPResponse | None:
-    """Attempt to get a group"""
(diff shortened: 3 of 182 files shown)


src/globus_cli/commands/endpoint/user_credential/delete.py
```diff
@@ -7,7 +7,7 @@
 from ._common import user_credential_id_arg
 
 
-@command("delete", short_help="Delete a specific User Credential on an Endpoint")
+@command("delete", short_help="Delete a specific User Credential on an Endpoint.")
```
kurtmckee, line 10: In some places, the `short_help` parameter is deleted rather than updated. Should this be removed, too?

src/globus_cli/commands/endpoint/user_credential/create/s3.py
```diff
@@ -12,7 +12,7 @@
 from .._common import user_credential_create_and_update_params
 
 
-@command("s3", short_help="Create a User Credential for an S3 Storage Gateway")
+@command("s3")
```
kurtmckee, line 15: Example of a `short_help` parameter being removed.

src/globus_cli/commands/group/get_by_subscription.py
```diff
@@ -13,14 +13,18 @@
...
 @LoginManager.requires_login("groups")
 def group_get_by_subscription(
     login_manager: LoginManager, *, subscription_id: uuid.UUID
 ) -> None:
-    """Show the Group which provides a specific Subscription.
+    """
+    Show the group which provides a specific subscription.
 
-    If the Group is not visible to the current user, only the Group ID will be shown.
+    If the group is not visible to the current user, only the Group ID will be shown.
```
kurtmckee, line 27:
```suggestion
    If the group is not visible to the current user, only the group ID will be shown.
```

src/globus_cli/commands/search/index/create.py
```diff
@@ -14,7 +14,7 @@
 def create_command(
     login_manager: LoginManager, *, display_name: str, description: str
 ) -> None:
-    """Create a new Index"""
+    """Create a new Index."""
```
kurtmckee, line 17: If our standard for groups, flows, and timers are lowercase, should this be lowercase as well?

src/globus_cli/commands/search/index/list.py
```diff
@@ -12,7 +12,7 @@
 @command("list")
 @LoginManager.requires_login("search")
 def list_command(login_manager: LoginManager) -> None:
-    """List indices where you have some permissions"""
+    """List indices where you have some permissions."""
```
kurtmckee, line 15: For example, "indices" is lowercase!

src/globus_cli/commands/search/task/list.py
```diff
@@ -14,11 +14,11 @@
 ]
 
 
-@command("list", short_help="List recent Tasks for an index")
+@command("list", short_help="List recent Tasks for an index.")
```
kurtmckee, line 17:
```suggestion
@command("list", short_help="List recent tasks for an index.")
```

src/globus_cli/commands/search/task/show.py
```diff
@@ -23,7 +23,7 @@
 @task_id_arg
 @LoginManager.requires_login("search")
 def show_command(login_manager: LoginManager, *, task_id: uuid.UUID) -> None:
-    """Display a Task"""
+    """Display a Task."""
```
kurtmckee, line 26:
```suggestion
    """Display a task."""
```

kurtmckee (changes requested):

src/globus_cli/commands/endpoint/user_credential/delete.py
sirosen: In this case, the `short_help` is very slightly different from the first line of the docstring, which is why it's kept separate here.

I've tried to only make more extensive changes when it has 0 effect on the text or is strictly necessary. (But, of course, when working through a large changeset like this one it's hard to stay disciplined.)

src/globus_cli/commands/search/index/create.py
sirosen: Yeah, I would like us to lowercase these simply for consistency's sake. (I don't feel very strongly about what convention we use, so long as I have some kind of guideline to follow.)

I'll apply the changes you've suggested on this front, but I want to avoid trying to enforce that universally in the scope of this PR. IMO including these changes is a little bit of (acceptable) scope creep but shouldn't be a blocker for getting this PR merged. We may need another pass over these anyway -- I want to see if `click` can support boldface ANSI codes in helptext, which is the styling we're using on the docs site.

kurtmckee (approved):

PR #1017, opened 2024-08-01
Added --delete flag to timer create transfer

derek-globus:
## What?
Added a `--delete` flag to `globus timer create transfer` to match the one in `globus transfer` ([docs](https://docs.globus.org/cli/reference/transfer/#options)).

### Note
This flag does not perform a "deletion" task but rather performs a "transfer" task with "delete_destination_extra" set to true.

From the [docs](https://docs.globus.org/api/transfer/task_submit/#transfer_specific_fields), this field requests that the task:
>Delete files, directories, and symlinks in the destination directory which don’t exist on the source directory or are a different type. Only applies for recursive directory transfers. Default: false.

I chose to use the flag name `--delete` instead of `--delete-destination-extra` to maintain consistency with the regular transfer cli command.

## Testing

Created a timer which ran once, performing a transfer that delete extra files on the destination. Observed timer run and transfer task success.

```zsh
globus timer create transfer --delete $MY_GCP:/~/gcp/subdir/ $MY_GCP:/~/gcp/subdir2/ --stop-after-runs=1
```

### DIFF
--- a/src/globus_cli/commands/timer/create/transfer.py
+++ b/src/globus_cli/commands/timer/create/transfer.py
@@ -55,16 +55,6 @@ e.g. '1h30m', '500s', '10d'
 """
 
 
-def resolve_optional_local_time(
-    start: datetime.datetime | None,
-) -> datetime.datetime | globus_sdk.utils.MissingType:
-    if start is None:
-        return globus_sdk.MISSING
-    # set the timezone to local system time if the timezone input is not aware
-    start_with_tz = start.astimezone() if start.tzinfo is None else start
-    return start_with_tz
-
-
 @command("transfer", short_help="Create a recurring transfer timer.")
 @click.argument(
     "source", metavar="SOURCE_ENDPOINT_ID[:SOURCE_PATH]", type=ENDPOINT_PLUS_OPTPATH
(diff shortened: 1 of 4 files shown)


src/globus_cli/commands/timer/create/transfer.py
```diff
@@ -108,6 +98,15 @@ def resolve_optional_local_time(
     help="Stop running the transfer after this number of runs have happened.",
 )
 @mutex_option_group("--stop-after-date", "--stop-after-runs")
+@click.option(
+    "--delete",
```
m1yag1, line 102: I wonder if this should be named "--delete-destination-extra" instead of simply "--delete" to keep the explict-ness.

Additionally, in the help section, it's unclear what "extraneous files in the destination directory" means. Perhaps, using what's in the [transfer docs](https://docs.globus.org/api/transfer/task_submit/#recursive_directory) would be better?

> Used to provide "mirroring" by deleting any files in the destination directory not in the source.

m1yag1 (approved):
Looks good to me! I added a comment about the command name and the help but I defer to your expertise in what things should be named in the CLI and what would be considered a good help message.

src/globus_cli/commands/timer/create/transfer.py
m1yag1: Oh just read your comment again and saw you chose specifically to use `--delete`.

derek-globus: Yeah, I definitely think that `delete-destination-extra` is a much better flag than `delete`. I just don't want this command to diverge at all from the other one.

How about we add this & I'll file a story to deprecate `--delete` in favor of `--delete-destination-extra`. We'll need to support a long deprecation tail on that keyword no matter what but this lets us sync up the commands as they are right now.

---

For the docs, I'm happy to update. I'll update the docs for `globus transfer --delete` as well though

derek-globus: Updated the shorttext in [3524f82](https://github.com/globus/globus-cli/pull/1017/commits/3524f820843c9c9a027f33bd1347c3849b98971b). 

It's not identical to the transfer docs you linked cause it's missing some of the surrounding context in that style of doc. Lemme know what you think.

PR #1022, opened 2024-08-17
Stop deleting the templated client on logout

ada-globus:
image


kurtmckee (approved):

sirosen (approved):
I just read down the path I was worried about -- login with no tokens in the presence of the templated client -- and it's all good. The entire codepath follows a "get-or-create" paradigm. That was my only hesitation, so I think we're good to go!

PR #1024, opened 2024-08-28
Remove more activation support code

sirosen:
This was flagged by a conversation about some trailing 'activation' text in another context.
Upon review, we still have some vestiges of activation support.

Because `globus endpoint is-activated`, in particular, may be in use in long-standing scripts, these commands are given some extra deference.
They are already marked deprecated and hidden from help text and docs, but now they have their original functions replaced with an echo to stderr with a warning about upcoming removal.
This is even gentler than a removal (or, equivalently, a usage error), and precedes a likely upcoming release in which these commands will be outright removed.

---

- `activate`, `deactivate`, and `is-activated` now all warn to stderr when used, and no longer have any other effects -- tests are updated to check this
- `--skip-activation-check` for task submission commands is now hidden -- and commented as a deprecated option but not otherwise changed
- remove activation helper functions
- remove the autoactivate success fixture data from tests
- remove `Activated` from `globus endpoint show` text output


changelog.d/20240828_101834_sirosen_activation2stub.md
```diff
@@ -0,0 +1,12 @@
+### Other
+
+* The CLI has removed remaining support for activation.
```
kurtmckee, line 3:
```suggestion
* The CLI has removed remaining support for endpoint activation.
```

src/globus_cli/commands/endpoint/deactivate.py
```diff
@@ -1,18 +1,25 @@
...
 @endpoint_id_arg
-@LoginManager.requires_login("transfer")
-def endpoint_deactivate(login_manager: LoginManager, *, endpoint_id: uuid.UUID) -> None:
+def endpoint_deactivate(*, endpoint_id: uuid.UUID) -> None:
     """
-    Remove the credential previously assigned to an endpoint via
-    'globus endpoint activate' or any other form of endpoint activation
+    Deactivate an endpoint.
+
+
```
kurtmckee, line 14:
```suggestion
```

src/globus_cli/commands/list_commands.py
```diff
@@ -43,11 +43,6 @@ def list_commands() -> None:
     """
     Prints the name and a short description of every command available in the globus
     cli. Commands are grouped by their parent commands,
```
kurtmckee, line 45: I think that the example could be updated, rather than deleting it.

If it's to stay deleted, though, this needs to convert to a period.

```suggestion
    cli. Commands are grouped by their parent commands.
```

kurtmckee (changes requested):

src/globus_cli/commands/list_commands.py
sirosen: Let's make it a period! I forgot to mention this in the PR commentary: when I checked the command for another example, I found that the text was simply inaccurate!

kurtmckee (approved):

PR #1030, opened 2024-09-17
Fix a whitespace issue in the `globus collection update` help text

kurtmckee:
### Bugfixes

* Fix a whitespace issue in the `globus collection update` help text.

_This also makes the text use full sentences and reflows some of the lines._


src/globus_cli/commands/collection/update.py
```diff
@@ -106,9 +105,9 @@ def get_value(self, data: t.Any) -> t.Any:
     multiple=True,
     callback=emptyable_opt_list_callback,
     help=(
-        "Connector-specific username denied permission to create guest "
-        f"collections. {_MULTI_USE_OPTION_STR} to deny multiple users. "
-        'Set a value of "" to clear this'
+        "Connector-specific username denied permission to create guest collections. "
+        "{_MULTI_USE_OPTION_STR} to deny multiple users. "
```
sirosen, line 109:
```suggestion
        f"{_MULTI_USE_OPTION_STR} to deny multiple users. "
```

sirosen (changes requested):
I think we should maybe keep the clickbait PR titles out of our public repos, but I'm generally good with the change itself. Found one small (but important) typo.

src/globus_cli/commands/collection/update.py
kurtmckee: Incorporated as an amend.

sirosen (approved):

PR #1032, opened 2024-09-26
Remove the redundant tox `skip_missing_interpreters` setting

kurtmckee:
This PR removes the redundant tox `skip_missing_interpreters` setting in `tox.ini`.
This codifies the recently-discovered fact that it's true by default.

<sub>This PR was generated using [turbolift](https://github.com/Skyscanner/turbolift).</sub>

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -6,7 +6,6 @@ envlist =
     cov-combine
     cov-report
     mypy
-skip_missing_interpreters = true
 minversion = 4.3.5
 
 [testenv]


sirosen (approved):
This must have changed in v4, right? I remember tox failing on missing interpreters in the past, I'm almost certain.

kurtmckee:
I'm very certain that it used to fail, but it seems to be chugging along happily now.

The docs are unhelpful here, though:

https://tox.wiki/en/4.20.0/config.html#skip_missing_interpreters

> The default value is `config`
>
> Setting it to `config` means that the value is read from the config file.

Not sure what to make of that.

PR #1052, opened 2024-11-08
add size and checksum fields to task show --successful-transfers

pjhinton-globus:
Add the following fields to the tabular output of

globus task show --successful-transfers

* size
* checksum_algorithm
* checksum

If there is more that needs to be done from a testing or documentation standpoint, please let me know as this is my first foray into the CLI development way of doing things. :-)

refs sc-27244


derek-globus:
We use scriv to generate changelog fragments along with each external change; this change is simple enough so I added it for you.

Once build passes, I'll merge

derek-globus (approved):

PR #1060, opened 2024-12-06
Refactor to scope the lifecycle of SQLite storage

sirosen:
The storage adapter is here rewritten entirely away from a "pure functional style" in which attributes and values live in global scopeinto a more object-oriented style. The reason for this refactor is that we want to better control the lifecycle of the storage and ensure that it is closed cleanly when the program exits; furthermore, we want to simulate that same open/close behavior in the testsuite.

The overall tack taken here is
- wrap SQLiteAdapter in a new container type, CLITokenstorage
- give CLITokenstorage as many passthrough methods as are necessary to make this transition easy as a rewrite for various usage sites
- take steps to rewrite usages which do not have a handle on a LoginManager, and use the LoginManager as a way of getting access to the token storage
- start work on testsutie fixes

This approach has proven more successful than alternatives which were attempted. At time of writing, `mypy` is passing on the result, some interactive testing seems to work, and tests are in need of numerous adjustments, but not so severe as some other failed rewrites:

  50 failed, 977 passed, 13 errors

Therefore, there should be at most 63 remediation steps, after which this should be a functioning option for us.

---

Opened as a draft PR to share my work as I progress, and collect early feedback.


sirosen:
I've finished getting the refactor in working order. Most notably, in the final change here I finally achieved one of the goals here, which is to have `click.Context.call_on_close` close the sqlite connection. This is setup as part of `LoginManager.requires_login`.

I don't have new tests which ensure that we use the new behavior yet. I'm still thinking a little about whether or not to add new tests and what they should contain. Given that all of the existing tests are passing, I think we can call this ready for review while I think about that.

kurtmckee (approved):

src/globus_cli/login_manager/tokenstore.py
```diff
@@ -10,15 +11,218 @@
...
+    """
+    A wrapper over the globus-sdk's v1 tokenstorage which provides simplified
+    capabilities specific to the CLI.
+    """
+
+    def __init__(self) -> None:
+        self._adapter: SQLiteAdapter | None = self._construct_adapter()
+
+    def _construct_adapter(self) -> SQLiteAdapter:
+        from globus_sdk.tokenstorage import SQLiteAdapter  # noqa: F811
```
derek-globus, line 30: Some local testing showed that this import statement takes about 80ms; is that too high of a penalty?

src/globus_cli/login_manager/tokenstore.py
```diff
@@ -10,15 +11,218 @@
...
+
+        # when initializing the token storage adapter, check if the storage file exists
+        # if it does not, then use this as a flag to clean the old config
+        fname = _get_storage_filename()
+        if not os.path.exists(fname):
+            from ._old_config import invalidate_old_config
```
derek-globus, line 36: Did the same eyeballing with local testing here and it's in the single digit milliseconds added to run this import.

src/globus_cli/login_manager/tokenstore.py
```diff
@@ -10,15 +11,218 @@
...
+    def internal_native_client(self) -> globus_sdk.NativeAppAuthClient:
+        """
+        This is the client that represents the CLI itself (prior to templating).
+        """
+        template_id = _template_client_id()
+        return globus_sdk.NativeAppAuthClient(
+            template_id, app_name="Globus CLI (native client)"
+        )
+
+    def internal_auth_client(self) -> globus_sdk.ConfidentialAppAuthClient:
```
derek-globus, line 57: The name `internal_*` sounds like it's exposing the internals of the class. What do you think about using the term `common_*`? Or even just drop the prefix (`token_storage.native_client`)?

Additionally I'd recommend using `confidential` rather than `auth` in the second one to more accurately highlight the differences between these two. 


```suggestion
    def common_native_client(self) -> globus_sdk.NativeAppAuthClient:
        """
        This is the client that represents the CLI itself (prior to templating).
        """
        template_id = _template_client_id()
        return globus_sdk.NativeAppAuthClient(
            template_id, app_name="Globus CLI (native client)"
        )

    def common_confidential_client(self) -> globus_sdk.ConfidentialAppAuthClient:
```

src/globus_cli/login_manager/tokenstore.py
```diff
@@ -10,15 +11,218 @@
...
+            client_secret = credential_data["secret"]
+
+            self.store_well_known_config(
+                "auth_client_data",
+                {"client_id": client_id, "client_secret": client_secret},
+            )
+
+        return globus_sdk.ConfidentialAppAuthClient(
+            client_id, client_secret, app_name="Globus CLI"
+        )
```
derek-globus, line 91: Is it worth thinking about caching this value?

I realize we want to do the `is_client_login` check, but could still throw it in an instance variable and return it if it's already been instantiated.

src/globus_cli/login_manager/tokenstore.py
derek-globus: _Alternatively if these are not meant to be shared, I'd recommend `_native_client` and `_confidential_client`_

src/globus_cli/login_manager/tokenstore.py
```diff
@@ -10,15 +11,218 @@
...
+        # this could overwrite data from another CLI version *earlier or later* than
+        # the current one
+        #
+        # in the case that the old data was from a prior version, this makes sense
+        # because we have added new constraints or behaviors
+        #
+        # if the data was from a *newer* CLI version than what we are currently
+        # running we can't really know with certainty that "downgrading" the version
+        # numbers is correct, but because we can't know we need to just do our best
+        # to indicate that the tokens in storage may have lost capabilities
```
derek-globus, line 198: How are we indicating that the version was downgraded? Looks like we're just overriding with whatever the current version is no matter what.

derek-globus (commented):
Commenting early

src/globus_cli/login_manager/tokenstore.py
sirosen: We discussed! In the past, `globus_sdk.tokenstorage` was part of the "bad path" for eventually importing `requests`, but we fixed that. So now I'm going to let the eager import happen and we should be fine. :smile:

src/globus_cli/login_manager/tokenstore.py
sirosen: I'll try moving this, but I note that the module in question loads `GLOBUS_SDK_ENVIRONMENT` at import time and might not respond well in the testsuite to the change.

If it works, cool. If not, I'll revert and comment back here with details.

src/globus_cli/login_manager/tokenstore.py
sirosen: We talked it through; going with `cli_{thing}`, and improving the confidential client docstring to give some better explanation.

src/globus_cli/login_manager/tokenstore.py
sirosen: Yeah, simply wrapping this in `cached_property` poses no issues, as far as I can tell (one mock needed a tweak, that was all). I'm including it.

src/globus_cli/login_manager/tokenstore.py
sirosen: I think this comment is correct. `store()` being called means we're writing tokens to disk.

We are, as you say, overwriting any previous version marker (top of the comment calls this out as well). The comment indicates that this could "downgrade" tokens -- i.e. store tokens which are strictly weaker than the ones which were available on disk, which would presumably have a higher version number in most cases -- and it would not be recoverable information.

In fact, we've only used the contract version mechanism once, at the time we introduced it, so I wonder if it's a less general problem than we believed at the time, or if we've simply not been forced to reckon with scope changes in other services yet.

src/globus_cli/login_manager/tokenstore.py
sirosen: It worked just fine. I cannot recall why we did it that way. Maybe it was significant back when. :shrug:

src/globus_cli/login_manager/tokenstore.py
derek-globus: Do we need to evict the cache if someone calls `delete_templated_client`?

sirosen: I don't think it's necessary (although it would be more conceptually clean). It can only happen during logout when a user uses the (hidden) `--delete-client` flag, and the client is never used after that in that process.

Would you like me to add that? I believe it's just a matter of doing a `del self.cli_confidential_client` at the end of the delete.

derek-globus: If we're going to cache it, then yeah we should evict on delete.

It's true that in the current usage mode, a cache eviction will also terminate the process (command ends after logging out). But that's not a part of this object's contract and could easily result in a future bug if we forget that subtlety.

derek-globus (changes requested):
[This thread](https://github.com/globus/globus-cli/pull/1060#discussion_r1904456283) is my only outstanding sticking point.

src/globus_cli/login_manager/tokenstore.py
sirosen: I added the `del`, nothing broke in existing tests (I didn't expect it would, but hey! good to confirm!), and I've "locked it in" with a couple of new tests.

derek-globus (approved):

PR #1063, opened 2024-12-20
add access permission expiration support

aaschaer:
SC: https://app.shortcut.com/globus/story/30787/cli-support-showing-and-setting-expiration-date-field-for-acls

Main changes are the addition of `--expiration-date` to `globus endpoint permission create` and `globus endpoint permission update` and the `Expiration Date` output field for `globus endpoint permission show`

`--permissions` is also no longer a required option for `permission update` so users may only update expiration date.

I found a bug in the parsing of principal outputs while working on this caused by the `PrincipalDictFormatter` erroring on response objects being passed. Allowing it to accept `GlobusHTTPResponse` and extract their data seemed like the simplest fix.

Also includes fixtures and tests for `permission show` and `permission update` which were previously untested.

### DIFF
--- a/src/globus_cli/commands/endpoint/permission/update.py
+++ b/src/globus_cli/commands/endpoint/permission/update.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import typing as t
 import uuid
 
@@ -6,6 +8,8 @@ import click
 from globus_cli.login_manager import LoginManager
 from globus_cli.parsing import command, endpoint_id_arg
 from globus_cli.termio import display
+
+from ._common import expiration_date_option
 
 
 @command(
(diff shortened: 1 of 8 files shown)


src/globus_cli/commands/endpoint/permission/update.py
```diff
@@ -25,17 +29,18 @@
...
+@expiration_date_option
 @LoginManager.requires_login("transfer")
 def update_command(
     login_manager: LoginManager,
     *,
-    permissions: t.Literal["r", "rw"],
+    permissions: t.Literal["r", "rw"] | None,
     rule_id: str,
     endpoint_id: uuid.UUID,
+    expiration_date: str | None,
```
ada-globus, line 43: Is it worth ensuring that at least _one_ of {`permissions` | `expiration_date`} is set? Not sure what the behavior of the API is in this case; if it results in a nice enough error via the CLI, then maybe we leave as-is, but wanted to check since we're loosening the input expectations.

ada-globus (commented):
This all LGTM, but have one question re: the loosening of input for this command, which I couldn't trivially answer myself via the API docs.

src/globus_cli/commands/endpoint/permission/update.py
aaschaer: This is the response with neither option given. It seems clear enough to me, but also wouldn't be too hard to require one of the options if that would be better.

```
HTTP status:      400
request_id:       ugM0HuK7L
code:             BadRequest
message:          No fields were specified to update
```

ada-globus: I'm waffling (my preference to delegate as much as possible to the service is competing with my desire to give the most helpful error message to the user), but I think I'm okay with it as-is, knowing that we could easily revisit the topic later if we've got it wrong!

ada-globus (approved):

PR #1068, opened 2025-01-07
Fix mypy configuration in tox and CI, control python versions and update click-type-test usages

sirosen:
The overall goal of this PR is a cleanup to our `mypy` usage to get CI back into a healthy state.
We want to remove `base_python` specification (which is what's causing `tox_uv` to fail in surprising ways), and to achieve this a minor refactor is in order.

Updating to the latest `mypy` version and explicitly testing it on our lowest supported Python version (3.8) fails due to runtime usages which are never _in fact_ reached on lower Pythons, but which `mypy` can't be expected to recognize as such.
Refine these to pass on `py3.8`, then update so that CI is running `mypy --python-version 3.8` and `mypy --python-version 3.13`.

A very minor, unrelated bugfix to tox `depends` config is included.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -55,7 +55,8 @@ jobs:
             cpythons:
               - "3.13"
             tox-environments:
-              - "mypy"
+              - "mypy-py38"
+              - "mypy-py313"
               - "reference"
               - "twine-check"
             cache-key-prefix: "quality"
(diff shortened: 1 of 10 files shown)


sirosen:
I'm working on trying to solve the issues with `tuple` vs `typing.Tuple` comparisons in `click-type-test` before I invest in trying to fix them here. I'd like to support this smoothly, but we'll see how hard it is to do over there.

kurtmckee (approved):

kurtmckee (approved):

PR #1055, opened 2024-11-12
add support for entity type filtering on endpoint search

pjhinton-globus:
Add new command line option to globus endpoint search which limits the returned values to certain types of collections. Make use of change to Python SDK work in 37389.

refs sc-29242

PR #1071, opened 2025-01-11
Add initial tests for 'globus endpoint search'

sirosen:
This command has not had any functional test coverage, so these are the first tests for the command.

- Add an initial test for 'endpoint search'
- Add a test for --filter-entity-type

---

`--filter-entity-type` was added in #1055, but it was not possible to test it at the time because we had no fixture data to work off of.
To build initial data, I grabbed a live service search result and rewrote it thoroughly to remove my account and collection details.

@pjhinton-globus, the second commit here shows what I wanted you to be able to add.
When I realized that we'd need to setup new fixture data for endpoint search, I felt it was unfair to ask for this as part of an incremental addition to an existing, untested command.

I haven't discussed my decision to use `str.split()` to build out the data with anyone yet, which is probably the most contentious thing I've done here.

### DIFF
--- a/tests/functional/endpoint/test_endpoint_search.py
+++ b/tests/functional/endpoint/test_endpoint_search.py
@@ -5,7 +5,7 @@ import typing as t
 import uuid
 
 import pytest
-from globus_sdk._testing import RegisteredResponse
+from globus_sdk._testing import RegisteredResponse, get_last_request
 
 
 def _make_mapped_collection_search_result(
@@ -107,3 +107,33 @@ def test_search_shows_collection_id(run_line, singular_search_response):
     assert len(lines) == 3
     assert endpoint_id not in lines[-1]
     assert collection_id in lines[-1]
+
+
+@pytest.mark.parametrize(
+    "entity_type",
+    (
+        "GCP_mapped_collection",
+        "GCP_guest_collection",
+        "GCSv5_endpoint",
+        "GCSv5_mapped_collection",
+        "GCSv5_guest_collection",
+    ),
+)
+def test_search_can_send_entity_type_parameter(
+    run_line, singular_search_response, entity_type
+):
+    singular_search_response.add()
+    run_line(
+        [
+            "globus",
+            "endpoint",
+            "search",
+            "mytestquery",
+            "--filter-entity-type",
+            entity_type,
+        ]
+    )
+
+    # confirm that the entity type is sent in the query string
+    last_req = get_last_request()
+    assert entity_type in last_req.url
--- a/tests/functional/endpoint/test_endpoint_search.py
+++ b/tests/functional/endpoint/test_endpoint_search.py
@@ -1,0 +1,109 @@
+from __future__ import annotations
+
+import random
+import typing as t
+import uuid
+
+import pytest
+from globus_sdk._testing import RegisteredResponse
+
+
+def _make_mapped_collection_search_result(
+    collection_id: str,
+    endpoint_id: str,
+    *,
+    display_name: str = "dummy result",
+    authentication_timeout_mins: int = 15840,
+) -> dict[str, t.Any]:
+    username = "u_abcdefghijklmnop"  # not a real b32 username
+    owner_string = "globus@globus.org"
+    manager_fqdn = "a0bc1.23de.data.globus.org"
+    collection_fqdn = f"m-f45678.{manager_fqdn}"
+    gcs_version = "5.4.50"
+    data = {
+        "DATA_TYPE": "endpoint",
+        "_rank": random.random() * 10,
+        "authentication_timeout_mins": authentication_timeout_mins,
+        "canonical_name": f"u_{username}#{collection_id}",
+        "default_directory": "/{server_default}/",
+        "display_name": display_name,
+        "entity_type": "GCSv5_mapped_collection",
+        "expires_in": -1,
+        "gcs_manager_url": f"https://{manager_fqdn}",
+        "gcs_version": gcs_version,
+        "id": collection_id,
+        "location": "Automatic",
+        "max_concurrency": 4,
+        "max_parallelism": 8,
+        "my_effective_roles": [],
+        "myproxy_server": "myproxy.globusonline.org",
+        "name": collection_id,
+        "network_use": "normal",
+        "non_functional_endpoint_display_name": f"host of {display_name}",
+        "non_functional_endpoint_id": endpoint_id,
+        "owner_id": endpoint_id,
+        "owner_string": owner_string,
+        "preferred_concurrency": 2,
+        "preferred_parallelism": 4,
+        "public": True,
+        "shareable": True,
+        "tlsftp_server": f"tlsftp://{collection_fqdn}:443",
+        "username": username,
+    }
+    # there are so many null and False values in a typical item, it's worth packing them
+    # in strings and filling them in loops -- makes the document above more readable
+    null_keys = """
+        acl_max_expiration_period_mins authentication_assurance_timeout
+        authentication_policy_id contact_email contact_info department description
+        expire_time gcp_connected gcp_paused globus_connect_setup_key host_endpoint
+        host_endpoint_display_name host_endpoint_id host_path https_server info_link
+        keywords last_accessed_time local_user_info_available
+        mapped_collection_display_name mapped_collection_id myproxy_dn oauth_server
+        organization s3_url sharing_target_endpoint sharing_target_root_path
+        subscription_id user_message user_message_link
+    """.split()
+    false_keys = """
+        acl_available acl_editable activated disable_anonymous_writes disable_verify
+        force_encryption force_verify french_english_bilingual high_assurance in_use
+        is_globus_connect mfa_required non_functional requester_pays s3_owner_activated
+    """.split()
+    for k in null_keys:
+        data[k] = None
+    for k in false_keys:
+        data[k] = False
+    return data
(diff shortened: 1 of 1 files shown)


tests/functional/endpoint/test_endpoint_search.py
```diff
@@ -0,0 +1,139 @@
...
+        keywords last_accessed_time local_user_info_available
+        mapped_collection_display_name mapped_collection_id myproxy_dn oauth_server
+        organization s3_url sharing_target_endpoint sharing_target_root_path
+        subscription_id user_message user_message_link
+    """.split()
+    false_keys = """
+        acl_available acl_editable activated disable_anonymous_writes disable_verify
+        force_encryption force_verify french_english_bilingual high_assurance in_use
+        is_globus_connect mfa_required non_functional requester_pays s3_owner_activated
+    """.split()
```
kurtmckee, line 69: Oh, I see what you mean about this being contentious, but it's not `.split()` that catches my eye. Rather, this string block has to be manually flowed or re-flowed if additional null or false keys are identified/added.

I'd prefer that this just be a set of strings, and that two sets have an assertion that they don't intersect.

```python
null_keys = { ... }
false_keys = { ... }
assert null_keys & false_keys == set()
```

tests/functional/endpoint/test_endpoint_search.py
```diff
@@ -0,0 +1,139 @@
...
+from globus_sdk._testing import RegisteredResponse, get_last_request
+
+
+def _make_mapped_collection_search_result(
+    collection_id: str,
+    endpoint_id: str,
+    *,
+    display_name: str = "dummy result",
+    authentication_timeout_mins: int = 15840,
+) -> dict[str, t.Any]:
```
kurtmckee, line 17: Providing `display_name` and `authentication_timeout_mins` defaults here means that it's incumbent on fixtures to either know these values and fill them in as `metadata`, or to call this method with their own values and remember to fill in the values as `metadata`.

Below, I see that the single response fixture relied on the defaults and didn't fill them in as metadata. As this is the CLI, not the SDK, this is probably mostly fine. However, I want to point out that using default parameters here means that fixtures can fail to provide all of the metadata about the responses, and this is the silent, default outcome.

tests/functional/endpoint/test_endpoint_search.py
```diff
@@ -0,0 +1,139 @@
...
+    meta = singular_search_response.metadata
+    collection_id = meta["collection_id"]
+    endpoint_id = meta["endpoint_id"]
+
+    result = run_line("globus endpoint search mytestquery")
+
+    lines = result.output.rstrip("\n").split("\n")
+    assert len(lines) == 3
+    assert endpoint_id not in lines[-1]
+    assert collection_id in lines[-1]
```
kurtmckee, line 109: Why is the final line important to check here? I can't read this and understand the significance of these tests.

Should `endpoint_id` _never_ appear in the output when a collection ID is known? If so, these tests are written with an assumption about conditions elsewhere in the code, and I recommend asserting that `endpoint_id` doesn't appear anywhere in `result.output`.

tests/functional/endpoint/test_endpoint_search.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+@pytest.mark.parametrize(
+    "entity_type",
+    (
+        "GCP_mapped_collection",
+        "GCP_guest_collection",
+        "GCSv5_endpoint",
+        "GCSv5_mapped_collection",
+        "GCSv5_guest_collection",
```
kurtmckee, line 119: Do these values appear in an enum somewhere? If so, it will be worthwhile to use that enum directly (or perhaps to compose this list and then make assertions that it matches the enum).

tests/functional/endpoint/test_endpoint_search.py
```diff
@@ -0,0 +1,139 @@
...
+    singular_search_response.add()
+    run_line(
+        [
+            "globus",
+            "endpoint",
+            "search",
+            "mytestquery",
+            "--filter-entity-type",
+            entity_type,
+        ]
```
kurtmckee, line 134: Now _this_ is where I would prefer to see a `.split()` call! 😀 Like

```python
cmd = f"globus endpoint search mytestquery --filter-entity-type {entity_type}"
run_line(cmd.split())
```

tests/functional/endpoint/test_endpoint_search.py
```diff
@@ -0,0 +1,139 @@
...
+    )
+
+    # confirm that the entity type is sent in the query string
+    last_req = get_last_request()
+    assert entity_type in last_req.url
```
kurtmckee, line 139: Technically this isn't checking that the entity type is sent in the query string.

kurtmckee (changes requested):

tests/functional/endpoint/test_endpoint_search.py
sirosen: In the original, I felt that these fields were hurting readability in two ways:
- mixes "background data" with some of the essential, identifiable fields in the fixture
- makes the vertical scroll very long

I'll switch to sets. `black` will "help" here -- the sets won't be very compact -- but at least keeps the data separate. I'm okay with that; just sharing why I went with this approach as my initial pitch. :slightly_smiling_face:

sirosen: The moment I posted this, it occurred to me that this might be a rare time when `fmt: off` is tempting. I admit that I am tempted, but don't plan to use it. It loses the ability to add to the sets by slapping a string down, triggering your `black` integration, and walking away whistling.
I sometimes write single-quoted strings and let black fix it for me, and very commonly I'll produce long (e.g., 300 character) lines during refactors, and rely on `black` to take me under the flake8 limit.

tests/functional/endpoint/test_endpoint_search.py
sirosen: I don't really know how I want to phrase fixtures for the SDK to provide via `_testing`. These sorts of parameters are part of that fuzzy thinking.

I'm going to leave it as-is right now, as I view this helper as an implementation detail for this test module, but I hear what you're saying about the data being implicit where maybe it should be explicit.

tests/functional/endpoint/test_endpoint_search.py
sirosen: Let me see if we have a nice matcher somewhere in CLI fixtures for this. The output is table-formatted and therefore has (with one result in the fixture data) the following format:
```
header field 1 | header field 2
---------------|---------------
value 1        | value 2
```

I should really assert, I think:
- that there are two lines at the start
- the first one is a header with the following fields: ... (fill this in)
- the second one is a separator line
- the rest of the lines are "data"
  - there is only one "data line"
  - that "data line" contains the collection_id
- the endpoint_id does not appear in the output

I can obviously do all that "from scratch", but I think we have some tools we made for this.

tests/functional/endpoint/test_endpoint_search.py
sirosen: They do not appear in an enum (IMO, "thank goodness", str enums are so clunky...).

They're in a `Literal` type annotation.
Would you like me to write the "magic" to extract them from that annotation? I've done it before and it's pretty easy, but it adds a little bit of complexity to the test module.

tests/functional/endpoint/test_endpoint_search.py
sirosen: Interesting; I disagree.

`run_line(list | str)` lets us use whichever form fits the usage. In this case, we want to make sure that some variable, `entity_type`, is passed as the value for `--filter-entity-type`.

Consider this alternative way of writing what you suggest:
```python
run_line(
    f'globus endpoint search mytestquery --filter-entity-type "{entity_type}"'
)
```

Is it more correct? That depends on the `shlex` usage which happens in `run_line`, and whether or not `entity_type` can contain a space (which we can see above, it does not).

So in this case, no difference.

But I think the right _habit_ is to be passing variable arguments in the list form, which removes any question of `shlex.split` behavior from the tests.

tests/functional/endpoint/test_endpoint_search.py
sirosen: Yeah, I took a shortcut. Forgive me! :joy: 

I'll pull out the `urllib.parse` stuff and get you the thing which belongs here. :wink:

tests/functional/endpoint/test_endpoint_search.py
kurtmckee: Good point!

tests/functional/endpoint/test_endpoint_search.py
kurtmckee: Get the right tool!

Get the right tool for the job!

tests/functional/endpoint/test_endpoint_search.py
sirosen: Eh, now that I'm in the module I'm changing my mind. I didn't have a clear rationale at play here for selecting these fields in this way.
You'll see `authentication_timeout_mins` disappear (it's just a fixed field in the response), and `display_name` become required.

tests/functional/endpoint/test_endpoint_search.py
sirosen: We have tools for record-formatted output parsing, but not for tables (or at least, not in the places I looked). So I'll be doing this a bit manually, inline here for now.

PR #1074, opened 2025-01-13
Start using xdist in pytest opts

sirosen:
We have `pytest-xdist` already installed, but we aren't using it.

Set `-n auto --maxprocesses 8` in our config, so that it will be activated. The max prevents `-n auto` from starting large numbers of workers (e.g., 20) and losing any speed benefit to overheads. Using `-n auto` rather than, for example, `-n 4` ensures that in CI and other variable environments, we get the right number of workers.

One test with a randomized parameter needed a tweak, as it made an xdist parameter consistency check fail.

---

After CI runs for this setting, I'll be able to share overall numbers from that context.
In my local testing, this was a 4x speed boost.


kurtmckee (approved):

sirosen:
I have to convert this to a draft for the moment (should convert back later today).

I've run into some issues with local coverage and need to test more to proceed.

Notably, I'm also rebasing this onto #1070 to pick up the improved config there.

I've had to customize our `coverage` invocation a bit more (we have prior art for this in another project) to inject a sitecustomize the same way that `pytest-cov` does. That, in turn, _can_ make parallel tox runs (`tox p`) fail as there are too many parallel processes and some tests which start subprocess interpreters exceed the 3s pytest-timeout window. A bit of tuning and massaging and we should be good to go.

sirosen:
I finally got time to circle back on this. The blocker was that `tox run-parallel` was degraded significantly by this PR, as the parallelism from xdist was running on top of the `tox run-parallel` behavior. I got significantly worse performance and had to figure out how to fix.

It took a bit of work to figure out how to get information about the tox invocation into the pytest context, but once I got a version of that working (always open to refining it further), I was able to set xdist to "0 workers" (no worker overheads, basically disables xdist).

As a result, timing running `tox r` is reduced from ~135s to ~85s on my machine, and `tox p` is ~30s both on `main` and this branch.

PR #1076, opened 2025-01-27
Refactored termio display functions into Printers

derek-globus:
## Changes

* Refactored the display functions to use a new `Printer` class heirarchy
   * This started from a goal to add a field `ObjectFormatter` (for dependent scope displays). But in doing so, I was really struggling to conceptualize all of the formats that a serialized value could be rendered. So I refactored those into their own types to hopefully make testing, validation, and comprehension more transparent, explicit, and extensible.
   * As such, this is a large step but really just a first step to rethink how commands should be interacting with the `termio` module.

* Replaced `Field`'s `.__call__` function with a `.serialize` one.
* Removed `display.render_table`; this had one call site that could trivially be replaced with a call to `display()`


src/globus_cli/termio/printers/table_printer.py
```diff
@@ -0,0 +1,118 @@
...
+        cells = []
+        if values is None:
+            values = [""] * table.num_columns
+
+        for x, value in enumerate(values):
+            width = self._column_width(table, x)
+            cells.append(value.ljust(width, fillchar))
+        return " | ".join(cells)
+
+    @functools.lru_cache(maxsize=16)  # noqa: B019
```
sirosen, line 77: Why is this cache size limited? I would have expected that since there's only one `DataTable` built, we wouldn't have needed a limit.

Can we make the limit higher? (And what would the consequences of that be?)

src/globus_cli/termio/printers/table_printer.py
```diff
@@ -0,0 +1,118 @@
...
+
+class DataTable:
+    def __init__(self, fields: tuple[Field, ...], data: t.Iterable[t.Any]) -> None:
+        self._fields = fields
+        self._data = tuple(data)
+
+        self.num_columns = len(fields)
+        self.num_rows = len(self._data)
+
+    @functools.lru_cache(maxsize=256)  # noqa: B019
```
sirosen, line 95: In this case, I think I understand the rationale behind limiting this more, but because I can understand how we'd reach the limit, I have a stronger feeling that we should not be using a bounded size cache like this. Either we should get comfortable with the idea of it being unbounded (I think I am) or we should rework a little more.

Let's suppose there are 5 columns and 100 rows of data.
Because we'll walk all 100 rows in the first pass to get column widths, we'll go through the full cache size (256 < 500) and `serialize(...)` each item. And then we'll do it again to actually render the output.

We might have already been doing this in the existing implementation (probably -- it wasn't super optimized), but this is really a _lot_ of rerunning of code for data sizes which almost definitely fit easily in memory. And now that my eyes are on it I'd like us to address it.

Sticking with 5 columns (it's a pretty average number for our table outputs), I can imagine 1000 lines of output as an easy case which we should handle. 5000 lines of output seems like a reasonable guesstimate at an upper bound. 20000+ lines of output we can probably not worry about optimizing much (since getting column widths before printing is already a big issue at that size).

So if we have thousands of rows, times 5, then let's say we need `maxsize=25000`. How bad would that be?
If it is bad, do we need to convert and discard the original `Iterable`? I think yes.
So we'd have a pattern like...
```python
data = <some generator>
table = DataTable(fields)
for item in data:
    table.add_row(item)  # only stores the serialized values

# not that we'd necessarily explicitly delete, but the point
# is that Python is now free to discard the original
del data
```

(I'm not arguing for this interface, only for this paradigm in which the original data is ephemeral.)

---

Am I way off base here? I think there's a benefit to calling `serialize` only once per cell and that it's worth pursuing. Perhaps you've thought about this and decided against it for some reason?

tests/functional/groups/test_get_by_subscription.py
```diff
@@ -136,7 +136,7 @@ def test_group_get_by_subscription_text(run_line):
     )
 
 
-@pytest.mark.parametrize("output_format", ("text", "json", "unix"))
+@pytest.mark.parametrize("output_format", ("unix",))  # ("text", "json", "unix"))
```
sirosen, line 139: ?

changelog.d/20250127_151216_derek_printers.md
```diff
@@ -0,0 +1,7 @@
+
+### Enhancements
+
+*   Added `globus_cli.termio.printers` module with an explicit interface for different
+    printing formats.
+
+     *   Refactored `globus_cli.termio.display` to use this framework
```
sirosen, line 7: I would like us to not put in this changelog entry.
Remember that the CLI changelog is user-facing.

sirosen (commented):
My main area of attention in this is the table format and the fixed-size caches being used.

Overall I think it's really telling that almost no code in `src/globus_cli/commands/` changed! :100:

src/globus_cli/termio/__init__.py
```diff
@@ -12,9 +12,9 @@
     term_is_interactive,
     verbosity,
 )
+from .display import display
```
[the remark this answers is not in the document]
derek-globus: Yeah that's fine. I never know how to deal with this type of nested duplicate naming scheme

src/globus_cli/termio/printers/base.py
```diff
@@ -0,0 +1,34 @@
...
+
+from ..context import get_jmespath_expression
+
+DataType = t.TypeVar("DataType")
+
+
+class Printer(abc.ABC, t.Generic[DataType]):
+
+    @abc.abstractmethod
+    def print(self, data: DataType, stream: t.IO[str] | None = None) -> None:
```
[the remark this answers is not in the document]
derek-globus: What about echo?

src/globus_cli/termio/printers/table_printer.py
derek-globus: I mostly think of it as a question of compute time vs memory prioritization.

I didn't put too too much thought into it so Im happy to bump it up. The CLI executes in a context where it shouldn't need to be concerned with large enough amounts of data to be too concerned with memory constraints so I don't mind prioritizing compute time.

What do you think are better cache bounds than 16 and 256? 32 and 1024?

changelog.d/20250127_151216_derek_printers.md
derek-globus: Is there a better category to classify it? Or do we simply not document development-only changes in our changelog?

src/globus_cli/termio/printers/table_printer.py
sirosen: I agree, the CLI context should never realistically hit memory limits. And if it does, our cache sizing is unlikely to be the main driver of the issue.

I would actually specify `maxsize=0` for this one, on the grounds that "the number of values should always be so small that setting a limit cannot have any meaningful impact". Since we cannot find a meaningful value to set, we shouldn't set one.

I'm strongly tempted to suggest the same for the other limit as well -- taking the attitude that the CLI will systematically prioritize compute over memory until we're shown that there's a problem doing so. I'm starting to look more closely and I don't think even 1 million values would make the cache unreasonably large.
Here's some sizing info:
```python
>>> sys.getsizeof(1)  # an interned integer, 28 bytes
28
>>> sys.getsizeof("hello sized world")  # some string, 58 bytes
58
>>> sys.getsizeof("")  # base of 41 bytes for the empty string
41
>>> sys.getsizeof("a")  # plus a byte per char
42
>>> sys.getsizeof(str(uuid.uuid1()))  # a longer string, 77 bytes
77
>>> sys.getsizeof(str(uuid.uuid1())) - len(str(uuid.uuid1()))  # minus length gets back to the base 41
41
```

Meaning that there's an overhead of 41 bytes per string, plus the actual length of the string.
Suppose the cache is populated with 1 million 36 byte (UUID-length) strings.
10^6 * 77 bytes = 73.4MB

The cache itself -- the datastructure and the keys -- consume another chunk of memory. If we conservatively triple that 73.4MB, we get 210MB used.

My current system has many long-running processes which happily chug along using more memory than this, and I think it's fine. (Although in principle I'd like it if they used less.)

So all of that is to show my work as I reason through to a point of "do we have a meaningful limit we should set on this?" I would like us to start limiting our memory usage only in the most extreme cases. Perhaps an initial limit of 8192 or 100,000 to indicate "only in absurd cases would we expect the maxsize to be reached".

changelog.d/20250127_151216_derek_printers.md
sirosen: Definitely we don't for _this_ changelog. More generally, I do not believe that documenting all development-specific changes in a changelog is a good practice and I think some of our projects are _way_ too strongly biased towards documenting every internal change.

Changelogs compress and **lossily** encode the information in a diff (between releases) for some specific audience to consume. In the case of the CLI, we have a very particular audience in mind (the end users) and our current documentation integration requires that our changelog be the verbatim content for that audience. We can improve or change the toolchain to allow us to target multiple audiences (e.g., add a `Maintenance` section to the changelog which gets stripped automatically for the doc site render), but of course that's a body of work we'd have to put together.

If we had a toolchain here which produced a separate changelog for maintainers and contributors vs users, I would definitely be in favor of documenting this one (because it's a significant change which won't necessarily trace back to a single obvious commit). But I think the realistic workflow for a developer trying to understand project history is to use tools like `git shortlog` and blame, so I'm not all that convinced that a "developers and contributors" changelog is a resource worth intentionally producing.

The whole "check for a changelog workflow" was something I added to make sure we always were making an intentional decision about whether or not to document a change -- it was never meant to push us to document everything and I very, very frequently use the escape-hatch label.

src/globus_cli/termio/printers/table_printer.py
derek-globus: Removed the cache limit.

src/globus_cli/termio/printers/table_printer.py
derek-globus: Rewrote `DataTable` to preload cells as you suggested.

sirosen (changes requested):
Everything in the src tree looks good. I'd really like `textwrap.dedent` in the tests because I think it makes the tests more readable (no explicit newline chars, less quoting).

We're very close to having this merged, I think.

tests/unit/termio/printer/test_json_printer.py
```diff
@@ -0,0 +1,50 @@
...
+
+    # fmt: off
+    assert printed_json == (
+        "{\n"
+        '  "a": 2,\n'
+        '  "b": 1,\n'
+        '  "c": 3\n'
+        "}\n"
+    )
+    # fmt: on
```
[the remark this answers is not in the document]
derek-globus: I started out using dedent but it doesn't work everywhere; specifically with newline placement on a string.

The exact output for test_table_printer_handles_missing_fields for instance is
```
"A    | B   \n"
"---- | ----\n"
"1    | None\n"
"None | 5   \n"
```

If we try to dedent wrap that, it becomes
```
"A    | B\n"
"---- | ----\n"
"1    | None\n"
"None | 5\n"
```

---

Given that newline placement location can matter; I opted to keep the string representations consistent across the board; although I do agree that I'd prefer to use multi-line strings rather than disabling black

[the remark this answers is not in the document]
sirosen: Oh, I see what you were fighting against -- trailing whitespace.

Okay, I don't think this is as clear and obvious of an improvement as I thought. Let's leave it the way you did it for now, and if I can cook up something clever later, I might put up a proposal in a PR.

sirosen (approved):

tests/unit/termio/printer/test_json_printer.py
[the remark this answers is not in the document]
sirosen: Er, to be clear: I understand that it's not an improvement because it doesn't _work_. But the things we might do to make it work (like stripping trailing whitespace before comparison) seem _worse_. :sweat_smile:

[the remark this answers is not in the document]
derek-globus: > I might put up a proposal in a PR

I'd encourage it; I believe there's clearer ways to lay these expected outputs out I just couldn't figure it out myself & didn't want to block too long on it.

---

Since we have a thread though I do want to stipulate that if we change it, we need to maintain the legibility above all else. Our tests should be utterly transparent about what the desired output is, no rendering newlines in a developer's head.

(I imagine we agree on this point already)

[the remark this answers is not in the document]
sirosen: We not only agree; it is why I wanted us to use multiline strings (not realizing the problem it causes)!

Right now my best idea is some kind of `assert_lines_match()` helper which could be used
```python
assert_lines_match(
    output,
    [
        "A    | B   ",
        "-----|-----",
        ...,
    ]
)
```

There's some pytest magic you can do so that assertions from a helper are transparently lifted into the calling context, but I can never remember it.
This might be the way I try.

PR #1077, opened 2025-01-31
Added globus auth scope show command

derek-globus:
## Added
 * New command `globus auth scope show SCOPE_ID_OR_STRING`
 * New FieldFormatters:
    * `ArrayMultilineFormatter(ArrayFormatter)` - Supports multiline formatted elements, adding a yaml-style `-` element start indicator to the beginning of each new one.
    * `RecordFormatter` - Prints an object using the `RecordPrinter` into a buffered string.

## Draft
* This PR is in draft until:
   1. I have a chance to put the command in front of the auth team.
   2. I figure out how to figure out reserved_width in RecordFormatter; it's a pretty minute thing but is hardcoded right now.

#### By Scope ID

Standard Output *(dependent scope strings get loaded in a secondary batched call)*
```
> globus auth scope show 24f3dcbe-7655-4721-bc64-d1c5d635b9a1
Scope String:          https://auth.globus.org/scopes/actions.globus.org/hello_world
Scope ID:              24f3dcbe-7655-4721-bc64-d1c5d635b9a1
Name:                  Hello World Action
Description:           Allow the Hello World action to extend greetings.
Client ID:             5fac2e64-c734-4e6b-90ea-ff12ddbf9653
Allows Refresh Tokens: True
Required Domains:      
Advertised:            True
Dependent Scopes:      
  - Scope String:           urn:globus:auth:scope:nexus.api.globus.org:groups
    Scope ID:               69a73d8f-cd45-4e37-bb3b-43678424aeb7
    Optional:               False
    Requires Refresh Token: False
  - Scope String:           urn:globus:auth:scope:groups.api.globus.org:view_my_groups_and_memberships
    Scope ID:               73320ffe-4cb4-4b25-a0a3-83d53d59ce4f
    Optional:               False
    Requires Refresh Token: False
```

JSON Output *(no secondary dependent scope resolution call happens)*
```
> globus auth scope show  24f3dcbe-7655-4721-bc64-d1c5d635b9a1 -F json                
{
  "scope": {
    "advertised": true,
    "allows_refresh_token": true,
    "client": "5fac2e64-c734-4e6b-90ea-ff12ddbf9653",
    "dependent_scopes": [
      {
        "optional": false,
        "requires_refresh_token": false,
        "scope": "69a73d8f-cd45-4e37-bb3b-43678424aeb7"
      },
      {
        "optional": false,
        "requires_refresh_token": false,
        "scope": "73320ffe-4cb4-4b25-a0a3-83d53d59ce4f"
      }
    ],
    "description": "Allow the Hello World action to extend greetings.",
    "id": "24f3dcbe-7655-4721-bc64-d1c5d635b9a1",
    "name": "Hello World Action",
    "required_domains": [],
    "scope_string": "https://auth.globus.org/scopes/actions.globus.org/hello_world"
  }
}
```

#### By Scope String

Similarly when accessing by scope string

```
> globus auth scope show https://auth.globus.org/scopes/actions.globus.org/hello_world
Scope String:          https://auth.globus.org/scopes/actions.globus.org/hello_world
Scope ID:              24f3dcbe-7655-4721-bc64-d1c5d635b9a1
Name:                  Hello World Action
Description:           Allow the Hello World action to extend greetings.
Client ID:             5fac2e64-c734-4e6b-90ea-ff12ddbf9653
Allows Refresh Tokens: True
Required Domains:      
Advertised:            True
Dependent Scopes:      
  - Scope String:           urn:globus:auth:scope:nexus.api.globus.org:groups
    Scope ID:               69a73d8f-cd45-4e37-bb3b-43678424aeb7
    Optional:               False
    Requires Refresh Token: False
  - Scope String:           urn:globus:auth:scope:groups.api.globus.org:view_my_groups_and_memberships
    Scope ID:               73320ffe-4cb4-4b25-a0a3-83d53d59ce4f
    Optional:               False
    Requires Refresh Token: False
```

JSON by scope string uses a the get-scopes route so response is a list of one element
```
> globus auth scope show https://auth.globus.org/scopes/actions.globus.org/hello_world -F json
{
  "scopes": [
    {
      "advertised": true,
      "allows_refresh_token": true,
      "client": "5fac2e64-c734-4e6b-90ea-ff12ddbf9653",
      "dependent_scopes": [
        {
          "optional": false,
          "requires_refresh_token": false,
          "scope": "69a73d8f-cd45-4e37-bb3b-43678424aeb7"
        },
        {
          "optional": false,
          "requires_refresh_token": false,
          "scope": "73320ffe-4cb4-4b25-a0a3-83d53d59ce4f"
        }
      ],
      "description": "Allow the Hello World action to extend greetings.",
      "id": "24f3dcbe-7655-4721-bc64-d1c5d635b9a1",
      "name": "Hello World Action",
      "required_domains": [],
      "scope_string": "https://auth.globus.org/scopes/actions.globus.org/hello_world"
    }
  ]
}
```


kurtmckee:
> globus auth is a hidden command tree; Stephen and I discussed keeping it that way until we have 5+ commands registered under there

Why?

sirosen:
> > globus auth is a hidden command tree; Stephen and I discussed keeping it that way until we have 5+ commands registered under there
> 
> Why?

Because the launch of a `globus auth` command tree should potentially include aliasing/renaming and deprecation of existing commands like `globus get-identities`. We agreed on hidden commands as an easy way for us to continue to discuss exactly what should happen to the CLI when we add `globus auth` commands without blocking work, and that once there are a handful of commands implemented, the downside of holding back a release outweighs the potential upsides of making more adjustments for the initial release of these commands.

We can reopen this debate, but let's not do it in this PR thread.

src/globus_cli/termio/terminal_info.py
```diff
@@ -0,0 +1,40 @@
...
+        yield
+        self._column_delta += size
+
+    @property
+    def columns(self) -> int:
+        computed_columns = self._base_columns + self._column_delta
+        return max(self.MIN_COLUMNS, computed_columns)
+
+
+TERM_INFO = VirtualTerminalInfo()
```
sirosen, line 40: Can these refinements, to use `VirtualTerminalInfo` to capture this functionality and use it in `_item_wrapper` for fields, be separated into a distinct PR?

I think these are good and separable improvements. Having them isolated will make review much easier.

src/globus_cli/types.py
```diff
@@ -35,9 +35,28 @@
...
+def is_json_dict(value: t.Any) -> TypeGuard[JsonDict]:
+    if not isinstance(value, dict):
+        return False
+    return all(isinstance(k, str) and is_json_value(v) for k, v in value.items())
+
+
+def is_json_list(value: t.Any) -> TypeGuard[JsonList]:
+    if not isinstance(value, list):
+        return False
+    return all(is_json_value(item) for item in value)
```
sirosen, line 59: I'm a bit uncomfortable with these being mutually recursive helpers which can crawl a large object[^1].

I see where this is being used, to check an API response to ensure it is well-formed.
Would you be willing to give these functions a depth limit to their checks, which they can pass around recursively?

i.e.
```python
def is_json_dict(value: t.Any, depth: int = 2) -> TypeGuard[JsonDict]:
    if not isinstance(value, dict):
        return False
    if depth <= 0:
        return True
    return all(is_json_value(item, depth=depth - 1) for item in value)
```

That way, we'll never infinitely recurse and we won't see the CLI spin-lock if an API feeds back unexpectedly large nested data. (I'm imagining other unplanned usage in the future, like on an `input_schema` or `flow_definition` from a Flow.)

[^1]: or infinitely recursive, if fed malicious data... but I don't think we need to worry about that one

src/globus_cli/utils.py
```diff
@@ -284,9 +284,68 @@ def resolve_principal_urn(
...
         return True
     except ValueError:
         return False
+
+
+K = t.TypeVar("K")
+V = t.TypeVar("V")
+
+
+class LazyDict(t.Dict[K, V]):
```
sirosen, line 299: One of the things I'd like is for us to have a tighter signature on `__init__` (so that we can describe valid vs invalid arguments). I think the easiest path to that is to _contain_ a dict rather than _inheriting from_ a dict.

This is not a complete change suggestion, but what about using `typing.Mapping`?

```suggestion
class LazyDict(t.Mapping[K, V]):
```

`__init__` and other elements would also need refinement.

derek-globus: I'm inheriting directly from dict rather than mapping because of [json encoding](https://docs.python.org/3/library/json.html#encoders-and-decoders). Unless this thing is a dict, the builtin `json` module can't serialize it without a custom encoder.


```
Traceback (most recent call last):
  File "/Applications/PyCharm.app/Contents/plugins/python-ce/helpers/pydev/pydevconsole.py", line 364, in runcode
    coro = func()
  File "<input>", line 1, in <module>
  File "/Users/derekschlabach/.pyenv/versions/3.10.13/lib/python3.10/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/Users/derekschlabach/.pyenv/versions/3.10.13/lib/python3.10/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/Users/derekschlabach/.pyenv/versions/3.10.13/lib/python3.10/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/Users/derekschlabach/.pyenv/versions/3.10.13/lib/python3.10/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type MyMapping is not JSON serializable
```

This object needs to be serializable by our JsonPrinter for the `-F json` flag & I didn't want to add a custom serializer just for it.

sirosen: Looking at this again, I think we should be using `collections.abc.Mapping`.
The typing generic forms are not really where we want to inherit from for runtime behaviors, and as of 3.9+ the collections classes support class getitem.

derek-globus: Try telling that to py38. I had `dict[K, V]` initially but had to [change it](https://github.com/globus/globus-cli/pull/1077/commits/9c8e230f5a693609e7a2d8e47c6d79f3df6100e1) even with `__future__.annotations`

src/globus_cli/types.py
derek-globus: I don't mind adding depth limits but it'd have to be be quite large; not 2.

We're not doing anything compute intensive on these objects, just checking their mro. Similar to where we landed in our discussion about caching table cells, I want to err on the side of not being overly concerned w/ memory & compute usage constraints until we see a case or two that hits it.

How about a recursion limit of 100? That's 1/10th of my system default's 1000.

src/globus_cli/termio/terminal_info.py
derek-globus: Help me understand the scope you're envisioning?

I bundled them all toegether because:
1. The `VirtualTerminalInfo` refactor is justified by a need to be able to indent your terminal.
1. The ability to indent makes sense when you need nested multiline string formatters
1. Multiline string formatters make sense when you add a command that needs nested multiline string formatting
1. That command is `globus auth scope show`.

I don't see anything besides item (4) that justifies itself or establishes why we would shape a thing a certain way. But maybe I'm missing it?

src/globus_cli/utils.py
sirosen: This isn't critical, but inheriting from a `typing` definition feels wrong to me, and I'd like to be in a position where we never have to think about it.
Maybe we just punt on the issue.

More thoughts below:

Python 3.8 is out of support. We're still testing against it, but I think we should fix that so that we can write ergonomic code.

I still think the right _runtime_ type to inherit is `dict`. The 3.8-compatible way would be
```python
if t.TYPE_CHECKING:
    LazyDictBase = t.Dict[K, V]
else:
    LazyDictBase = dict
```

But that's pretty unpleasant to handle.

derek-globus: I basically agree with everything you've said here (including that being a pretty unpleasant to handle definition).

What if I added a comment to change the inherited class to `dict` once py38 was deprecated?

src/globus_cli/termio/terminal_info.py
sirosen: I'm looking at this from the perspective of the PR reviewer, and finding that the semantic change -- addition of a new command -- is harder to review because the PR itself includes a body of refactoring work.

If I can focus on reviewing the refactor itself, I can provide better feedback.
It may be driven by this feature, but my opinion is that the work can be trivially split off and that it therefore should be.

sirosen: We discussed and agreed:
1. This really isn't important.
2. A rebase of this feature PR is enough work to make us decide against it.

Agreement reached! :tada:

src/globus_cli/termio/printers/record_printer.py
```diff
@@ -88,6 +97,19 @@ def _key_len(self) -> int:
...
+        """
+        Access the printers TextWrapper, modifying the width if necessary.
+
+        :returns: a TextWrapper instance for wrapping item values.
+        """
+        # If the class was instantiated with an explicit max_content_width, don't
+        # override it.
+        if not self._is_max_content_width_explicit:
+            self._base_item_wrapper.width = TERM_INFO.columns
+        return self._base_item_wrapper
```
sirosen, line 111: I think there's a better path here where `TextWrapper` itself internalizes this setting.

1. Give `TextWrapper` an internal attribute e.g., `TextWrapper._width: int | None = None`
2. Make `TextWrapper.width` a property which sets `_width` to `TERM_INFO.columns` if it is not set
3. Given `TextWrapper.width` a setter which explicitly sets `_width`
4. Get rid of `_item_wrapper` as a property here -- go back to it being an attribute
5. In `__init__`, get rid of `_is_max_content_width_explicit` and instead do the (currently implemented) conditional setting of `width`

I think that solves this and has a less leaky abstraction.

src/globus_cli/termio/formatters/compound.py
```diff
@@ -38,17 +60,94 @@ def __init__(
...
+        in:  ["a", "b", "c\nd"]
+        out: "  - a\n  - b\n  - c\n    d"
+    """
+
+    def __init__(self, formatter: FieldFormatter[t.Any] | None = None) -> None:
+        super().__init__(delimiter=os.linesep, sort=False, element_formatter=formatter)
+
+    def render(self, value: list[str]) -> str:
+        prefix = os.linesep if value else ""
+        return prefix + super().render(value)
```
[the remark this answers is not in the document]
derek-globus: > Can you explain why the prefix needs to handle the empty array case specially?
If we don't, we'd add an extra newline.
**Without handling**
```
MyCoolId:               754eb9a9-ec6a-4a62-9a86-fca353caf307
My Cool Subobjects: 

My Cool Regular String: hello there
```

**With handling**
```
MyCoolId:               754eb9a9-ec6a-4a62-9a86-fca353caf307
My Cool Subobjects: 
My Cool Regular String: hello there
```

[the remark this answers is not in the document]
derek-globus: > I'm sort of of the opinion that if we're going to think about how to handle [] as an input here, we should actually skip the call to super().render and instead we should output "[]" (no newline decoration)

I agree with the result, but disagree the implementation.
We should implement empty array rendering in the parent class (`ArrayFormatter`), not the specialized multiline version.

In an earlier version of the code I had a parameter on `ArrayFormatter` called `"empty_render"` where you could configure which specific string was output when an empty list was supplied. In that implementation it used `"-"` instead of `"[]"` but the idea is the same: print something to signal "nothing here!" instead of leaving it blank. I pulled it from the implementation cause it felt like unnecessary bloat that we could add back in when there wasn't quite so much happening in the PR.

src/globus_cli/termio/formatters/compound.py
```diff
@@ -38,17 +60,94 @@ def __init__(
...
+
+    @classmethod
+    def _indent_element(cls, value: str) -> str:
+        """Indent a multi-line formatted element string."""
+        first_ = "  - "
+        if not value:
+            return first_
+        indent = "    "
+        indented = textwrap.indent(value, indent, lambda _: True)
+        return first_ + indented[len(indent) :]
```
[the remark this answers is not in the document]
derek-globus: > I found this function very hard to read.

I found it very hard to write! Open to improving.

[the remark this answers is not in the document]
derek-globus: > It's not exactly clear why "" is a special case for value. Recommended rephrase below.

```python
>>> textwrap.indent("", "   ", lambda _: True)
''
>>> textwrap.indent(" ", "   ", lambda _: True)
'    '
```

`textwrap.indent` won't indent an empty line. We're relying on there being enough characters in the string that we can replace the first 4 with the `first_` prefix string.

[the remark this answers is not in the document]
derek-globus: > predicate is a rarely used argument to indent and I don't think we need it.

The point you identify at the end of your comment (ensuring empty lines are indented) is why I'm using predicate there. From an implementation perspective I can add a comment to make that more clear but sounds like our guts disagree on the output as well: I do think we should indent empty strings.

1. It mirrors what we do with the RecordPrinter (extending the key's line up until the value, even if the value is an empty string).
2. It performs an easier to imagine string transformation (to me at least): we're inserting a rectangle of characters into the left side of your string. If subsequent steps want to reformat to remove erroneous spaces that's fine; but it shouldn't be the responsibility of this method. It's complicated internally enough that I want the interface to be super understandable.

src/globus_cli/termio/formatters/compound.py
```diff
@@ -38,17 +60,94 @@ def __init__(
...
+        Additional support for other printers/nested records should be added as needed.
+
+    Example:
+        in:  {"a": 1, "bb": 2}
+        out: "a:  1\nbb: 2"
+    """
+
+    def __init__(self, fields: t.Iterable[Field]) -> None:
+        self._fields = fields
+        self._printer = RecordPrinter(fields)
```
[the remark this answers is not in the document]
derek-globus: Oh that's an interesting idea. Maybe; I hadn't considered it. Open to seeing if it improves the state of play.

src/globus_cli/termio/formatters/compound.py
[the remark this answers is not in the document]
sirosen: We do disagree. I think we'll want to discuss to resolve.
I think `"[]"` is an appropriate rendering _specifically_ for the multiline formatter which imitates YAML formatting.
The reason is that in YAML, the following transformations to JSON hold:
```
# yaml
foo:
  - bar
# json
{"foo": ["bar"]}

# yaml
foo: []
# json
{"foo": []}

# yaml
foo:

# json
{"foo": null}
```

It is specifically the decision to use a YAML-like syntax that introduces an ambiguity here about the empty list. There is not the same ambiguity with comma-separated strings degenerating to an empty string.

In any case, I would find the paradigm easier to reason about as
```python
if value:
    return os.linesep + super().render(value)
else:
    return <whatever we like>
```

src/globus_cli/termio/formatters/compound.py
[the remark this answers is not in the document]
sirosen: Sorry, I was unclear. I followed (after reading 3 times) why `value == ""` needed special handling.
I think it is easier to understand as a reader if we get rid of `first_` and just write
```python
# <explanatory comment, per suggestion above>
if not value:
    return "  -"  # or some other string here
```

---

Regarding text wrapping, I think use of `lambda _: True` is sufficiently cryptic that it demands an explanatory comment.

Personally, I don't find that mild consistency with record formatting is compelling. I would rather treat all trailing whitespace as an artifact of implementation with no guaranteed semantics. That leaves us free to not include it here and to potentially strip it from the record and table printers in the future.

However, I don't feel strongly enough about this to make it a blocking concern. What I feel strongly about is:
If the predicate is being specified, a comment is needed to not only explain why it is being set but actually to remind the reader that `textwrap.indent` has an optional third argument. It's _so_ rare that I don't think it's safe to assume that people will know what it is, and I think the official docs are pretty poor for this one.

src/globus_cli/termio/formatters/compound.py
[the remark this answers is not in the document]
sirosen: I'll leave it up to you whether or not you want to try applying it here. If you prefer us to do it later, please just mark this thread resolved. :slightly_smiling_face:

src/globus_cli/termio/printers/record_printer.py
derek-globus: Applied; I put it in `terminal_info` so that it's a more general purpose utility but it works the way you described. Nice idea.

src/globus_cli/termio/formatters/compound.py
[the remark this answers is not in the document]
derek-globus: Discussed offline; I won't do this as a part of this body of work.

But may do it (or stephen may do it) in a later one.

src/globus_cli/termio/formatters/compound.py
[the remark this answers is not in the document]
derek-globus: Renamed, refactored, and added more explicit examples in the docstrings.

Lemme know if you're happy to resolve this thread.

src/globus_cli/termio/formatters/compound.py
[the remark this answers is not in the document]
derek-globus: Discussed offline;

* Updated the execution path to make the control flow process more explicit.
* Updated the array render to output "[]" for empty arrays instead of ""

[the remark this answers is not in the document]
sirosen: Yeah, I think the refactor looks great!

I now have the bias of someone who has looked closely at these codepaths and knows what the context is, but I think this is as straightforward a presentation as we can get.

src/globus_cli/termio/formatters/compound.py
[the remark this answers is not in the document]
sirosen: The example in the docstring helps, and I think the comment is sufficient. Overall, I'm very happy with the adjustments you made!

src/globus_cli/termio/terminal_info.py
```diff
@@ -0,0 +1,64 @@
...
+from textwrap import TextWrapper
+
+__all__ = ("TERM_INFO", "TerminalTextWrapper")
+
+
+class VirtualTerminalInfo:
+    MIN_COLUMNS = 6
+
+    def __init__(self) -> None:
+        self._column_delta = 0
```
sirosen, line 22: May I suggest that we name this to align with `indented()` as the indent? Also requires that we invert some of the arithmetic around it.

```suggestion
        self._current_indent = 0
```

I suggest this after a moment of confusion reading the `+=`, `-=`, etc during review.

tests/functional/auth/scope/test_show_scope.py
```diff
@@ -0,0 +1,57 @@
...
+    "scope_id_or_string",
+    (
+        "24f3dcbe-7655-4721-bc64-d1c5d635b9a1",
+        "https://auth.globus.org/scopes/actions.globus.org/hello_world",
+    ),
+)
+def test_show_scope(run_line, scope_id_or_string):
+    load_response_set("cli.scopes")
+
+    result = run_line(f"globus auth scope show {scope_id_or_string}")
```
sirosen, line 20: I notice that we have pytest params matched to the data here. I think it's fine but if you wanted to break the dependency up a bit, you can
```python
@pytest.mark.parametrize("lookup_method", ("id", "scope_string"))
def test_...:
    rset = load_response_set("cli.scopes")
    if lookup_method == "id":
        scope_arg = rset.metadata["hello_world_id"]
    elif lookup_method == "scope_string":
        scope_arg = rset.metadata["hello_world_string"]
    else:
        raise NotImplementedError(lookup_method)
```

It's totally optional, but if you'd like to do so, there's a path to doing this and potentially something similar to the expected output formulation.

sirosen (commented):
I only have a couple of minor comments I'd like us to evaluate before approval. I did a reread of the parts I had already reviewed and feel pretty good about the state of this.

tests/functional/auth/scope/test_show_scope.py
kurtmckee: It doesn't look like a conditional is required.

```python
@pytest.mark.parametrize("metadata_key", ("hello_world_id", "hello_world_string"))
def test_...:
    rset = load_response_set("cli.scopes")
    scope_arg = rset.metadata[metadata_key]
```

kurtmckee (approved):

src/globus_cli/utils.py
```diff
@@ -284,9 +284,70 @@ def resolve_principal_urn(
...
+
+
+# Note: This class extends `t.Dict` instead of `dict` for python3.8 compatibility.
+# Once we drop support for python3.8, we can switch to `dict[K, V]`.
+class LazyDict(t.Dict[K, V]):
+    """
+    A dictionary with support for lazily loaded values.
+    Lazy loaders are registered using the `register_loader` method.
+
+    Note: This class extends `dict`, not `UserDict` for json serializability.
```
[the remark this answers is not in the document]
sirosen: Probably the comment reflects some of my influence via code I've left around. :grin:
We have it in the SDK as well, as the base for the payload classes!

I think `UserDict` has some differences vs subclassing `dict` regarding the consistency of various mutators. It's all very niche stuff and I don't think there's much call for `UserDict` these days (and less for `UserList` and less _than that_ for `UserString`!).

I also think it's incidentally _nice_ that `self.data` is available on `UserDict`, since it means you can implement `__getitem__` in terms of `self.data.__getitem__` and have a clearer flow of control. So there is at least some rationale for still liking it!

But, of course, I'm not particularly attached to one implementation vs another for this sort of thing. :slightly_smiling_face:

PR #1085, opened 2025-03-04
Improve error messaging on bad client creds

sirosen:
When the CLI is used with client creds and the creds themselves are
invalid, Auth replies with an error, and the prior enforcement logic
treated this as a case to prompt for login.

Now, we check if we know that the client creds env vars are set and
validate and message about those vars based on what's observed.

Additionally, the non-client-creds case is reworded slightly to match
the way that `globus session` prompts are provided to users.

### DIFF
--- a/src/globus_cli/exception_handling/hooks.py
+++ b/src/globus_cli/exception_handling/hooks.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import json
+import os
 import textwrap
 import typing as t
 
@@ -9,7 +10,7 @@ import globus_sdk
 import globus_sdk.gare
 
 from globus_cli.endpointish import WrongEntityTypeError
-from globus_cli.login_manager import MissingLoginError
+from globus_cli.login_manager import MissingLoginError, is_client_login
 from globus_cli.termio import PrintableErrorField, outformat_is_json, write_error_info
 from globus_cli.types import JsonValue
 from globus_cli.utils import CLIAuthRequirementsError
@@ -244,13 +245,7 @@ def _concrete_consent_required_hook(
 def authentication_hook(
     exception: globus_sdk.TransferAPIError | globus_sdk.AuthAPIError,
 ) -> None:
-    click.echo(
-        (
-            "Globus CLI Error: No Authentication provided. Make sure "
-            "you have logged in with 'globus login'."
-        ),
-        err=True,
-    )
+    _concrete_unauthorized_hook()
 
 
 @sdk_error_handler(error_class="TransferAPIError")
@@ -414,13 +409,7 @@ def flows_error_hook(exception: globus_sdk.FlowsAPIError) -> None:
     condition=lambda err: err.message == "invalid_grant",
 )
 def invalidrefresh_hook(exception: globus_sdk.AuthAPIError) -> None:
-    click.echo(
-        (
-            "Globus CLI Error: Your credentials are no longer "
-            "valid. Please log in again with 'globus login'."
-        ),
-        err=True,
-    )
+    _concrete_unauthorized_hook()
 
 
 @sdk_error_handler(error_class="AuthAPIError")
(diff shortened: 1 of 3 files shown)


src/globus_cli/exception_handling/hooks.py
```diff
@@ -515,6 +504,80 @@ def _handle_gare(gare: globus_sdk.gare.GARE, message: str | None = None) -> None
         )
 
 
+def _concrete_unauthorized_hook() -> None:
```
derek-globus, line 507: What do you think about naming it as what it does rather than the hook it's implemented.

```suggestion
def _emit_unauthorized_message() -> None:
```

The hooks I assume are there with conventional names so that other locations can override them. But This is internal & won't ever be overridden so no utility in structuring it like that.

src/globus_cli/exception_handling/hooks.py
```diff
@@ -515,6 +504,80 @@ def _handle_gare(gare: globus_sdk.gare.GARE, message: str | None = None) -> None
...
+        click.echo(
+            click.style("MissingLoginError: ", fg="yellow")
+            + (
+                "No Authentication provided.\n"
+                "Please run:\n\n"
+                "    globus login\n\n"
+                "to ensure that you are logged in."
+            ),
+            err=True,
+        )
```
derek-globus, line 548: I'd recommend inverting these clauses.
The first is 3x as long as the second making it harder when reading the else block to see what conditional it's tied to ("This block executes if the 'if' paired with my 'else' is False, but the 'if' is off screen"). The inverted option is short enough that it doesn't.


```suggestion
    if not is_client_login():
        click.echo(
            click.style("MissingLoginError: ", fg="yellow")
            + (
                "No Authentication provided.\n"
                "Please run:\n\n"
                "    globus login\n\n"
                "to ensure that you are logged in."
            ),
            err=True,
        )
    else:
        click.echo(
            click.style("MissingLoginError: ", fg="yellow")
            + (
                "Invalid Authentication provided.\n\n"
                "'GLOBUS_CLI_CLIENT_ID' and 'GLOBUS_CLI_CLIENT_SECRET' are set but do "
                "not appear to be valid client credentials.\n"
                "Please check that the values are correctly set with no missing "
                "characters.\n"
            ),
            err=True,
        )
        if not _client_id_is_valid():
            click.secho(
                "'GLOBUS_CLI_CLIENT_ID' does not appear to be a valid client ID.",
                bold=True,
                fg="red",
                err=True,
            )
        if not _client_secret_appears_valid():
            click.secho(
                (
                    "'GLOBUS_CLI_CLIENT_SECRET' does not appear to "
                    "be a valid client secret."
                ),
                bold=True,
                fg="red",
                err=True,
            )
```

src/globus_cli/exception_handling/hooks.py
```diff
@@ -515,6 +504,80 @@ def _handle_gare(gare: globus_sdk.gare.GARE, message: str | None = None) -> None
...
+    try:
+        uuid.UUID(os.environ["GLOBUS_CLI_CLIENT_ID"])
+        return True
+    except ValueError:
+        return False
+
+
+def _client_secret_appears_valid() -> bool:
+    """
+    This check is known to be potentially incorrect if the encoding of Auth secrets
```
derek-globus, line 563: I'd include a higher level piece in the docstring explaining under which circumstances it should/shouldn't be used, not just the details of what makes it fallible. Doesn't need to be exhaustive; just something concise with less details as a framing sentence.

```suggestion
    """
    Used to diagnose an error associated with client credentials.
    
    This check is known to be potentially incorrect if the encoding of Auth secrets
```

derek-globus (changes requested):

src/globus_cli/exception_handling/hooks.py
sirosen: Yeah, I'm happy with that rename.

src/globus_cli/exception_handling/hooks.py
sirosen: If the size of the blocks is driving us to re-order, then I think that suggests that these should simply be separate functions. Either ordering should be equally readable, and if it isn't, I think that's telling us something else.

sirosen: I started retooling and decided that this just needs messages to be pulled out so that it reads more smoothly with more things inlined. After I get the update pushed, I'm happy to take more feedback on this, but I think it came out okay without needing more functional breakdown.

derek-globus (approved):

src/globus_cli/exception_handling/hooks.py
derek-globus: Looks good as-is in the latest branch. Thanks 👍

sirosen:
@derek-globus, just FYI, I have a vague idea about following up with some rearranging. I think we can make the hooks more maintainable with some elbow grease. Inspired a bit by the feedback on this PR.

PR #1086, opened 2025-03-05
Split up error hooks to be more modular, making ordering explicit

sirosen:
Two commits:

-   First a "boring" preparatory move to move some strings and messaging helpers to a distinct module
-   Second, "the change" itself: "Refactor to make error handler ordering explicit"

    Rather than making order driven by definition order, expose the declared hook type (was `_RegisteredHook`), and convert the decorated functions into the declared hook objects. Then, ordering can be done by any sorting function -- currently, a strict ordered listing with no implicit ordering semantics -- and we can explicitly register the results.

    This allows us to break the giant hook module down into its constituent parts which can be further refined and reorganized over time.


src/globus_cli/exception_handling/hooks.py
```diff
@@ -56,18 +63,8 @@ def _jsonpath_from_pydantic_loc(loc: list[str | int]) -> str:
...
-_DEFAULT_CONSENT_REAUTH_MESSAGE = (
-    "The resource you are trying to access requires you to "
-    "consent to additional access for the Globus CLI."
-)
-
-
 @sdk_error_handler(
-    error_class="GlobusAPIError",
-    condition=lambda err: err.raw_json is None,
+    error_class="GlobusAPIError", condition=lambda err: err.raw_json is None
```
kurtmckee: Do you have a linter set up that is rewriting these lines? This seems to have simply been a removal of a comma and a subsequent reflow of the lines throughout this commit.

src/globus_cli/exception_handling/hooks/auth_requirements.py
```diff
@@ -0,0 +1,91 @@
...
+
+    # specialized message for data_access errors
+    # otherwise, use more generic phrasing
+    if exception.message == "Missing required data_access consent":
+        message = (
+            "The collection you are trying to access data on requires you to "
+            "grant consent for the Globus CLI to access it."
+        )
+    else:
+        message = f"{DEFAULT_CONSENT_REAUTH_MESSAGE}\nmessage: {exception.message}"
```
kurtmckee, line 78:
```suggestion
        message = f"{DEFAULT_CONSENT_REAUTH_MESSAGE}\nMessage: {exception.message}"
```

src/globus_cli/exception_handling/hooks/auth_requirements.py
```diff
@@ -0,0 +1,91 @@
...
+    return None
+
+
+@sdk_error_handler(condition=lambda err: bool(err.info.consent_required), exit_status=4)
+def consent_required_hook(exception: globus_sdk.GlobusAPIError) -> int | None:
+    """
+    Expects an exception with a required_scopes field in its raw_json.
+    """
+    if not exception.info.consent_required.required_scopes:
+        msg = "Fatal Error: ConsentRequired but no required_scopes!"
```
kurtmckee, line 66:
```suggestion
        msg = (
            "Fatal Error: A ConsentRequired error was encountered "
            "but the error did not contain a required_scopes field!"
        )
```

src/globus_cli/exception_handling/hooks/endpoint_types.py
```diff
@@ -0,0 +1,24 @@
...
+def wrong_endpoint_type_error_hook(exception: WrongEntityTypeError) -> None:
+    msg = exception.expected_message + "\n" + exception.actual_message + "\n\n"
+    click.secho(msg, fg="yellow", err=True)
+
+    should_use = exception.should_use_command()
+    if should_use:
+        click.echo(
+            "Please run the following command instead:\n\n"
+            f"    {should_use} {exception.endpoint_id}\n",
+            err=True,
```
kurtmckee, line 20: slyp appears to have overlooked this.

src/globus_cli/exception_handling/hooks/generic_hooks.py
```diff
@@ -0,0 +1,51 @@
...
+        "GlobusAPINullDataError",
+        [PrintableErrorField("error_type", exception.__class__.__name__)],
+    )
+
+
+@sdk_error_handler(
+    error_class="GlobusAPIError", condition=lambda err: outformat_is_json()
+)
+def json_error_handler(exception: globus_sdk.GlobusAPIError) -> None:
+    msg = json.dumps(exception.raw_json, indent=2, separators=(",", ": "))
```
kurtmckee, line 27: We no longer need `separators` specified anywhere. The default changed in Python 3.4 to match what's written here and elsewhere.

```suggestion
    msg = json.dumps(exception.raw_json, indent=2)
```

src/globus_cli/exception_handling/messages.py
```diff
@@ -169,3 +171,9 @@ def emit_message_for_gare(
             domains=session_domains,
             message=message or DEFAULT_SESSION_REAUTH_MESSAGE,
         )
+
+
+def pretty_json(data: JsonValue, compact: bool = False) -> str:
+    if compact:
+        return json.dumps(data, separators=(",", ":"), sort_keys=True)
+    return json.dumps(data, indent=2, separators=(",", ": "), sort_keys=True)
```
kurtmckee, line 179:
```suggestion
    return json.dumps(data, indent=2, sort_keys=True)
```

kurtmckee (changes requested):

src/globus_cli/exception_handling/hooks.py
sirosen: Not tool-driven. The reflow you see was all manual -- I was trying to get a handle on the content here during the semantic refactor and touched a lot of the content.

I think the decorator lines are a little regrettable as a change I made at the time. If I could do it over again I'd leave those alone. But the message emission `click.echo` and `click.secho` usages were the parts I was actively rewriting and trying to find my way to a uniform style.

Do you want me to pull these back? (I assume not worth the effort to rewrite the history, but I'm happy to revert this part of the change if that would be nice.)

src/globus_cli/exception_handling/hooks/auth_requirements.py
sirosen: Let me apply this last. I agree with it, but that's the original/old behavior and I just want clear signal in case this surprises us with a broken test or two to fix.

src/globus_cli/exception_handling/hooks/endpoint_types.py
sirosen: This was the kind of usage I brought up recently (last week? I can't recall when). We seemed to have agreement not to autoformat such a usage with parens. I raised it because this feels to me like an odd outlier?

src/globus_cli/exception_handling/hooks/flows_hooks.py
```diff
@@ -0,0 +1,156 @@
...
+    return None
+
+
+@sdk_error_handler(
+    error_class="FlowsAPIError",
+    condition=lambda err: err.code == "UNPROCESSABLE_ENTITY",
+)
+def flows_validation_error_hook(exception: globus_sdk.FlowsAPIError) -> None:
+    message_string: str = exception.raw_json["error"]["message"]  # type: ignore[index]
+    details: str | list[dict[str, t.Any]] = exception.raw_json["error"]["detail"]  # type: ignore[index] # noqa: E501
```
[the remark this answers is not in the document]
sirosen: I think this comment might have been from commit-by-commit review, and the final took this in the direction you're talking about here?

`mypy` has some odd rules around narrowing the types of values accessed via `__getitem__`, so pulling things out into variables and then narrowing the types of those variables seems to work most reliably. (And it's actually pretty readable as a result.)

src/globus_cli/exception_handling/hooks/generic_hooks.py
sirosen: Showing my age here, I guess! I had completely missed that change when it went out. I'm so happy to not have to keep doing this (pointlessly, it turns out!).

Thanks!

src/globus_cli/exception_handling/hooks/auth_requirements.py
sirosen: As above, I want clear signal in case there's a test which was asserting on the old message. Unlike the prior case, I _expect_ breakage from this one, so I'm actually going to save it for last, making the above a penultimate change.

src/globus_cli/exception_handling/hooks.py
kurtmckee: No need to revert! If it was a linter I would just want to make sure that we agree to use it and then do a mass mechanical PR, rather than seeing extra changes coming through in a trickle. :+1:

src/globus_cli/exception_handling/hooks/endpoint_types.py
kurtmckee: I don't remember the conversation, but this definitely doesn't match my expectations.

src/globus_cli/exception_handling/hooks/flows_hooks.py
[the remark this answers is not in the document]
kurtmckee: Ah, yep, this was commit-by-commit. Great!

src/globus_cli/exception_handling/hooks/generic_hooks.py
kurtmckee: Yep! I couldn't imagine that this was necessary but had to check the docs, which more-or-less say "changed in Python 3.4 so the defaults are sane". :laughing:

src/globus_cli/exception_handling/hooks/endpoint_types.py
sirosen: It was ~2 weeks ago by the look of things in our chat logs (March 6th). I shared a potential motivating example file and you agreed pretty strongly that the example showed poor style, but we didn't have full team agreement that we should auto-fix it.

I raised this when I encountered this at the time of the original work because I also found it surprising. So you're not alone in having your expectation violated here. Let me start with the example file I used back on the 6th, and then I'll give a little more detail on why the fixer would need a change for this case. I asked if we should change fixing behavior under the following scenarios...

<details><summary><code>scenarios.py</code></summary>

```python
import typing as t


# the demo function
def style(
    message: str,
    *,
    color: str | None = None,
    bold: bool = False,
    border: str | None = None,
    comic_sans: t.Literal["no", "never", "nice-try", "lolnope"] = "no",
    mono: bool = True,
) -> str: ...


# Usage 1: Just a message
# we know what we want in this case

style("a small message")  # good!
style(  # also good!
    "a message spanning...                                             "
    "                     spanning...                                  "
    "                                spanning...                       "
    "                                           many lines!            "
)
style(  # bad, we do not want this!
    (
        "a message spanning...                                          "
        "                     spanning...                               "
        "                                 many lines!                   "
    )
)


# Usage 2: Message + one arg
# some questions

style("a small message", bold=True)  # good!
style(  # also good?
    "a message spanning...                                             "
    "                     spanning...                                  "
    "                                spanning...                       "
    "                                           many lines!            ",
    color="yellow",
)
style(  # better? worse? equivalent?
    (
        "a message spanning...                                          "
        "                     spanning...                               "
        "                                many lines!                    "
    ),
    comic_sans="never",
)

# Usage 3: Message + two args
style("a small message", bold=True, mono=False)  # good!
style(  # also good?
    "a message spanning...                                             "
    "                  spanning...                                     "
    "                          spanning...                             "
    "                                  spanning...                     "
    "                                          spanning...             "
    "                                                     many lines!  ",
    color="yellow",
    border="dotted",
)
style(  # better? worse? equivalent?
    (
        "a message spanning...                                             "
        "                  spanning...                                     "
        "                          spanning...                             "
        "                                  spanning...                     "
        "                                          spanning...             "
        "                                                     many lines!  "
    ),
    comic_sans="nice-try",
    color="mauve",
)

# Usage 4: Message + many args
style(  # good!
    "a small message",
    color="prussian-blue",
    bold=False,
    border="happy-little-cloud",
    comic_sans="lolnope",
    mono=True,
)
style(  # also good?
    "a message spanning...                                             "
    "                     spanning...                                  "
    "                                spanning...                       "
    "                                           many lines!            ",
    color="prussian-blue",
    bold=False,
    border="happy-little-cloud",
    comic_sans="lolnope",
    mono=True,
)
style(  # better? equivalent?
    (
        "a message spanning...                                          "
        "                     spanning...                               "
        "                                spanning...                    "
        "                                           many lines!         "
    ),
    color="prussian-blue",
    bold=False,
    border="happy-little-cloud",
    comic_sans="lolnope",
    mono=True,
)
```

</details>

The reason we'd need an additional rule for this case or cases like it is that we know of a case we definitely _don't_ want fixed, which is
```python
print(
    "a.............................................................."
    "very long................................."
    "multiline string"
)
```

There's a carve-out in the auto-parenthesizing fix for the case of "there is a multiline string as the only positional argument". Usage with only one positional _might_ be okay, but seems to stretch the intent of the fixer a bit:
```python
print(
    "a.............................................................."
    "very long................................."
    "multiline string",
    file=sys.stderr,
)
```

But if you stack a lot of keyword args I think it's clear that it's wrong:
```python
print(
    "a.............................................................."
    "very long................................."
    "multiline string",
    file=sys.stderr,  # please, someone!
    sep=",",  # save us!
    end="...\n",  # oh the humanity! 
)
```

I therefore phrased the initial thread as being "is there a threshold number of kwargs which should make us rethink that rule about parenthesizing the solitary positional string arg?"

One possible conclusion is "yes, and that threshold is 1". For now, status-quo wins, but I'm open to adjusting this part of the behavior -- so long as we get solid team buy-in when we do it.

kurtmckee: I think I remember the conversation now.

Having an f-string immediately after is a trick to my brain when reviewing this. Like "ah, first keyword argum...oh, that 'f' was the modifier to a string literal, not the first letter of a keyword parameter."

Obviously it's a microsecond thing, but I'm dissatisfied with this formatting.

kurtmckee (approved):

PR #1096, opened 2025-04-09
Change the name of the Globus Timers service rendered in the `globus api timer` command

kurtmckee:
_Observed while working on a different task._

----

The goal of this change is to fix the name of the Globus Timers service in the `globus api timer` help text.

In addition, the rendering of the example service/route request URL should now be code text.

image

### DIFF
--- a/src/globus_cli/commands/api.py
+++ b/src/globus_cli/commands/api.py
@@ -370,7 +370,7 @@ For example, a call of
 
     globus api {legacy_name} GET /foo/bar
 
-sends a 'GET' request to '{_get_url(service_name)}foo/bar'
+sends a 'GET' request to `{_get_url(service_name)}foo/bar`
 """
 
     if service_name != "gcs":
--- a/src/globus_cli/commands/api.py
+++ b/src/globus_cli/commands/api.py
@@ -134,7 +134,7 @@ def _get_resource_server(service_name: str) -> str:
         "groups": globus_sdk.GroupsClient.resource_server,
         "search": globus_sdk.SearchClient.resource_server,
         "transfer": globus_sdk.TransferClient.resource_server,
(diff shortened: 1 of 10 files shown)


sirosen (commented):
I'm having a really hard time figuring out what I think of this change. It manipulates a lot of internal details to produce different generated docs.

I don't really like the introduction of `legacy_name` because it looks like the declaration of a deprecation without a migration strategy. I'd rather that we start supporting both `globus api timers` and `globus api timer`, but mark `globus api timer` hidden so that it doesn't go into the generated docs.

Would you be willing to expand this PR in that direction? It would give me a clearer handle as a reviewer on what we're changing and what kinds of change I should look for and expect, and I think it's pretty easy to make that change. (Most of these same changes, `api_command.add_command(build_command("timer"))`, and some special casing for `"timer"`.)

kurtmckee:
> I don't really like the introduction of `legacy_name` because it looks like the declaration of a deprecation without a migration strategy.
>
> I'd rather that we start supporting both `globus api timers`

I changed a Python type annotation to change the name of the service on an HTML page on the Globus Docs website. It could have been done with a lone conditional in `build_command()` instead, but the service name is now correct throughout the code base.

Only the subcommand name "timer" remains, and I'm not trying to work on that now.

sirosen (commented):
Okay, I'll go after the command aliasing in a follow-up. This doesn't really leave us any worse off.

I still think that if your only goal were to change the doc page, you could have updated the helptext
```diff
-Make API calls to Globus {service_name.title()}
+Make API calls to Globus {("timers" if service_name == "timer" else service_name).title()}
```

So the fact that you've made the systematic change to update the internal name of the service sends a clear message to me that you're trying to take the first step in a migration.

sirosen (approved):

PR #1098, opened 2025-04-11
Add support for flow run roles to flow commands

m1yag1:
[sc-40144](https://app.shortcut.com/globus/story/40144/cli-add-support-for-flow-run-roles-to-flow-commands)

This PR is waiting until https://github.com/globus/globus-sdk-python/pull/1164 is merged and released to the SDK. At that point, the testing data will be available and allow this to be merged.

### DIFF
--- a/changelog.d/20250411_105010_8730430+m1yag1.md
+++ b/changelog.d/20250411_105010_8730430+m1yag1.md
@@ -1,0 +1,5 @@
+### Enhancements
+
+* Added the `--run-managers` and `--run_monitors` options to the `globus flows create` command
+  and the `globus flows update` command.
+* Added "Run Managers" and "Run Monitors" to the output of the `globus flows show` command.
--- a/changelog.d/20250411_105010_8730430+m1yag1.md
+++ b/changelog.d/20250411_105010_8730430+m1yag1.md
@@ -1,5 +1,5 @@
 ### Enhancements
 
-* Added the `--run-managers` and `--run_monitors` options to the `globus flows create` command
-  and the `globus flows update` command.
-* Added "Run Managers" and "Run Monitors" to the output of the `globus flows show` command.
+* Added `--run-manager` and `--run-monitor` options to `globus flows create`.
+* Added `--run-managers` and `--run-monitors` options to `globus flows update`.
+* Added "Run Managers" and "Run Monitors" to the output of `globus flows show`.
--- a/src/globus_cli/commands/flows/_common.py
+++ b/src/globus_cli/commands/flows/_common.py
@@ -115,6 +115,32 @@ keywords_option = click.option(
     """,
 )
 
+run_managers_option = click.option(
+    "--run-manager",
+    "run_managers",
+    type=str,
+    multiple=True,
+    help="""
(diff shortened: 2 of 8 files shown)


changelog.d/20250411_105010_8730430+m1yag1.md
```diff
@@ -0,0 +1,5 @@
+### Enhancements
+
+* Added the `--run-managers` and `--run_monitors` options to the `globus flows create` command
+  and the `globus flows update` command.
+* Added "Run Managers" and "Run Monitors" to the output of the `globus flows show` command.
```
sirosen, line 5:
```suggestion
* Added `--run-manager` and `--run-monitor` options to `globus flows create`.
* Added `--run-managers` and `--run-monitors` options to `globus flows update`.
* Added "Run Managers" and "Run Monitors" to the output of `globus flows show`.
```

I'm not a big fan of the `--run-manager`/`--run-managers` dichotomy in this PR -- I see this isn't the first case of this -- but if it's going to exist we need to be careful to document it accurately.

src/globus_cli/commands/flows/_common.py
```diff
@@ -115,6 +115,32 @@
...
+    type=str,
+    multiple=True,
+    help="""
+        A principal that may monitor the flow's runs.
+
+        This option can be specified multiple times
+        to create a list of run monitors.
+    """,
+)
+
```
sirosen, line 143: These are only used in the `create` command right now, correct?

I would prefer that we not put them into this `_common` module and instead declare them inline on the appropriate command.

The `_common` modules (there are several, for different families of commands) are for things which are being shared across commands.
Putting things here gives the false impression that we have a kind of uniformity between `create` and `update` which we don't have.

Fixing the divergence between those two commands is clearly beyond the scope of this PR, but we don't need to continue to declare things in a way which may mislead us in the future into not understanding how much of our code is _not_ being reused across components.

tests/functional/flows/test_show_flow.py
```diff
@@ -98,6 +103,24 @@ def test_show_flow_text_output(run_line):
...
+                        },
+                        {
+                            "username": "gandalf@rivendell.middleearth",
+                            "name": "Ian McKellen",
+                            "id": gandalf_id,
+                            "identity_provider": "c8abac57-560c-46c8-b386-f116ed8793d5",
+                            "organization": "Fellowship of the Ring",
+                            "status": "used",
+                            "email": "gandalf@rivendell.middleearth",
+                        },
```
sirosen, line 123: :heart_hands:

sirosen (commented):
Minor changes requested; but also this needs a rebase.
Overall I think this should be an easy merge with some tweaks.

There's something I really don't like about how `globus flows create` and `globus flows update` have diverged, but we'll have to settle that some other time, for all roles.

src/globus_cli/commands/flows/_common.py
m1yag1: @sirosen This should be resolved now along with the changelog item.

sirosen (approved):

PR #1102, opened 2025-04-18
Implement flow start notification policy parsing

sirosen:
`globus flows start` has a new option,
`--activity-notification-policy`, with specialized custom parsing. It
accepts either a comma-delimited list or a JSON blob (or file),
deriving from our existing JSON support.

A simple heuristic tries to determine which type of input is being
given, checking if it "looks like comma delimited names" and
dispatching between the parse paths as appropriate.
The same heuristic drives tab completion, letting the completer switch
between the two modes.

Unit tests confirm the functionality to ensure that it parses as
desired in a variety of scenarios, and emits appropriate errors.

The resulting parsed policy object is then simply passed to
`start_flow`.

Additionally, in order to provide good helptext for this option, the
overall helptext for `globus flows start` has been enhanced.

### DIFF
--- a/src/globus_cli/commands/flows/start.py
+++ b/src/globus_cli/commands/flows/start.py
@@ -206,8 +206,9 @@ class ActivityNotificationPolicyType(JSONStringOrFile):
     help="""
         The activity notification policy for the run.
 
-        This may be given as a JSON file or JSON data containing a full policy
-        document, or a comma-delimited list of statuses for notification.
+        This may be given as a comma-delimited list of statuses for notification;
+        alternatively, this can also be provided as JSON data--or a path to a
+        JSON file--containing a full notification policy document.
     """,
 )
 @LoginManager.requires_login("flows")
--- a/src/globus_cli/commands/flows/start.py
+++ b/src/globus_cli/commands/flows/start.py
@@ -1,5 +1,8 @@
 from __future__ import annotations
 
+import os
+import string
+import typing as t
 import uuid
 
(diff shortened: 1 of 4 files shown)


MaxTueckeGlobus (approved):

src/globus_cli/commands/flows/start.py
```diff
@@ -78,6 +200,16 @@
...
     """,
 )
+@click.option(
+    "--activity-notification-policy",
+    type=ActivityNotificationPolicyType(),
+    help="""
+        The activity notification policy for the run.
+
+        This may be given as a JSON file or JSON data containing a full policy
+        document, or a comma-delimited list of statuses for notification.
```
ada-globus, line 210: Current phrasing makes it sound like JSON is preferred, but I don't think that's what we should suggest.

How about:

```suggestion
        This may be given as a comma-delimited list of statuses for notification;
        alternatively, this can also be provided as JSON data--or a path to a
        JSON file--containing a full notification policy document.
```

ada-globus (approved):
Noting here since our discussion around this was not in the PR itself: I'm okay with moving forward on this provisionally, though I'm still unsure how I feel about the path this suggests for future complex options (i.e., this JSON support nudges us down a particular path, and may continue to prove more complex if we continue to adopt this convention in more places, as this has become quite a bit of overloading). I'm holding this as a something of a trial balloon that we could later back away from if this proves more troublesome to generalize than hoped.

PR #1105, opened 2025-04-23
add support for activity-notifications for GCP and GCS

pjhinton-globus:
The code in this PR  implements a syntax based on the GCS CLI's approach to configuring guest collection activity notification policies. Guest collection create and update operations accept a new command line option:

--activity-notifications

To enable notifications under any circumstance, simply set the value of the option to `all`. Specifying an empty string (`""`) will disable all notifications.

Finer grained policies may be specified through a syntax that employs comma-separated strings corresponding to the role of the collection in the transfer (`transfer_use`) and outcome of the transfer (`status`).

The valid values for status are `"succeeded"` and `"failed"`. For `transfer_use`, `"source"` and `"destination"` are accepted.

The syntax builds up the policy structure that looks like this:

```
{
    "status": [(string), (string)...],
    "transfer_use": [(string), (string)...]
}
```

Specifying a policy value for one of the attributes while leaving out values of the others will implicitly enable all other values from the other attribute. For example, if `failed` is given, this is equal to the policy:

```
{
    "status": ["FAILED"],
    "transfer_use": ["source", "destination"]
}
```

Specifying one value for each attribute enables only those situations. Fore example, specifying the value `destination,succeeded` is equivalent to:

```
{
    "status": ["SUCCEEDED"],
    "transfer_use": ["destination"]
}
```

The syntax above is implemented through two click.ParamType subclasses:

`GCSManagerGuestActivityNotificationParamType`
`TransferGuestActivityNotificationParamType`

Two classes exist because of some differences between the GCS Manager API and the Transfer API.  GCS collection document schema 1.14.0 implements the status values as all lowercase, whereas Transfer expresses them as all capitals because that is how they are represented elsewhere in the API.

Because of the way the code for the GCS commands weed out null values from payloads, the `convert()` methods emit different structures for the `"" `string case. For the GCS class, the structure looks like:

```
{
    "status": [],
    "transfer_use": []
}
```

whereas the Transfer class yields an `ExplictNullType` instance.

Both routes rely on a common decorator (`activity_notifications_option`), which takes a string value of `"GCS"` or `"GCP"` to determine whether the data type class associated with the option will be the GCS or Transfer variant.

There is also support for shell completion on this option. Code borrows heavily (if not shamelessly) from the `NotifyParamType`.

There is complete unit test coverage for option parsing and shell completion.

A changelog fragment for `--activity-notifications` is provided.

### DIFF
--- a/src/globus_cli/parsing/param_types/guest_activity_notify_param.py
+++ b/src/globus_cli/parsing/param_types/guest_activity_notify_param.py
@@ -1,0 +1,200 @@
+from __future__ import annotations
+
+import click
+from click.shell_completion import CompletionItem
+
+from globus_cli._click_compat import shim_get_metavar
+from globus_cli.constants import ExplicitNullType
+
+
+class GCSManagerGuestActivityNotificationParamType(click.ParamType):
+    """
+    For the GCS Manager API:
+
+    * Status values are lowercase strings.
+
+        {
+            "status": ["failed", "succeeded],
+            ...
+        }
+
+    * Disabling all notifications is expressed by setting all elements to
+      empty lists.
+
+        {
+            "status": [],
+            "transfer_use": []
+        }
+    """
+
+    VALID_STATUSES = {
+        "succeeded",
+        "failed",
+    }
+
+    VALID_TRANSFER_USES = {
+        "source",
+        "destination",
+    }
+
+    VALID_NOTIFICATION_VALUES = VALID_TRANSFER_USES | VALID_STATUSES
+
+    @shim_get_metavar
+    def get_metavar(self, param: click.Parameter, ctx: click.Context) -> str:
+        return "{all,succeeded,failed,source,destination}"
+
+    def convert(
+        self,
+        value: str,
+        param: click.Parameter | None,
+        ctx: click.Context | None,
+    ) -> dict[str, list[str]] | ExplicitNullType:
+
+        if value == "":
+            return {
+                "status": [],
+                "transfer_use": [],
+            }
+
+        if value.lower() == "all":
+            return {
+                "status": sorted(self.VALID_STATUSES),
+                "transfer_use": sorted(self.VALID_TRANSFER_USES),
+            }
+
+        policy: dict[str, list[str]] = {
+            "status": [],
+            "transfer_use": [],
+        }
+
+        # ignore white space, parse input sans case-sensitivity
+        lowercase_vals: set[str] = {s.strip().lower() for s in value.split(",") if s}
+
+        if "all" in lowercase_vals:
+            raise click.UsageError(
+                '--activity-notifications cannot accept "all" with other values'
+            )
+
+        val: str
+        if lowercase_vals <= self.VALID_NOTIFICATION_VALUES:
+            for val in lowercase_vals:
+                if val in self.VALID_TRANSFER_USES:
+                    policy["transfer_use"].append(val)
+                else:
+                    policy["status"].append(val)
+        else:
+            invalid_values = sorted(lowercase_vals - self.VALID_NOTIFICATION_VALUES)
+            raise click.UsageError(
+                "--activity-notifications received these invalid values: "
+                f"{invalid_values}"
+            )
+
+        # Fill in implied values.
+        # If statuses were given but no uses, all uses are listed.
+        # If uses were given but no statuses, all statuses are listed.
+        for k, v in policy.items():
+            if k == "transfer_use" and not v:
+                policy[k] = list(self.VALID_TRANSFER_USES)
+            elif k == "status" and not v:
+                policy[k] = list(self.VALID_STATUSES)
+
+        return policy
+
+    def shell_complete(
+        self, ctx: click.Context, param: click.Parameter, incomplete: str
+    ) -> list[CompletionItem]:
+        all_compoundable_options = ["destination", "failed", "source", "succeeded"]
+
+        all_options = ["all"] + all_compoundable_options
+
+        # if the caller used `--activity_notifications <TAB>`, show all options
+        # if the caller used `--activity_notifications <TAB>`, show `source` and
+        # `succeeded` the logic below assumes there were commas
+        if "," not in incomplete:
+            return [CompletionItem(o) for o in all_options if o.startswith(incomplete)]
+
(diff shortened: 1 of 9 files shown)


src/globus_cli/parsing/param_types/guest_activity_notify_param.py
```diff
@@ -0,0 +1,171 @@
...
+        # Fill in implied values.
+        k: str
+        v: t.List[str]
+        for k, v in policy.items():
+            if k == "transfer_use" and not v:
+                v = list(self.VALID_TRANSFER_USES)
+            elif k == "status" and not v:
+                v = list(self.VALID_STATUSES)
+            # Ensure deterministic ordering for testability.
+            policy[k] = sorted(v)
```
sirosen, line 82: There are three minor changes I'd like here.

One is that I think the comment at the top about "implied values" should be a bit more explicit about what/why it's implied. I've suggested what I think is right for that but please review/tweak as appropriate.

The second is that the type of `policy` is declared above, so we shouldn't need to declare the types of `k` and `v`.

The last is that `policy[k] = sorted(v)` looks strange to me because we're basically doing `sorted(list(...))` and we can remove one of the two. I've suggested _a_ way of altering it, but others (e.g., `v = sorted(self.VALID_...)`) are also possible.

```suggestion
        # Fill in implied values.
        # If statuses were given but no uses, all uses are listed.
        # If uses were given but no statuses, all statuses are listed.
        for k, v in policy.items():
            # use sorted lists for deterministic ordering, for better testability
            if k == "transfer_use" and not v:
                policy[k] = sorted(self.VALID_TRANSFER_USES)
            elif k == "status" and not v:
                policy[k] = sorted(self.VALID_STATUSES)
```

src/globus_cli/parsing/param_types/guest_activity_notify_param.py
```diff
@@ -0,0 +1,171 @@
...
+
+        return [
+            CompletionItem(",".join(already_contains + [o])) for o in possible_options
+        ]
+
+
+class TransferGuestActivityNotificationParamType(
+    GCSManagerGuestActivityNotificationParamType
+):
+
```
sirosen, line 154: Can we put a class docstring in here which explains what this changes vs the GCS version?
My understanding is that:
- "no policy" is represented with nulls, not empty lists
- upper-case names are used, not lower-case

sirosen (changes requested):
Overall, this was pretty readable -- especially with the context of the existing `--notify` param -- and I just have some minor points we should address.

src/globus_cli/parsing/shared_options/__init__.py
```diff
@@ -376,3 +380,44 @@ def local_user_option(f: C) -> C:
...
+            )(f)
+        else:
+            return click.option(
+                "--activity-notifications",
+                default=None,
+                help=help,
+                type=GCSManagerGuestActivityNotificationParamType(),
+            )(f)
+
+    return decorator
```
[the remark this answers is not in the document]
pjhinton-globus: I like this suggested revision of the help text a lot, but I do have a reservation about the use of "activity monitors" in the second sentence. I chose the phrasing "activity-monitoring roles" originally because I wanted to convey that higher roles would receive these notifications, too.

If we already have established conventions for using "activity monitors" to imply not only `activity_monitor` but any higher role whose privileges include `activity_monitor`, I would be more at ease about that change.

pjhinton-globus:
@sirosen wrote:

> Overall, this was pretty readable -- especially with the context of the existing `--notify` param -- and I just have some minor points we should address.

I think I've addressed all of the change requests in spirit, if not in strict letter. :-)

See commit 0a6bfd2919f54f479f5e2163eb5570d4cd8b3008 for the changes.

If these wind up looking acceptable, I'll do a `rebase --interactive` against `main` to roll things up into one commit for final approval.

sirosen:
Yep, everything here looks good! I'll go through and mark the various threads resolved.

Before you rebase, it's probably best to revert the constraint on `click` -- `main` now has changes for support for `click` v8.2.0+. The only thing you should need to do after the rebase is to update `get_metavar` to use the compatibility decorator I added:

```python
from globus_cli._click_compat import shim_get_metavar
```

```diff
+    @shim_get_metavar
+    def get_metavar(self, param: click.Parameter, ctx: click.Context) -> str:
-    def get_metavar(self, param: click.Parameter) -> str:
         return "{all,succeeded,failed,source,destination}"
```

src/globus_cli/parsing/shared_options/__init__.py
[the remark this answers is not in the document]
sirosen: I just double-checked, and it looks like you may have used the "activity monitors" wording in your last commit. "Activity-monitoring roles", as you suggest, seems good, especially if that's more likely to be consistent with API docs, &c. I'll leave it to you whether or not you want to change it from the current state.

tests/unit/param_types/test_guest_activity_notification_param.py
```diff
@@ -0,0 +1,253 @@
+import click
+import pytest
+
+from globus_cli.parsing.param_types.guest_activity_notify_param import (
+    GCSManagerGuestActivityNotificationParamType,
+    TransferGuestActivityNotificationParamType,
+)
+
+
+@click.command()
```
sirosen, line 10: The tests are failing for a relatively silly reason -- as of v8.2.0, `click` now strips suffixes of `_command` or `_cmd` from automatic command names.
You could do it some other way, but I think the simplest quick fix is this:

```suggestion
@click.command("activity-notifications-cmd")
```

pjhinton-globus: > The tests are failing for a relatively silly reason -- as of v8.2.0, `click` now strips suffixes of `_command` or `_cmd` from automatic command names. You could do it some other way, but I think the simplest quick fix is this:

I just now saw this because I was busy trying to figure out the cause for the test failure. The approach I took was to adopt a less stringent form of assert like the ones used for similar errors tests in `test_notification_param.py`.

https://github.com/globus/globus-cli/blob/af12bb90a6396f99f030b86a3c67f8c651fd4e6b/tests/unit/param_types/test_notification_param.py#L81-L93

sirosen (approved):

PR #1106, opened 2025-04-30
Add support for `filter_roles` option when listing flows and runs

m1yag1:
[sc-40419](https://app.shortcut.com/globus/story/40419)
### Enhancements
 
 * Updated `--filter-role` parameter in  `globus flows list` to be
   supplied multiple times to specify multiple roles.
 * Added `--filter-role` parameter to `globus flows run list`, which can
   be supplied multiple times to specify multiple roles.

### DIFF
--- a/changelog.d/20250430_152200_8730430+m1yag1_sc_40419_filter_roles_list_flows_and_runs.md
+++ b/changelog.d/20250430_152200_8730430+m1yag1_sc_40419_filter_roles_list_flows_and_runs.md
@@ -1,6 +1,6 @@
 
 ### Enhancements
 
-* Updated `--filter-roles` parameter in  `globus flows list` to be
+* Updated `--filter-role` parameter in  `globus flows list` to be
   supplied multiple times to specify multiple roles.
-* Added `--filter-roles` parameter to `globus flows run list`.
+* Added `--filter-role` parameter to `globus flows run list`.
--- a/changelog.d/20250430_152200_8730430+m1yag1_sc_40419_filter_roles_list_flows_and_runs.md
+++ b/changelog.d/20250430_152200_8730430+m1yag1_sc_40419_filter_roles_list_flows_and_runs.md
@@ -1,7 +1,6 @@
 
 ### Enhancements
 
-* Updated `--filter-role` parameter in  `globus flows list` to be
+* Updated `--filter-roles` parameter in  `globus flows list` to be
   supplied multiple times to specify multiple roles.
-* Added `--filter-role` parameter to `globus flows run list` which can
-  be supplied multiple times to specify multiple roles.
+* * Added `--filter-roles` parameter to `globus flows run list`.
--- a/changelog.d/20250430_152200_8730430+m1yag1_sc_40419_filter_roles_list_flows_and_runs.md
+++ b/changelog.d/20250430_152200_8730430+m1yag1_sc_40419_filter_roles_list_flows_and_runs.md
@@ -1,0 +1,7 @@
+
+### Enhancements
+
+* Updated `--filter-role` parameter in  `globus flows list` to be
+  supplied multiple times to specify multiple roles.
+* Added `--filter-role` parameter to `globus flows run list` which can
+  be supplied multiple times to specify multiple roles.
--- a/changelog.d/20250430_152200_8730430+m1yag1_sc_40419_filter_roles_list_flows_and_runs.md
+++ b/changelog.d/20250430_152200_8730430+m1yag1_sc_40419_filter_roles_list_flows_and_runs.md
@@ -3,4 +3,4 @@
 
 * Updated `--filter-roles` parameter in  `globus flows list` to be
   supplied multiple times to specify multiple roles.
-* * Added `--filter-roles` parameter to `globus flows run list`.
+* Added `--filter-roles` parameter to `globus flows run list`.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ classifiers = [
 ]
 requires-python = ">=3.8"
 dependencies = [
-    "globus-sdk==3.55.0",
+    "globus-sdk==3.56.0",
     "click>=8.1.4,<9",
     "jmespath==1.0.1",
     "packaging>=17.0",
--- a/src/globus_cli/commands/flows/list.py
+++ b/src/globus_cli/commands/flows/list.py
@@ -36,6 +36,7 @@ ORDER_BY_FIELDS = (
 @command("list")
 @click.option(
     "--filter-role",
+    "filter_roles",
     type=click.Choice(ROLE_TYPES),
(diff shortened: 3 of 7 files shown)


changelog.d/20250430_152200_8730430+m1yag1_sc_40419_filter_roles_list_flows_and_runs.md
```diff
@@ -0,0 +1,7 @@
+
+### Enhancements
+
+* Updated `--filter-role` parameter in  `globus flows list` to be
+  supplied multiple times to specify multiple roles.
+* Added `--filter-role` parameter to `globus flows run list` which can
+  be supplied multiple times to specify multiple roles.
```
sirosen, line 7: Minor, but for a new flag I'd tend to reduce the level of detail in the changelog so that it stands out better.

```suggestion
* Added `--filter-role` parameter to `globus flows run list`.
```

I think in terms of how I read changelogs when I'm trying to figure out when a feature was added.

src/globus_cli/commands/flows/list.py
```diff
@@ -98,7 +113,7 @@ def list_command(
     paginator = Paginator.wrap(flows_client.list_flows)
     flow_iterator = PagingWrapper(
         paginator(
-            filter_role=filter_role,
+            filter_roles=",".join(filter_role),
```
sirosen, line 116: Given that the SDK will comma-join, I'd generally prefer that we use `filter_roles=filter_role`.
However, I see that we have it annotated as `list[str]`, which is overly strict. I've put up a fix PR, but in the meantime, perhaps `filter_roles=list(filter_role)`?

src/globus_cli/commands/flows/list.py
```diff
@@ -31,6 +38,7 @@
     "--filter-role",
     type=click.Choice(ROLE_TYPES),
     help="Filter results by the flow's role type associated with the caller",
+    multiple=True,
 )
```
sirosen, line 42: This creates a multiple-use option whose resulting parameter name is `filter_role`.
I think it would be preferable to update to
```python
@click.option(
    "--filter-role",
    "filter_roles"
    ...
)
```

Which sets the destination name to `filter_roles`. That way, we don't have a singular/plural discrepancy at the usage sites.

src/globus_cli/commands/flows/run/list.py
```diff
@@ -45,7 +73,10 @@ def list_command(
 
     paginator = Paginator.wrap(flows_client.list_runs)
     run_iterator = PagingWrapper(
-        paginator(filter_flow_id=filter_flow_id).items(),
+        paginator(
+            filter_flow_id=filter_flow_id,
+            filter_roles=",".join(filter_role),
```
sirosen, line 78: Similarly to the list_flows case, I might prefer `list(...)` to comma-joining when the SDK already does so.
But actually my favorite option is to write this as
```python
filter_roles=filter_role,  # type: ignore[arg-type]
```

Because it will work today and as soon as we ship an SDK version which fixes the annotation, it will be flagged as an unused ignore by `mypy`.

tests/functional/flows/test_list_runs.py
```diff
@@ -74,3 +74,35 @@ def test_list_runs_paginated_response(run_line, limit_delta):
...
+    assert header_row == ["Run ID", "Flow Title", "Run Label", "Status"]
+
+
+def test_list_runs_invalid_filter_role(run_line):
+    load_response("flows.list_runs")
+
+    run_line(
+        "globus flows run list --filter-role up-up-down-down-left-right-left-right-b-a",
+        assert_exit_code=2,
+    )
```
sirosen, line 108: It looks like these tests might be missing some assertions?

sirosen: Specifically:
- the one above only asserts the header row
- this one should probably check that the usage error message is what we expect

src/globus_cli/commands/flows/list.py
m1yag1: Yea, i should have changed this. i didn't revisit this code after adding commajoin in the other PR. 🤦

tests/functional/flows/test_list_runs.py
m1yag1: The above only asserts the header row b/c I felt the other tests did a good job testing the table output. I don't mind adding that.

I can add an error message assertion to the invalid filter role test, good call. I normally do add expected error messages to my tests (not sure why i didn't this time).

pyproject.toml
```diff
@@ -9,6 +9,7 @@ readme = "README.rst"
 authors = [
     { name = "Stephen Rosen", email = "sirosen@globus.org" },
 ]
+license = "Apache-2.0"
```
m1yag1, line 12: I wanted to mention this as something new in the last set of changes. In the tox quality tests, I noticed this was being mentioned. I can remove it if it shouldn't be in this set of changes.

This is the deprecation warning I was seeing.
```
/tmp/build-env-r91w10ei/lib/python3.13/site-packages/setuptools/config/_apply_pyprojecttoml.py:61: SetuptoolsDeprecationWarning: License classifiers are deprecated.
!!

        ********************************************************************************
        Please consider removing the following classifiers in favor of a SPDX license expression:

        License :: OSI Approved :: Apache Software License

        See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
        ********************************************************************************
```

sirosen: I would strongly prefer that this be done in a separate PR. We can merge it pretty easily, but it switches the core metadata fields in our published dists and deserves separate space and tracking in our project change history.

changelog.d/20250430_152200_8730430+m1yag1_sc_40419_filter_roles_list_flows_and_runs.md
```diff
@@ -0,0 +1,6 @@
+
+### Enhancements
+
+* Updated `--filter-roles` parameter in  `globus flows list` to be
```
sirosen, line 4:
```suggestion
* Updated `--filter-role` parameter in  `globus flows list` to be
```

changelog.d/20250430_152200_8730430+m1yag1_sc_40419_filter_roles_list_flows_and_runs.md
```diff
@@ -0,0 +1,6 @@
...
+  supplied multiple times to specify multiple roles.
+* Added `--filter-roles` parameter to `globus flows run list`.
```
sirosen, line 6:
```suggestion
* Added `--filter-role` parameter to `globus flows run list`.
```

pyproject.toml
m1yag1: @sirosen, I dropped those commits not to impact git blame/history, and force pushed.

sirosen (approved):

PR #1115, opened 2025-05-22
Remove README badges

kurtmckee:
These represent a risk, as they rely on external services (like shields.io) and current repository state ("build status").

Driving home that point, the build status badge is broken and this has harmed the rendering of all past versions' READMEs hosted on PyPI.


sirosen (approved):
Yeah, the harm of bit-rot for these is real.
I suppose I _wish_ for a world in which these worked differently, but dealing with the reality of them I agree with your judgement that we should remove.

CI is failing, but I assume that's somehow unrelated -- maybe some drift which broke `main`?

PR #1123, opened 2025-06-05
Reorder (consent_required_hook, session_hook)

sirosen:
These two error hooks are adjacent to one another in the CLI's error
handling hooks, and their ordering has historically not been
significant. Prior to formulation of the GARE format, the two error
classes they handled were mutually exclusive.

Now that APIs can respond with GARE-formatted ConsentRequired data,
it's important that the ConsentRequired handler
(consent_required_hook) is checked first.

A new unit test captures this change.

### DIFF
--- a/tests/functional/exception_handling/test_session_hooks.py
+++ b/tests/functional/exception_handling/test_session_hooks.py
@@ -4,6 +4,40 @@ import pytest
 from globus_sdk._testing import RegisteredResponse
 
 
+def tests_consent_required_gets_preference_over_authorization_parameters(run_line):
+    """
+    Confirm that when an error matches both the 'session update' and 'session consent'
+    hook rules, we choose the 'session consent' option.
+
+    The likely way for this to happen is a GARE-formatted error for ConsentRequired.
+    """
+    ep_id = str(uuid.UUID(int=1))
+
+    required_scope = (
+        "urn:globus:auth:scope:transfer.api.globus.org:all"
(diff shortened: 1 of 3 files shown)


tests/functional/exception_handling/test_session_hooks.py
```diff
@@ -4,6 +4,40 @@
...
+                "session_message": "Missing required data_access consent",
+            },
+            "code": "ConsentRequired",
+            "message": "Missing required data_access consent",
+            "request_id": "m8RfX3cES",
+            "required_scopes": [required_scope],
+            "resource": f"/operation/endpoint/{ep_id}/ls",
+        },
+    ).add()
+    result = run_line(f"globus ls {ep_id}:/", assert_exit_code=4)
```
kurtmckee, line 37: Do we have documentation for what the error codes mean? If not, we need both the documentation and probably constants within the codebase.

kurtmckee (approved):

tests/functional/exception_handling/test_session_hooks.py
sirosen: Good call that we should convert these to constants. It is documented -- specifically it goes into our generated web doc ([code ref](https://github.com/globus/globus-cli/blob/36a60433b2f00de4989b947ed0c885e5a87fdc7e/reference/_generate.py#L41-L42)). I'm not sure if it's discoverable via `--help` though...

PR #1135, opened 2025-07-29
Create a 'gcs collection role show' command

sirosen:
- Add a GCS 'collection role' subcommand group
- Add a collection role show command
- Add a test for 'globus gcs collection role show'


ada-globus (approved):
Reviewed and ran locally. Looks good! Thought that it was interesting that supplying a non-existent role appears to result in a 403—but that's not germane to this change.

PR #1150, opened 2025-08-04
Add a script for confirming min python is tested

sirosen:
Based on the version of this script in `globus-sdk`, this change is a
preparatory move for removal of Python 3.8 support. When we do this,
we want to make sure we keep the minimal version of Python
well-aligned in the repo.

Use `mddj` to read the minimum version of Python from package
metadata and the minimum listed version in tox environments.
Use YAML parsing to pull the relevant data from GitHub workflows.

Add a tox environment + a CI check which runs it to call the script.

In dependency-groups data, `yaml` is now a group so that we can keep a
single pinned version for both testsuite usage and the new
`check-project-metadata` group. This group now also supplies the
dependencies for the `build`+`twine check` tox environment.

### DIFF
--- a/scripts/ensure_min_python_is_tested.py
+++ b/scripts/ensure_min_python_is_tested.py
@@ -8,38 +8,36 @@ import ruamel.yaml
 YAML = ruamel.yaml.YAML(typ="safe")
 REPO_ROOT = pathlib.Path(__file__).parent.parent
 
-proc = subprocess.run(
+requires_python_version = subprocess.check_output(
     ["python", "-m", "mddj", "read", "requires-python", "--lower-bound"],
-    check=True,
-    capture_output=True,
     cwd=REPO_ROOT,
-)
-requires_python_version = proc.stdout.decode().strip()
+    text=True,
+).strip()
 
-with open(REPO_ROOT / ".github" / "workflows" / "test.yaml") as f:
+with open(REPO_ROOT / ".github/workflows/test.yaml") as f:
     workflow = YAML.load(f)
-    includes = workflow["jobs"]["test"]["strategy"]["matrix"]["include"]
-    for include in includes:
-        if include["name"] == "Linux":
-            break
-    else:
-        raise ValueError("Could not find 'Linux' in the test matrix.")
+includes = workflow["jobs"]["test"]["strategy"]["matrix"]["include"]
+for include in includes:
+    if include["name"] == "Linux":
+        break
+else:
+    raise ValueError("Could not find 'Linux' in the test matrix.")
 
-    for environment in include["tox-post-environments"]:
-        if environment.endswith("-mindeps"):
-            break
-    else:
-        raise ValueError("Could not find a '-mindeps' tox-post-environment.")
+for environment in include["tox-post-environments"]:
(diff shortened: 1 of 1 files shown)


scripts/ensure_min_python_is_tested.py
```diff
@@ -0,0 +1,60 @@
...
+YAML = ruamel.yaml.YAML(typ="safe")
+REPO_ROOT = pathlib.Path(__file__).parent.parent
+
+proc = subprocess.run(
+    ["python", "-m", "mddj", "read", "requires-python", "--lower-bound"],
+    check=True,
+    capture_output=True,
+    cwd=REPO_ROOT,
+)
+requires_python_version = proc.stdout.decode().strip()
```
kurtmckee, line 17: This looks like an ideal place for `subprocess.check_output()`, which gets us a decoded STDOUT. I think the code below is equivalent.

```suggestion
requires_python_version = subprocess.check_output(
    ["python", "-m", "mddj", "read", "requires-python", "--lower-bound"],
    cwd=REPO_ROOT,
    text=True,
).strip()
```

scripts/ensure_min_python_is_tested.py
```diff
@@ -0,0 +1,60 @@
...
+
+    python_version, _, _ = environment.partition("-")
+    if python_version != f"py{requires_python_version}":
+        print("ERROR: ensure_min_python_is_tested.py failed!")
+        print(
+            f"\nPackage data sets 'Requires-Python: >={requires_python_version}', "
+            f"but the test-mindeps job is configured to test '{python_version}'.\n",
+            file=sys.stderr,
+        )
+        sys.exit(1)
```
kurtmckee, line 42: Suggesting using in-string slashes, which `pathlib.Path` will handle and seems more readable to me; also suggesting dedenting almost all of the code so that the file is closed immediately.

```suggestion
with open(REPO_ROOT / ".github/workflows/test.yaml") as f:
    workflow = YAML.load(f)
includes = workflow["jobs"]["test"]["strategy"]["matrix"]["include"]
for include in includes:
    if include["name"] == "Linux":
        break
else:
    raise ValueError("Could not find 'Linux' in the test matrix.")

for environment in include["tox-post-environments"]:
    if environment.endswith("-mindeps"):
        break
else:
    raise ValueError("Could not find a '-mindeps' tox-post-environment.")

python_version, _, _ = environment.partition("-")
if python_version != f"py{requires_python_version}":
    print("ERROR: ensure_min_python_is_tested.py failed!")
    print(
        f"\nPackage data sets 'Requires-Python: >={requires_python_version}', "
        f"but the test-mindeps job is configured to test '{python_version}'.\n",
        file=sys.stderr,
    )
    sys.exit(1)
```

kurtmckee (changes requested):

scripts/ensure_min_python_is_tested.py
sirosen: Yep!

I'm fully onboard with the whitespace change and feel somewhat indifferent about the slashes. I guess it's nicer, but hopefully I never need to care about the difference! :grin:

sirosen:
I'll work faster, got it boss! :saluting_face:

PR #1170, opened 2025-09-12
Refactor the `--format=UNIX` printer for type checking and test coverage

sirosen:
Rewrite and add tests for the 'UNIX' formatter.

This changeset carefully makes almost no semantic changes to how the format behaves.
The one exception is that when a dict is being emitted, missing keys were rendered as the empty string. Instead, we're now rendering missing keys as `None` (which, yes, does conflate _missing_ with `null`), which is a bit more consistent with APIs which pad with `null` when fields are absent.
The behavior is encoded in one of the new tests (`test_dict_with_missing_values_treated_as_null`).

---

This change doesn't make the new code absolutely pristine WRT internal docs, but it is a lot better than what came before. At the cost of being a little bit more verbose, we now have logic which is a bit more self-describing in terms of *what* it does.
Some behaviors are notable as having unclear rationale, even after this refactor.


src/globus_cli/termio/printers/unix_printer/_formatter.py
```diff
@@ -0,0 +1,226 @@
...
+    for item in elements:
+        if isinstance(item, list):
+            nested.append(item)
+        else:
+            scalars.append(item)
+
+    if scalars:
+        yield from emit_scalar_list(scalars)
+    for nested_list in nested:
+        yield from emit_list(nested_list, identifier=identifier)
```
derek-globus, line 122: Do we have  a use case today for lists of lists?

This and list of dicts raise flags as particularly gnarly auto-unix-formatting.
Lists of dicts seems like a pretty common pattern for service list apis, but I don't know of if/where we need to handle lists of lists.

sirosen: We actually do have use cases! At first, I was going to say that I wanted to keep this to match `main`, even though APIs don't produce such data.

But the targets for `--format=UNIX` aren't just raw API outputs, but also the results of `jmespath` queries.
Here's a command which produces `list[list]` output, and formats it nicely:
```
globus search index list \
  --jmespath "index_list[].[display_name, join(',', permissions)]" \
  -Funix
```
Sample output:
```
do_searches	owner,writer
example_cookery	owner
Globus Tutorials	writer
gladier-demo	writer
foo	owner
```


I don't know if I consider the _mixed_ case here legitimate, in that this output is nice because I made sure that all of the elements are scalars. If I make the jmespath query `--jmespath "index_list[].[display_name, permissions]"`, dropping the `join()`, then we get...
```
do_searches
writer	owner
example_cookery
owner
Globus Tutorials
writer
gladier-demo
writer
foo
owner
```

which IMO is pretty useless.

derek-globus (approved):

PR #1173, opened 2025-09-17
Do not pass 'skip_activation_check' to Transfer API

sirosen:
This flag has been deprecated in both the SDK and CLI, and has no longer
has any effect when passed.

By setting it, we now trigger an SDK deprecation warning. Remove the
setting but still accept the flag (which is in its CLI deprecation
phase).


kurtmckee:
Hmm...I didn't realize it was deprecated in the CLI already. I think I see that it was deprecated almost a full year ago.

So! LGTM!

sirosen:
Sometime soon, let's take it from deprecated to removed, and that can have a changelog. :wink:

PR #1182, opened 2025-09-28
[SC 38762] Adding collection role create

jkube-globus:
Adding in `globus collection role create` and corresponding test in support of [SC-38762](https://app.shortcut.com/globus/story/38762/cli-add-support-for-creating-a-role-on-a-collection)

**Updated:**
- src/globus_cli/commands/collection/role/init.py
- tests/files/api_fixtures/collection_operations.yaml
- tests/functional/collection/role/test_role_delete.py
- - Correcting `globus collection role delete` test 
- tests/functional/collection/role/test_role_list.py
- - Correcting `globus collection role list` test

**Added:**
- src/globus_cli/commands/collection/role/create.py
- tests/functional/collection/role/test_role_create.py
- changelog.d/20250927_194304_james_SC_38762_Collection_Role_Create.md


tests/functional/collection/role/test_role_create.py
```diff
@@ -0,0 +1,50 @@
...
+                    "principal": f"urn:globus:auth:identity:{user_id}",
+                    "role": "activity_monitor",
+                }
+            ],
+            "detail": "success",
+            "has_next_page": False,
+            "http_response_code": 200,
+            "message": f"Created new role {role_id}",
+        },
+    ).add()
```
derek-globus: Why is this response necessary here?

You defined the same response mock inside of the `collection_operations.yaml` above which is being loaded with the call to `load_response_set("cli.collection_operations")`.

Was it not working without this additional in-python response addition?

tests/functional/collection/role/test_role_delete.py
```diff
@@ -30,7 +30,7 @@ def test_successful_gcs_collection_role_delete(
 
     # now test the command and confirm that a successful role deletion is reported
     run_line(
-        ["globus", "gcs", "collection", "role", "delete", collection_id, role_id],
+        ["globus", "collection", "role", "delete", collection_id, role_id],
```
derek-globus: Could your update the invocation prefix here & in the other test files to `["globus", "gcs", "collection", ...]

There's a bit of contexted, but those are namespaced intentionally.

---

Our ultimate vision for the cli is to move specific resource and action keys (i.e., "collection") out of the top-level namespace by always nesting them under their service name. That itself will definitely be a breaking change however, so initially we started by introducing `gcs` as an equivalent command to the existing ones ([in this PR](https://github.com/globus/globus-cli/pull/902)).

Given that direction, `globus gcs collection ...` is actually our primary invocation path, not `globus collection ...` so we should keep our tests using that.

src/globus_cli/commands/collection/role/create.py
```diff
@@ -0,0 +1,85 @@
...
+
+@command("create")
+@collection_id_arg
+@click.argument("ROLE", type=click.Choice(t.get_args(_VALID_ROLES)), metavar="ROLE")
+@click.argument("PRINCIPAL", type=str, required=False)
+@click.option(
+    "--principal-type",
+    type=click.Choice(["identity", "group"]),
+    help="Qualifier to specify what type of principal (identity or group) is provided.",
+)
```
sirosen: > [!NOTE]
> For any other reviewers, this `--principal-type` phrasing is already in use elsewhere in the CLI, and I think we should apply it here (as currently in the PR) for uniformity.
>
> That said, I'm very much interested in developing a better interface for passing these inputs. I think we could design something as a replacement which takes URNs, short-forms like `group:<id>`, and usernames.

PR #1185, opened 2025-10-01
Added flow timer creation

derek-globus:
## Additions
* Added flow timer creation under the command `globus timer create flow`
  * To support the per-flow consent required by this, added a new flag `globus login --timer flow:<flow_id>` which flow timer creation instructs to use.

## Note
This code was built against https://github.com/globus/globus-sdk-python/pull/1311 and thus won't succeed build/testing with the globus-sdk main branch.

Merge will need to wait until that merges/releases.

## Demo Usage

<details>
<summary>Successful Timer Creation</summary>

image

</details>

<details>
<summary>Rejected Timer Creation</summary>
image

</details>

### DIFF
--- a/src/globus_cli/commands/timer/create/flow.py
+++ b/src/globus_cli/commands/timer/create/flow.py
@@ -1,14 +1,12 @@
 from __future__ import annotations
 
-import datetime
-import textwrap
 import uuid
+from datetime import datetime
 
 import click
 import globus_sdk
-from globus_sdk.scopes import Scope, SpecificFlowScopeBuilder, TimersScopes
 
-from globus_cli.login_manager import LoginManager, is_client_login
+from globus_cli.login_manager import LoginManager
 from globus_cli.parsing import (
     ParsedJSONData,
     command,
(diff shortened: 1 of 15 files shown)


src/globus_cli/commands/timer/create/flow.py
```diff
@@ -0,0 +1,119 @@
...
+            textwrap.dedent(
+                f"""
+                Warning: Flow timer creation requires additional consent.
+                Please resolve with the following command, then rerun your original one:
+
+                    globus login --timer flow:{flow_id}
+                """
+            )
+        )
+        click.get_current_context().exit(4)
```
MaxTueckeGlobus, line 119: Small knit (that maybe you disagree with), looking at other `get_<service>_client` methods in `LoginManager`, it seems that this consent verification should belong in `get_timers_client` rather than in the actual command (see examples like `get_specific_flow_client` or `get_gcs_client`).

It's functionally identical, but for consistency I wonder if it should be moved.

MaxTueckeGlobus: Additionally, it looks like `LoginManager` has methods for checking that the consent forest meets scope requirements and the subsequent error messaging in the form of `LoginContext` and `assert_logins` (see `get_specific_flow_client` [here](https://github.com/globus/globus-cli/blob/1aaf3aeaa9a941528d171f7ac38f75b4abdd7250/src/globus_cli/login_manager/manager.py#L411-L417) for reference) that I think should be used here.

kurtmckee:
@derek-globus If you pull in SDK v3.65.0 this should be able to pass CI.

MaxTueckeGlobus (changes requested):

src/globus_cli/commands/timer/create/flow.py
```diff
@@ -0,0 +1,119 @@
...
+    flow_input_document_option,
+)
+from globus_cli.termio import display
+
+from ._common import CREATE_FORMAT_FIELDS, TimerSchedule, timer_schedule_options
+
+
+@command("flow", short_help="Create a recurring flow timer.")
+@flow_id_arg
+@flow_input_document_option
```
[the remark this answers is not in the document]
MaxTueckeGlobus: Works for me

MaxTueckeGlobus (approved):

PR #1190, opened 2025-10-07
Fix a boolen evaluation in timer-flow login

derek-globus:
Fix a small bug I introduced in pr-1185 that would cause flow timer creation to be impossible.


sirosen (approved):
It would be nice to capture this with a test as well, but IIRC the login tests are particularly hard to work with.

derek-globus:
> It would be nice to capture this with a test as well, but IIRC the login tests are particularly hard to work with.

Yeah, you need to properly mock consents which we can do, but takes a while to do well.

PR #1196, opened 2025-10-17
Unify flows command field rendering

derek-globus:
## What?
* Unify the format field set used when displaying a flow e.g., with `globus flow show ...` or `globus flow create ...`
* Unify the format field set used when displaying a flow run e.g., with `globus flow start ...` or `globus flow run show ...`
* Unify the identity-patching fixtures used to test flow/flow run rendering.

## Why?
* Standardize our cli output for CRUD-like operations in flows to provide a more cohesive and predictable user experience.
* Prep the code to make [sc-45516](https://app.shortcut.com/globus/story/45516) less expansive.

## Review note
* There's a lot of LoC but it's 75% deduplication of things we'd written down 5 or 6 times. To make the review process simpler, I bucketed the 3 changes into individual commits which I'd recommend reading sequentially:
1. [Define flow api fields centrally](https://github.com/globus/globus-cli/commit/cf75db9fb4cfb6d575422adaf22148bf00544386)
1. [Define flow run api fields centrally](https://github.com/globus/globus-cli/commit/d31520d4cf97de68e80220d4f3728f1b78d8c031)
1. [Refactor and unify flows functional tests](https://github.com/globus/globus-cli/commit/35f24499e43afc4018a6d3f048afc441c3776667)

### DIFF
--- a/src/globus_cli/commands/flows/_fields.py
+++ b/src/globus_cli/commands/flows/_fields.py
@@ -13,11 +13,11 @@ class FlowPrincipalFormatter(PrincipalURNFormatter):
         self, auth_client: globus_sdk.AuthClient, flow: dict[str, t.Any]
     ) -> None:
         super().__init__(auth_client)
-        self.add_item(flow.get("flow_owner"))
-        [self.add_item(admin) for admin in flow.get("flow_administrators", ())]
-        [self.add_item(viewer) for viewer in flow.get("flow_viewers", ())]
-        [self.add_item(manager) for manager in flow.get("run_managers", ())]
-        [self.add_item(monitor) for monitor in flow.get("run_monitors", ())]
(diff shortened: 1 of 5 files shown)


src/globus_cli/commands/flows/_fields.py
```diff
@@ -0,0 +1,110 @@
...
+    )
+    csv_list = formatters.ArrayFormatter(delimiter=", ")
+
+    flow_description_fields = (
+        [
+            Field("Flow Subtitle", "flow_description.subtitle"),
+            Field("Flow Description", "flow_description.description"),
+            Field("Flow Keywords", "flow_description.keywords", formatter=csv_list),
+        ]
+        if "flow_description" in flow_run
```
m1yag1: I like this decision to remove the conditional for `include_flow_description` and instead look at the response.

PR #1198, opened 2025-10-23
Added globus session update --scope

derek-globus:
## What?
* Add a new optional parameter `--scope` to `globus session update` which defines additional scopes to request when running the user through a login flow.
* Update `globus flows start <flow_id>` to always include the flow scope in remediating resolution `globus session update` command.

## Why?

In the current cli, `globus flows start <flow_id>` recommends using `globus session update` with exclusively the attached policy id from the api-returned GARE. This command doesn't function as expected however because of how flows operate as individual resource servers.

Inserting the scope can allow remediation login commands recommendations to ensure that any a user login will return scoped tokens not just for statically known resource servers, but also any dynamically discovered ones.

## Testing
image

### DIFF
--- a/tests/functional/flows/test_update_flow.py
+++ b/tests/functional/flows/test_update_flow.py
@@ -65,8 +65,6 @@ def test_update_flow_text_output(run_line, load_identities_for_flow):
         "Administrators",
         "Starters",
         "Viewers",
-        "High Assurance",
-        "Authentication Policy ID",
         "Run Managers",
         "Run Monitors",
     }


src/globus_cli/commands/flows/start.py
```diff
@@ -250,16 +253,46 @@ def start_command(
...
+        yield
+    except globus_sdk.GlobusAPIError as e:
+        gare = to_gare(e)
+        if gare:
+            scopes = gare.authorization_parameters.required_scopes or []
+            if scope not in scopes:
+                scopes.append(scope)
+
+            gare.authorization_parameters.required_scopes = scopes
+            raise CLIAuthRequirementsError("", gare=gare)
```
kurtmckee: Because this is being caught and transformed, I think this should be a `raise ... from None`. Thoughts?

```suggestion
            raise CLIAuthRequirementsError("", gare=gare) from None
```

derek-globus:
Fyi, while testing this change more locally, I discovered that I'd accidentally broken non-scoped `session updating` at initial time of PR publication. I've now added https://github.com/globus/globus-cli/pull/1198/commits/784b7fe64e33e21929e1cecc4eebce886ac1ab8b and https://github.com/globus/globus-cli/pull/1198/commits/25e82667aa662dcbf4d2f6e0ce4b0c9f25f5fae9 to introduce a bit more nuance and ensure that non-scoped session updates properly apply to all statically known resource servers.

PR #1222, opened 2025-12-23
Improve test coverage by adding tests and removing dead code, bump fail-under to 92%

sirosen:
Examining coverage reports for the CLI shows that there are several large untested blobs.
A case-by-case evaluation reveals that several of these can be improved dramatically with minimal effort.

1. Remove dead code (unused helpers, etc)
2. Exclude the local flake8 plugin from unit tests
3. Unit test `--map-http-status`
4. Unit test `IdentityType`
5. Unit test `globus version`

Only this last item (`globus version`) required any meaningful change in the src tree, as it uncovered a deprecation warning when `globus version -v` is run.
The fix is simply an update to use `importlib.metadata` to retrieve package version info.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -113,7 +113,7 @@ source = [
 [tool.coverage.report]
 show_missing = true
 skip_covered = true
-fail_under = 90
+fail_under = 92
 exclude_lines =[
     # the pragma to disable coverage
     "pragma: no cover",


pyproject.toml
```diff
@@ -102,6 +102,9 @@ filterwarnings = [
 parallel = true
 patch = ["subprocess"]
 source = ["globus_cli"]
+omit = [
+  "globus_cli_flake8.py",
```
kurtmckee: I have a feeling this shouldn't be in the `src/` tree to begin with, but this is currently a helpful omission!

PR #1224, opened 2025-12-29
Refresh the tox environment list and factorization

sirosen:
The overall goal here is to have clearer and more maintainable tox config, without accidentally inheriting things from the default testenv.
Base envs are also a good way to share config explicitly across envs.

- Add descriptions to all environments.
- Split the pytest envs off from the default, and make the default fail with a message.
- Use env inheritance to simplify the pytest env factors.
- Split up the metadata check environments.
- Update the default env_list.

Incidentally, add py3.14 to our tox envs (but not added yet in CI, to reduce scope creep).

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -2,7 +2,8 @@
 [tox]
 envlist =
     check-min-python-is-tested
-    check-package-data
+    twine-check{,-minimum}
+    check-sdist{,-minimum}
     reference
     cov-clean
     py{3.14,3.13,3.12,3.11,3.10,3.9}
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -10,97 +15,163 @@ minversion = 4.22.0
...
 # the tools in `check-project-metadata` require Python 3.10+
 # when we drop 3.9 support, this will become correct
-base_python = python3.10
-recreate = true
-skip_install = true
-dependency_groups = check-project-metadata
+base_python =
+    minimum: python3.10
+    !minimum: python
+# these checks build without build isolation, so install the build backend explicitly
```
kurtmckee: This phrasing appears to suggest a causal relationship that isn't intended.

> _Because_ build isolation is disabled, we _must therefore_ install the build backend ourselves.

Rather, the goal is to test that the package can be built on the minimum Python platform with the minimum version of the build backend, and `build` has no way to control for that, so we've wrested dependency management from `build` and given it to `tox`.

I think that the comment could be removed, or could be moved to the `commands` point where `--no-build-isolation` is passed.

kurtmckee: Yeah, looks good to me!

PR #1227, opened 2025-12-29
Add lines to make tox output grouped in GitHub

sirosen:
Group the outputs so that they are easier to navigate.


kurtmckee:
I think that this may conflict with an upcoming update to globus/workflows, which will add the tox-gh plugin.

kurtmckee:
I recommend closing in favor of a forthcoming bump to the shared workflow.

PR #1231, opened 2025-12-31
Implement "folded table" output, as an opt-in feature for all tabular outputs

sirosen:
This could be viewed as an alternative implementation / reimplementation of our table output printer.

Similar to the current table output, it loads the data to print into a data structure which represents our desired output.
*Unlike* the current table output, each row is treated as a meaningful data structure as well.
Instead of a grid of cells for the whole table, a table is a list of rows, each of which contains a grid.

Rows and tables are *not* responsible for correctness invariants like "every row has the same layout" -- the enforcement of such rules is left solely to the higher-level printer.
Rows define "folding", which is an accordion-like fold that converts from

    a b c d e f

to

    a c e
    b d f

or

    a d
    b e
    c f

and so forth for different fold sizes.

Rows *do not* support folding when they are already folded -- only a 1-height row can be folded.
As such, Rows and Tables (lists of Rows) are immutable, and the folding methods produce net-new rows and new tables.
The printer handles them as such, and can therefore compare a "fold by 2" table with a "fold by 3" table.

With the "folding table" implemented and checked against terminal width to determine how much/if it should fold, there is the matter of output formatting for the rows.
Each row knows how to format itself, and accepts a set of "style" flags (an enum) to control specifics.
The same "style" flags can control the printing of separator lines for the header and foot of the table.

Finally, there is the matter of control.
When the environment is not interactive (e.g., a pipeline), folding is always disabled.
The table folding output is off by default, but can be enabled as a replacement for all table printing by setting the new env var `GLOBUS_CLI_FOLD_TABLES=1`.
The intent, in this PR, is that the new folded table would fully *replace* our table output in the future, and users could opt out of the behavior with `GLOBUS_CLI_FOLD_TABLES=0`.

When a table is not folded, the output is *almost* identical to our current non-folded table. This is intentional.

### DIFF
--- a/src/globus_cli/termio/printers/folded_table_printer.py
+++ b/src/globus_cli/termio/printers/folded_table_printer.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import collections
+import dataclasses
 import enum
 import functools
 import shutil
@@ -27,6 +28,36 @@ class SeparatorRowType(enum.Enum):
     box_intra_row_separator = enum.auto()
     # bottom of a table
     box_bottom = enum.auto()
+
+
+@dataclasses.dataclass
+class SeparatorRowStyle:
+    fill: str
+    leader: str
+    trailer: str
+    middle: str
+
+
+SEPARATOR_ROW_STYLE_CHART: dict[SeparatorRowType, SeparatorRowStyle] = {
+    SeparatorRowType.ascii_header_separator: SeparatorRowStyle(
+        fill="-", leader="", trailer="", middle="+"
+    ),
+    SeparatorRowType.box_top: SeparatorRowStyle(
+        fill="═", leader="╒═", trailer="═╕", middle="╤"
+    ),
+    SeparatorRowType.box_header_separator: SeparatorRowStyle(
+        fill="═", leader="╞═", trailer="═╡", middle="╪"
+    ),
+    SeparatorRowType.box_row_separator: SeparatorRowStyle(
+        fill="─", leader="├─", trailer="─┤", middle="┼"
+    ),
+    SeparatorRowType.box_intra_row_separator: SeparatorRowStyle(
+        fill="─", leader="├─", trailer="─┤", middle="┼"
+    ),
+    SeparatorRowType.box_bottom: SeparatorRowStyle(
+        fill="─", leader="└─", trailer="─┘", middle="┴"
+    ),
+}
 
 
 class FoldedTablePrinter(Printer[t.Iterable[t.Any]]):
@@ -273,31 +304,7 @@ def _format_subrow(
 
 @functools.cache
 def _separator_line(col_widths: tuple[int, ...], row_type: SeparatorRowType) -> str:
-    if row_type is SeparatorRowType.ascii_header_separator:
-        fill = "-"
-        leader = ""
-        trailer = ""
-        middle_decorator = "+"
-    elif row_type is SeparatorRowType.box_top:
-        fill = "═"
-        leader = "╒═"
-        trailer = "═╕"
-        middle_decorator = "╤"
-    elif row_type is SeparatorRowType.box_header_separator:
-        fill = "═"
-        leader = "╞═"
-        trailer = "═╡"
-        middle_decorator = "╪"
-    elif row_type is SeparatorRowType.box_bottom:
-        fill = "─"
-        leader = "└─"
-        trailer = "─┘"
-        middle_decorator = "┴"
-    else:
-        fill = "─"
-        leader = "├─"
-        trailer = "─┤"
-        middle_decorator = "┼"
+    style = SEPARATOR_ROW_STYLE_CHART[row_type]
 
     # in intra-row separator lines, they are drawn as dashed box char lines
     if row_type is SeparatorRowType.box_intra_row_separator:
@@ -306,14 +313,14 @@ def _separator_line(col_widths: tuple[int, ...], row_type: SeparatorRowType) -> 
     else:
 
         def fill_column(width: int) -> str:
-            return width * fill
+            return width * style.fill
 
-    line_parts = [leader]
+    line_parts = [style.leader]
     for col in col_widths[:-1]:
         line_parts.append(fill_column(col))
-        line_parts.append(f"{fill}{middle_decorator}{fill}")
+        line_parts.append(f"{style.fill}{style.middle}{style.fill}")
     line_parts.append(fill_column(col_widths[-1]))
-    line_parts.append(trailer)
+    line_parts.append(style.trailer)
     return "".join(line_parts)
 
 


sirosen:
I've made several adjustments after a team review of the high-level change:

- The behavior is now opt-out, not opt-in, via the env var
- Rows have _internal separator lines_ presented as dashed lines
- I've reworked it to use box drawing chars (I struggled with this at first, but it's now controlled via the "output style" enum)
- Columns inside the table are now drawn with the dashed vertical separator

One of the concerns raised was that we want each entity (`Row`) to be indicated as both "one thing" and "separate fields". We experimented with extra whitespace and other patterns for showing that vertically joined fields were combined. We also thought about potentially wrapping some fields in the future.
This drove the addition of the dashed internal separator lines. In concert with the dashed verticals, each element inside the table is fully connected by whitespace and separated from the next element by a solid horizontal line.

I'm now marking this ready for review.

src/globus_cli/termio/printers/folded_table_printer.py
```diff
@@ -0,0 +1,333 @@
...
+            before_decorator = "╞"
+            after_decorator = "╡"
+            middle_decorator = "╪"
+
+        leader = f"{before_decorator}{fill}"
+        trailer = f"{fill}{after_decorator}"
+    else:
+        leader = ""
+        trailer = ""
+        middle_decorator = "+"
```
derek-globus: Could all of this content live in inside of the OutputStyle enum?

Something like
```
class OutputStyle(Enum):
    none = (" ", " ", " ", " ")
    decorated = (" ", "*", "*", "*")
    double = ("=", "=", "=", "=")
    double_transition = ("=", "/", "\\", "|")
    top = ("-", "-", "-", "-")
    bottom = ("_", "_", "_", "_")

    def __init__(self, fill: str, before: str, after: str, middle: str):
        self.fill = fill
        self.before_decorator = before
        self.after_decorator = after
        self.middle_decorator = middle
```

except with the real characters you're inserting here (I just inserted random ones for the sake of demonstrating shape).

---

That'd help define what each style is declaratively without having the reader walk through this conditional path and rebuild it for themselves imperatively.

src/globus_cli/termio/printers/folded_table_printer.py
```diff
@@ -0,0 +1,333 @@
...
+        table = self._fold_table(RowTable.from_data(self._fields, data))
+        col_widths = table.calculate_column_widths()
+
+        # if folded, print a leading separator line
+        table_style = OutputStyle.decorated if table.folded else OutputStyle.none
+
+        if table.folded:
+            echo(
+                _separator_line(
+                    col_widths, style=table_style | OutputStyle.double | OutputStyle.top
```
derek-globus: It doesn't look like this union will ever fall through beyond the first element.

`table_style` is set to either `OutputStyle.decorated` or `OutputStyle.none` both of which are truth-y.

src/globus_cli/termio/printers/folded_table_printer.py
```diff
@@ -0,0 +1,333 @@
...
+from .base import Printer
+
+
+class OutputStyle(enum.Flag):
+    none = enum.auto()
+    decorated = enum.auto()
+    double = enum.auto()
+    double_transition = enum.auto()
+    top = enum.auto()
+    bottom = enum.auto()
```
derek-globus: This being represented as a boolean flag makes it pretty difficult to understand.

From a lot of back and forth, I think I get the idea is that:
* `none` and `decorated` disambiguate broadly "how am I styling this"
* `top` and `bottom` delineate (for specifically separator lines) whether to modify the corners
* `double` and `double_transition` delineate stylization characters within a separator line.

Assuming I got that right, it took a lot of imperative piecing together to figure out.
Maybe some comments would help a reader grok this quicker but I think it's just not a great way to model this. The approach merges different abstraction levels (e.g., "decorated" and "bottom") into the same "bag" of style-elements, requiring the reader to look at all of the usage sites to understand anything about what they are.

---

My recommendation would be to model the "style of output" (e.g., `none` or `decorated`) distinctly from the "category of thing being output".

Something like the following would convey the same information more explicitly in method signatures with informative types of stylization than the permutation bag of styles offers.

```python
def _separator_line(
  col_widths: tuple[int, ...],
  style: t.Literal["none", "decorator"],
  fill_type: t.Literal["single", "double"],
  line_type: t.Literal["top", "bottom", "transition"]
) -> str

...

class Row:
   ...
   def serialize(
      use_col_widths: tuple[int, ...],
      style: t.Literal["none", "decorator"]
   ) -> str
      ...
```

Example Note:
I used literal types just because that helped me include the data type in the signature; for a real implementation those could be modeled as enums or dataclasses or whatever. The important part I'm trying to highlight is that different "types" of stylization are identified concretely in the signature instead of mushed together into style flags.

sirosen: I'm starting to tinker on this now. My first step is going to be to try removing `decorated` as a style -- using the `is_folded` bool more -- and see how that looks. After that, I think I will want to separate "style" from "line type" (which looks to me like the right split).

I'm sticking with enums, but as you say that's a detail. :crossed_fingers: it goes smoothly, but I agree with the overall feedback. Too many ideas got mushed into one here.

src/globus_cli/termio/printers/folded_table_printer.py
```diff
@@ -0,0 +1,347 @@
...
+    SeparatorRowType.box_row_separator: SeparatorRowStyle(
+        fill="─", leader="├─", trailer="─┤", middle="┼"
+    ),
+    SeparatorRowType.box_intra_row_separator: SeparatorRowStyle(
+        fill="─", leader="├─", trailer="─┤", middle="┼"
+    ),
+    SeparatorRowType.box_bottom: SeparatorRowStyle(
+        fill="─", leader="└─", trailer="─┘", middle="┴"
+    ),
+}
```
derek-globus: Did you try the charachter-data-in-enum version?

This version is certainly legible enough that I wouldn't block approval over this style. But curious to hear your thought process as to why you prefer the multi-data style.

```python
class SeparatorRowType(enum.Enum):
    # top of a table
    box_top = ("═", "╒═", "═╕", "╤")
    # between the header and the rest of the table (box drawing chars)
    box_header_separator = ("═", "╞═", "═╡", "╪")
    # the same, but for ASCII tables
    ascii_header_separator = ("-", "", "", "+")
    # between rows of the table
    box_row_separator = ("─", "├─", "─┤", "┼")
    # between element lines inside of a row of a table
    box_intra_row_separator = ("─", "├─", "─┤", "┼")
    # bottom of a table
    box_bottom = ("─", "└─", "─┘", "┴")

    def __init__(fill: str, leader: str, trailer: str, middle: str):
        self.fill = fill
        self.leader = leader
        self.trailer = trailer
        self.middle = middle
```

sirosen: Type checking was fine but I'm scrapping it because it is, as past experience has shown, weird as heck.

```pycon
>>> import enum
>>> class FooEnum(enum.Enum):
...     a = (1, 2)
...     b = (2, 3)
...     c = (2, 3)
...     
>>> FooEnum.b
<FooEnum.b: (2, 3)>
>>> FooEnum.c
<FooEnum.b: (2, 3)>
>>> FooEnum.c is FooEnum.b
True
```

:skull_and_crossbones:

PR #1244, opened 2026-02-01
Reduce Dependabot toil; configure cooldowns

kurtmckee:
This PR introduces the following changes:

* Configure Dependabot to run once every six months
* Configure Dependabot to use a 20-day cooldown period

<sub>This PR was generated using [turbolift](https://github.com/Skyscanner/turbolift).</sub>

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -4,10 +4,14 @@ updates:
     directory: "/"
     schedule:
       interval: "semiannually"
+    cooldown:
+      default-days: 20
   - package-ecosystem: "github-actions"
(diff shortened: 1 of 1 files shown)


sirosen (approved):
I've been meaning to roll out cooldowns.

Quick related question:
If `check-jsonschema` offered a hook which required `cooldown.default-days` to be set on all dependabot configs, would you use it? (Considering writing such a thing, since it's just a small schema + a run config for it.)

kurtmckee:
I wouldn't use it. That seems more appropriate for a repo-review-style check, though.

PR #1257, opened 2026-05-08
Added globus flows registered-api show command

m1yag1:
## WHAT

Add `globus flows registered-api show` command 

## RELATED

> This will fail tests until a release of the globus-sdk includes the change below.

- https://github.com/globus/globus-sdk-python/pull/1386

### DIFF
--- a/src/globus_cli/commands/flows/registered_api/show.py
+++ b/src/globus_cli/commands/flows/registered_api/show.py
@@ -16,9 +16,10 @@ from globus_cli.termio import display
 def show_command(login_manager: LoginManager, *, registered_api_id: str) -> None:
     """
     Show a registered API.
+
+    Accepts a registered API UUID.
     """
     flows_client = login_manager.get_flows_client()
-    auth_client = login_manager.get_auth_client()
 
     # Convert string to UUID if it's a valid UUID
     try:
@@ -28,6 +29,7 @@ def show_command(login_manager: LoginManager, *, registered_api_id: str) -> None
 
     res = flows_client.get_registered_api(api_id)
 
+    auth_client = login_manager.get_auth_client()
     fields = registered_api_format_fields(auth_client, res.data)
 
     display(res, fields=fields, text_mode=display.RECORD)
--- a/src/globus_cli/commands/flows/registered_api/show.py
+++ b/src/globus_cli/commands/flows/registered_api/show.py
@@ -1,0 +1,33 @@
+from __future__ import annotations
+
+import uuid
+
+import click
+
+from globus_cli.commands.flows._fields import registered_api_format_fields
+from globus_cli.login_manager import LoginManager
+from globus_cli.parsing import command
+from globus_cli.termio import display
+
+
+@command("show")
+@click.argument("registered_api_id")
+@LoginManager.requires_login("auth", "flows")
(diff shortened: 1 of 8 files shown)


src/globus_cli/commands/flows/registered_api/show.py
```diff
@@ -0,0 +1,33 @@
...
+from globus_cli.parsing import command
+from globus_cli.termio import display
+
+
+@command("show")
+@click.argument("registered_api_id")
+@LoginManager.requires_login("auth", "flows")
+def show_command(login_manager: LoginManager, *, registered_api_id: str) -> None:
+    """
+    Show a registered API.
```
kurtmckee, line 18: I recommend expanding this docstring to clarify the accepted formats of registered API IDs.

For example, "UUIDs, URNs, and (perhaps) partial URNs are accepted". (I anticipate that we might not have to require the full URN including the static prefix `urn:globus:flows:handle:`.)

src/globus_cli/commands/flows/registered_api/show.py
```diff
@@ -0,0 +1,33 @@
...
+    """
+    flows_client = login_manager.get_flows_client()
+    auth_client = login_manager.get_auth_client()
```
kurtmckee: Could `auth_client` be instantiated just above line 31? The instance isn't needed until then, and the call to `flows_client.get_registered_api()` could fail, which would make this instantiation unnecessary.

tests/functional/flows/test_show_registered_api.py
```diff
@@ -0,0 +1,58 @@
...
+    output_value = _get_output_value(field_name, result.output)
+    output_usernames = [x.strip() for x in output_value.split(",")]
+    assert expected_usernames == set(output_usernames)
+
+
+def _get_output_value(name, output):
+    """
+    Return the value for a specified field from the output of a command.
+    """
+    match = re.search(rf"^{name}:[^\S\n\r]+(?P<value>.*)$", output, flags=re.M)
```
kurtmckee: It's probably not necessary in practice, but I recommend using `re.escape(name)` to ensure that `name` cannot introduce regex characters.

tests/functional/flows/test_show_registered_api.py
```diff
@@ -0,0 +1,58 @@
...
+    assert_usernames(result, pool, "Administrators", roles.get("administrators", []))
+    assert_usernames(result, pool, "Viewers", roles.get("viewers", []))
+
+
+def assert_usernames(result, pool, field_name, principals):
+    expected_usernames = {pool.get_username(principal) for principal in principals}
+
+    output_value = _get_output_value(field_name, result.output)
+    output_usernames = [x.strip() for x in output_value.split(",")]
+    assert expected_usernames == set(output_usernames)
```
kurtmckee: As with `expected_usernames`, above, this could be a set comprehension:


```suggestion
    output_usernames = {x.strip() for x in output_value.split(",")}
    assert expected_usernames == output_usernames
```

kurtmckee (changes requested):

tests/functional/flows/test_show_registered_api.py
m1yag1: This has been updated

kurtmckee (approved):

PR #1260, opened 2026-05-21
Support click v8.4

sirosen:
8.4 introduces relatively few changes, but does make `ParamType` a
Generic parametrized over the return type of `convert()`. Unfortunately,
we also support Python 3.9 , which means we must support older click
versions. As such, `ParamType` is not even subscriptable on that version.

To work around this, we use `TYPE_CHECKING` guards to split definitions
between a generic and non-generic variant.

Only one line of runtime code is changed:
A superfluous call to `super().convert()` was flagged by `mypy` for
having the wrong type and is here removed.


src/globus_cli/commands/gcp/set_subscription_id.py
```diff
@@ -10,8 +11,13 @@
 from globus_cli.parsing import command, endpoint_id_arg
 from globus_cli.termio import display
 
+if t.TYPE_CHECKING:
+    _SubscriptionIdTypeBase = click.ParamType["uuid.UUID | ExplicitNullType"]
```
kurtmckee: Why is this one a string of types, rather than the types themselves?

```
    _SubscriptionIdTypeBase = click.ParamType["uuid.UUID | ExplicitNullType"]
```

vs 

```
    _SubscriptionIdTypeBase = click.ParamType[uuid.UUID | ExplicitNullType]
```

This question applies to all of the locations where strings are used like this.

src/globus_cli/commands/api.py
```diff
@@ -37,7 +44,6 @@ def convert(
         param: click.Parameter | None,
         ctx: click.Context | None,
     ) -> tuple[str, str] | None:
-        value = super().convert(value, param, ctx)
```
kurtmckee, line 40: Why is this line now removed?

This question applies to all locations where calls to `super().convert()` are removed.

src/globus_cli/parsing/param_types/endpoint_plus_path.py
```diff
@@ -67,7 +72,9 @@ def convert(
         # split the value on the first colon, leave the rest intact
         splitval = value.split(":", 1)
         # first element is the endpoint_id
-        endpoint_id = click.UUID(splitval[0])
+        # NOTE: for some reason, mypy thinks `click.UUID()` can return `None`, but that
+        #       does not match its annotation
+        endpoint_id: uuid.UUID = click.UUID(splitval[0])  # type: ignore[assignment]
```
kurtmckee: It appears to match [click's documentation](https://click.palletsprojects.com/en/stable/api/#click.UUID), though. 🤷 

I recommend removing this note, and possibly type-narrowing independently.

src/globus_cli/parsing/param_types/notify_param.py
```diff
@@ -1,10 +1,17 @@
...
+
 import click
 from click.shell_completion import CompletionItem
 
 from globus_cli._click_compat import shim_get_metavar
 
+if t.TYPE_CHECKING:
+    _NotificationTypeBase = click.ParamType[dict[str, bool]]
+else:
+    _NotificationTypeBase = click.ParamType
```
kurtmckee: This pattern is repeated in many, many places, and I'm assuming once lower click version support is dropped this will have to be re-updated to remove the bare `click.ParamType` annotation.

Is there any way to move this so all of these conditionals are grouped in one place? I see that there's already a `globus_cli._click_compat` module, and I would dearly love to see all of these conditionals -- which are explicitly required for click 8.4 compatibility -- move to that module, or one like it!

pyproject.toml
```diff
@@ -29,7 +29,7 @@ classifiers = [
 requires-python = ">=3.10"
 dependencies = [
     "globus-sdk==4.6.0",
-    "click>=8.1.4,<8.4",
+    "click>=8.1.4,<8.5",
```
kurtmckee: I'm concerned that we're continuing to find that minor version after minor version of click contains breaking changes, and that we have to increment pins like this.

I'm pretty okay with just pinning click to version 8.4 and dropping all of these conditionals entirely. This is an application, and I think we can treat its dependencies as such.

kurtmckee (commented):
It looks like click is continuing to release breaking changes in minor releases. We're on a treadmill.

I recommend that instead of "supporting click v8.4" this PR change to "require click v8.4". This is an application, and I think we can treat its dependencies as such by pinning them.

We already do this for a number of dependencies; I don't see a reason why click should be an outlier that requires a range of supported versions.

sirosen:
We discussed offline, but to surface in the permanent record:

`click` does not use semver. So it is expected and normal for minor releases to be breaking.
I would say there is no process problem here. On reflection, I may add a comment to our dependency spec about this.

We weren't able to drop old `click` versions when we picked up 8.3 support because `click` had dropped Python 3.9 and at the time we were not prepared to do so. Now that that has been freed up, we can update our lower bound and only support `>= 8.4 , < 8.5`.

src/globus_cli/commands/gcp/set_subscription_id.py
sirosen: When I wrote this, we were testing on 3.9, where `UnionType` did not exist yet.

So the string-ized annotation `"str | int"` was valid, but `str | int` was not. As part of the adjustments I'm making, I'll re-evaluate these.

src/globus_cli/commands/api.py
sirosen: I noted this in the PR description, although with a minor error:
> Only one line of runtime code is changed:
> A superfluous call to `super().convert()` was flagged by `mypy` for
> having the wrong type and is here removed.

In fact there are two such lines, both in this file.

---

When we upgrade to 8.4+ only, the type checking error goes away.
However, the body of `ParamType.convert` is literally:
```
        # The default returns the value as-is so subclasses that only customize
        # metadata are not forced to redeclare ``convert``.
        return t.cast("ParamTypeValue", value)
```

As such, I'd prefer to omit it, barring any objection.

---

Reviewing `ParamType.convert()` made me notice that there's a new bit of doc which says that it's not called when the value is `None`. I'll see if we can adjust to drop all of our `None` handling in these converter functions.

src/globus_cli/parsing/param_types/endpoint_plus_path.py
sirosen: I re-examined everything and figured it out. `__call__` can return `None` if the input is `None`, but `convert()` cannot.
Simply tweaking this line to call `convert()` directly gets the right type.

`click` could define overloads, but I think upstream is fine on this. If I were a maintainer there -- and as a sometimes-contributor -- I wouldn't want a bunch of overloads just for such niche usage.

src/globus_cli/commands/api.py
sirosen: _I'm not marking this one resolved to ensure that the last note here gets visibility. I changed some `convert()` definitions; this is why._

src/globus_cli/parsing/param_types/notify_param.py
sirosen: With the upgrade to `>= 8.4` this is resolved.

I will note that I wouldn't have favored moving type definitions into `_click_compat` if we hadn't bumped the lower bound though. It would hurt the locality of type information too much.

pyproject.toml
sirosen: _Per top-level comment, resolved as a misunderstanding of `click`'s versioning as semver._

sirosen:
I've done a full pass here, in independently reviewable commits. One side-effect of removing older `click` versions is that `_click_compat` and its usages are all gone.

src/globus_cli/commands/api.py
kurtmckee: Ah, sorry, I missed that in the PR text. Thanks for the info.

kurtmckee (approved):

PR #1267, opened 2026-06-11
Support explicit GLOBUS_CLI_FOLD_TABLES=1

sirosen:
This is either a missing feature or a bugfix, depending on how you view it.
I can't recall if we thought explicitly about `GLOBUS_CLI_FOLD_TABLES=1` before, but I currently am of the mind that not respecting it is a bug.

I have verified that `GLOBUS_CLI_FOLD_TABLES=1 globus group list | cat` produces a (narrow) folded table.

---

This changeset intentionally does not touch the `GLOBUS_CLI_FOLD_TABLES=0` path, which uses the older/legacy table printer.
This was done to allow some escape-hatch if users' scripts need the old format exactly.

### DIFF
--- a/changelog.d/20260611_125949_sirosen_fix_handling_of_truthy_fold_tables.md
+++ b/changelog.d/20260611_125949_sirosen_fix_handling_of_truthy_fold_tables.md
@@ -1,7 +1,7 @@
 ### Bugfixes
 
-* Fixed a bug in which TTY detection ran even when `GLOBUS_CLI_FOLD_TABLES=1`
-  is set.
+* Fixed a bug in which terminal device (TTY) detection ran even when
+  `GLOBUS_CLI_FOLD_TABLES=1` is set.
   Setting the variable to true now forces use of folded tables.
   Non-TTY output devices will be detected as having a fixed width of 80
   characters.
--- a/changelog.d/20260611_125949_sirosen_fix_handling_of_truthy_fold_tables.md
+++ b/changelog.d/20260611_125949_sirosen_fix_handling_of_truthy_fold_tables.md
@@ -1,0 +1,7 @@
+### Bugfixes
+
+* Fixed a bug in which TTY detection ran even when `GLOBUS_CLI_FOLD_TABLES=1`
+  is set.
+  Setting the variable to true now forces use of folded tables.
+  Non-TTY output devices will be detected as having a fixed width of 80
+  characters.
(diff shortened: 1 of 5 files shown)


changelog.d/20260611_125949_sirosen_fix_handling_of_truthy_fold_tables.md
```diff
@@ -0,0 +1,7 @@
+### Bugfixes
+
+* Fixed a bug in which TTY detection ran even when `GLOBUS_CLI_FOLD_TABLES=1`
```
derek-globus: What is TTY detection?

sirosen: The folded table printer checks if stdout is a terminal device (TTY) and if the session is interactive (impossible to do perfectly; we have a heuristic which seems to be working well enough).

Should I describe it differently? "TTY detection" is accurate but if it seems too esoteric, maybe "interactive terminal detection"?

derek-globus: Yeah, I wasn't familiar with it & couldn't find the answer easily by googling.

For a public changelog like this, I'd write it out the first time you use the abbreviation.

```suggestion
* Fixed a bug in which teminal device (TTY) detection ran even when `GLOBUS_CLI_FOLD_TABLES=1`
```

src/globus_cli/termio/context.py
```diff
@@ -112,4 +112,4 @@ def term_is_interactive() -> bool:
 
 def fold_tables() -> bool | None:
     val = os.getenv("GLOBUS_CLI_FOLD_TABLES")
-    return val is None or utils.str2bool(val)
+    return val if val is None else utils.str2bool(val)
```
kurtmckee: I have a preference to phrase this as:


```suggestion
    return None if val is None else utils.str2bool(val)
```

kurtmckee (changes requested):

src/globus_cli/termio/context.py
sirosen: I share that preference! This one _slipped_ past me. 

I haven't tinkered with slyp in a long while, but this feels like an extension of this rule: 
https://slyp.readthedocs.io/en/latest/fixer_rules.html#prefer-return-none-after-none-check

sirosen: Oh, to be clear, let's fix this here without waiting, but I definitely want to make an auto-fixer for this.

sirosen: I have the fixer working in my tests -- I'm running some water through it to make sure there are no surprises and will probably release it shortly.

kurtmckee (approved):

PR #1268, opened 2026-06-16
Make table headers in folded tables bold

sirosen:
- rows can be bolded with ANSI escapes
- the header row passes `bold=True`

---

Sample output:
image

### DIFF
--- a/src/globus_cli/termio/printers/folded_table_printer.py
+++ b/src/globus_cli/termio/printers/folded_table_printer.py
@@ -117,7 +117,7 @@ class FoldedTablePrinter(Printer[t.Iterable[t.Any]]):
         if table.folded:
             echo(_separator_line(col_widths, row_type=SeparatorRowType.box_top))
         # print the header row and a separator
-        echo(table.header_row.serialize(col_widths))
+        echo(table.header_row.serialize(col_widths, bold=True))
         echo(
             _separator_line(
                 col_widths,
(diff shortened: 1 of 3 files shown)


src/globus_cli/termio/printers/folded_table_printer.py
```diff
@@ -315,13 +317,25 @@ def _format_subrow(
...
+            _format_element(subrow[idx], width, bold)
+            if idx < len(subrow)
+            else " " * width
+        )
     return leader + f" {separator} ".join(line) + trailer
 
 
+def _format_element(element: str, width: int, bold: bool) -> str:
+    if bold:
+        return f"\033[1m{element.ljust(width)}\033[0m"
```
derek-globus: Using my newly learned abbreviation, do we need to do TTY detection before including escape characters?

_(This is very much a question, I expect you know this space a lot better than I do)_

derek-globus (commented):
One question, but definitely on board; looks like a super easy win.

src/globus_cli/termio/printers/folded_table_printer.py
kurtmckee: I have a preference for `\N{ESCAPE}` here for clarity (I had to double-check that `\033` matched `\x1b`, which is the "name" that I know the escape byte by).

It may also be helpful to use constants as well, whether created in-function or created globally:

```suggestion
        start_bold = "\N{ESCAPE}[1m"
        end_bold = "\N{ESCAPE}[0m"
        return f"{start_bold}{element.ljust(width)}{end_bold}"
```

sirosen: We would, if this output were being sent in that mode!

We go down a different printing path entirely when stdout is not-a-TTY, so it ends up not being necessary to pass that info along to this point.

sirosen: Ah, yes, named sequences are nice. They're not in my fingertips though.

sirosen: I just got in the weeds to start testing this and I realized it also passes the data through `click.echo`, which strips off ANSI escapes if it doesn't detect a TTY. So we're safe even if we were to change things around in the printing!

kurtmckee (approved):

PR #1276, opened 2026-07-01
Add a check which tests test-mindeps bounds

sirosen:
Load `test-mindeps` and `dependencies` out of pyproject data, and compare
them to ensure that the specifiers in `test-mindeps` always contain the
minimum version listed in `dependencies`.

This is a new tox env and runs in CI. It should force bad test-mindeps
updates from dependabot to fail.

### DIFF
--- a/scripts/ensure_mindeps_match_lower_bounds.py
+++ b/scripts/ensure_mindeps_match_lower_bounds.py
@@ -1,0 +1,49 @@
+# run via `tox r -e check-mindeps-lower-bounds`
+import pathlib
+import re
+import sys
+import tomllib
+
+import mddj.api
+from packaging.dependency_groups import DependencyGroupResolver
+from packaging.requirements import Requirement
+
+dj = mddj.api.DJ()
+REPO_ROOT = pathlib.Path(__file__).parent.parent
+
+# load 'test-mindeps', filtered only to the directly declared requirements, not the
+# includes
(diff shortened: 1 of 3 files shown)


scripts/ensure_mindeps_match_lower_bounds.py
```diff
@@ -0,0 +1,49 @@
...
+# includes
+with open(REPO_ROOT / "pyproject.toml", "rb") as f:
+    pyproject = tomllib.load(f)
+
+group_resolver = DependencyGroupResolver(pyproject["dependency-groups"])
+mindeps = [
+    r for r in group_resolver.lookup("test-mindeps") if isinstance(r, Requirement)
+]
+
+dependencies: tuple[str, ...] = dj.read.dependencies()
```
derek-globus, line 24: Can you remind me where MDDJ is sourcing this metadata from? Is this also coming from pyproject.toml?

derek-globus (approved):

scripts/ensure_mindeps_match_lower_bounds.py
sirosen: It pulls data from pyproject.toml as much as possible. If a field is not present there, it will fallback to a more complex (and slower) path which asks the package build system to get that information.

I should do a better job of documenting this, probably.

